PR #3, opened 2016-10-11
kafkatest support, golint, and more

edenhill:

PR #6, opened 2016-10-17
Fix invalid JSON

bruth:


edenhill:
Thanks, this was just fixed in #3

PR #9, opened 2016-11-04
integration tests

hqin:
@edenhill @ewencp Please review the code


ewencp:
https://godoc.org/golang.org/x/tools/cmd/cover might also help indicate if we're getting good coverage of the APIs without having to manually audit everything.

examples/integration-test/verify_consumer.go
```diff
@@ -0,0 +1,317 @@
...
+	fmt.Fprintf(os.Stderr, "Closing consumer\n")
+	c.Close()
+
+	if exitRequested {
+		os.Exit(1)
+	}
+
+}
+
+// test channel based producer API
```
ewencp: comment doesn't match function name (producer vs consumer)

examples/integration-test/verify_consumer.go
```diff
@@ -0,0 +1,317 @@
...
+// integration tests of consumer APIs
+
+import (
+	"fmt"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
+	"os"
+	"time"
+)
+
+// gather the performance stats of function based consumer API.
```
ewencp: seems duplicated from below method.

more generally, do we need both of these? @edenhill do you see a benefit to having both instead of just having the performance version?

edenhill: We actually already have performance tests for both the Producer and Consumer, so I think they should be removed from these integration tests (and possibly factored into the existing benchmarks if there is a gap).

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+	{100000, []byte("key5"), []byte("value5"), ErrUnknownPartition.String()},
+}
+
+var groupId = "integration-test-group.go"
+//var topic = "integration-test.go"
+
+//Test LibraryVersion()
+func TestLibraryVersion(t *testing.T) {
+	ver, verstr := LibraryVersion()
+	t.Logf("Library version %d: %s\n", ver, verstr)
```
edenhill: Add some kind of verification here, e.g., `ver >= 0x00090200`

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+}
+
+//Test TimestampType
+func TestTimestampType(t *testing.T) {
+	timestamp_map := map[TimestampType]string{TimestampCreateTime: "CreateTime",
+		TimestampLogAppendTime: "LogAppendTime",
+		TimestampNotAvailable:  "NotAvailable"}
+	for ts, desc := range timestamp_map {
+		if ts.String() != desc {
+			t.Errorf("Wrong description")
```
edenhill: "Wrong timestamp description for %s, expected %s"

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+		"earliest": OffsetBeginning,
+		"end":      OffsetEnd,
+		"latest":   OffsetEnd,
+		"unset":    OffsetInvalid,
+		"invalid":  OffsetInvalid,
+		"stored":   OffsetStored}
+	for key, expected_offset := range testOffsets {
+		offset, err := NewOffset(key)
+		if err != nil {
+			t.Errorf("Cannot create offset for %s\n", key)
```
edenhill: print the error too

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+			if offset != expected_offset {
+				t.Errorf("Offset does not equal expected: %s != %s\n", offset, expected_offset)
+			}
+		}
+	}
+
+	// test numeric string conversion
+	offset, err := NewOffset("10")
+	if err != nil {
+		t.Errorf("Cannot create offset for 10\n")
```
edenhill: print err

kafka/event_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+package kafka
+
+import (
+	"testing"
+)
+
+// TestEventAPIs dry-tests the public event related APIs, no broker is needed.
+func TestEventAPIs(t *testing.T) {
+	assigned_partitions := AssignedPartitions{}
```
edenhill: I think snake_case fails golint, use camelCase instead

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+
+	// test OffsetTail()
+	tail := OffsetTail(offset)
+	t.Logf("offset tail %v\n", tail)
+
+}
+
+//Test config map APIs
+func TestConfigMapAPIs(t *testing.T) {
+	if !testconfRead() {
```
edenhill: This test doesn't seem to need to connect to a real broker so the testconf shouldnt be used.

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+	if err != nil {
+		t.Errorf("Failed to set key via SetKey()")
+	}
+
+	// set a good key-value pair via Set()
+	err = kConfig.Set("group.id=test.id")
+	if err != nil {
+		t.Errorf("Failed to set key-value pair via Set()")
+	}
+
```
edenhill: Can you extend this with a test taking an integer too?
And extra points for a custom type that implements fmt.Stringer!

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+	kConfig := &ConfigMap{"bootstrap.servers": testconf.Brokers,
+		"group.id": groupId}
+
+	// Create consumer
+	c, err := NewConsumer(kConfig)
+	if err != nil {
+		t.Errorf("Failed to create consumer: %s\n", err)
+		return
+	}
+	low, high, err := c.QueryWatermarkOffsets(testconf.Topic, 0, 5*1000)
```
edenhill: This doesnt really check whether the returned offsets are valid in any way. What about this:
 * 1: Read QueryWatermarkOffsets
 * 2: Produce N messages to topic
 * 3: Read QueryWatermarkOffsets again
 * 4: Verify that high offest changed by at least N

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+	if err != nil {
+		t.Errorf("Failed to query watermark offsets for topic %s\n", testconf.Topic)
+		return
+	} else {
+		t.Logf("Watermark offsets fo topic %s: low=%d, high=%d\n", testconf.Topic, low, high)
+	}
+}
+
+//Test producer QueryWatermarkOffsets API
+func TestProducerQueryWatermarkOffsets(t *testing.T) {
```
edenhill: Could this one and the one above be unified into a helper function that is called from each test, to avoid code duplication

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+			Key:            testmsg.Key}
+
+		// coverage test for message String() API
+		msgContent := m.String()
+		t.Logf("Producing message %s\n", msgContent)
+
+		err = p.Produce(m, deliveryChan)
+		if err != nil {
+			if testmsg.ExpectedError != "" && err.Error() == testmsg.ExpectedError {
+				fmt.Fprintf(os.Stderr, "OK: Expected error: %s\n", testmsg.ExpectedError)
```
edenhill: t.Logf`?

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+
+	t.Logf("Created Producer %v\n", p)
+
+	deliveryChan := make(chan Event)
+
+	for _, testmsg := range testmsgs {
+		m := &Message{
+			TopicPartition: TopicPartition{Topic: &testconf.Topic, Partition: testmsg.Partition},
+			Value:          testmsg.Value,
+			Key:            testmsg.Key}
```
edenhill: Could we add an opaque too that we check down below before verify?

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+	p, err := NewProducer(kConfig)
+	if err != nil {
+		t.Fatalf("Failed to create producer: %s\n", err)
+	}
+
+	t.Logf("Created Producer %v\n", p)
+
+	deliveryChan := make(chan Event)
+
+	for _, testmsg := range testmsgs {
```
edenhill: camelcase

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+}
+
+// verify event info
+func verifyEvent(e Event, testmsg testmsgType, t *testing.T) {
+	switch e.(type) {
+	case *Message:
+		m := e.(*Message)
+		if m.TopicPartition.Error != nil {
+			err := m.TopicPartition.Error
+			if testmsg.ExpectedError != "" && err.Error() == testmsg.ExpectedError {
```
edenhill: I think it might be better to have the ExpectedError as a Kafka Error and matching directly on that, rather than the string representation.
I believe that's more correct, showcases better API usage (for users that checkout the tests for inspiration) and also makes the code cleaner.

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+				fmt.Fprintf(os.Stderr, "OK: Expected error: %s\n", testmsg.ExpectedError)
```
edenhill: t.Logf

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+			(testmsg.Partition >= 0 && m.TopicPartition.Partition == testmsg.Partition)) {
+			t.Errorf("Partition check failed for test message %v. Got back {[%d] [%s] [%s]}\n",
+				testmsg, m.TopicPartition.Partition, m.Key, m.Value)
+			return
+
+		}
+
+		// check Key
+		if testmsg.Key == nil {
+			if !(m.Key == nil || len(m.Key) == 0) {
```
edenhill: Maybe we should differentiate between null key and empty key.

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+				t.Errorf("Key verification failed. Expected %s, got %s\n", testmsg.Key, m.Key)
+				return
```
edenhill: If we remove the 'return's in this fail clauses we could get all errors reported for a message, which might help troubleshooting.

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+			}
+		} else {
+			if string(m.Value) != string(testmsg.Value) {
+				t.Errorf("Value verification failed. Expected %s, got %s\n", testmsg.Value, m.Value)
+				return
+			}
+		}
+
+		fmt.Printf("%% Message delivered to topic %s [%d] at offset %v: key=(%s), value=(%s)\n",
```
edenhill: t.Logf

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+	runConsumerFuncBasedAPI(t, "Commit")
+}
+
+func runConsumerFuncBasedAPI(t *testing.T, commitMode string) {
+	if !testconfRead() {
+		t.Skipf("Missing testconf.json")
+	}
+	c, err := NewConsumer(&ConfigMap{
+		"bootstrap.servers":    testconf.Brokers,
+		"group.id":             groupId,
```
edenhill: Since these tests will be run on someones broker we probably want the groupid to be configurable through testconf.json too

kafka/integration_test.go
edenhill: Also add a test for  auto.offset.reset using the offset constants, e.g.: OffsetEarliest

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+		fmt.Fprintf(os.Stderr, "Failed to create consumer: %s\n", err)
+		os.Exit(1)
+	}
+
+	t.Logf("Created Consumer %v\n", c)
+
+	err = c.Subscribe(testconf.Topic, nil)
+
+	maxMsgCnt := 3
+	msgcnt := 0
```
edenhill: inconsistent (non-)camelcasing

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+	run := true
+
+	savedMessages := make([]*Message, maxMsgCnt)
+
+	for run == true {
+		select {
+		default:
+			ev := c.Poll(100)
```
edenhill: Since nothing else runs you could increase the timeout to "10*1000"

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+			*m.TopicPartition.Topic, m.TopicPartition.Partition, m.TopicPartition.Offset, m.Key, m.Value)
+
+	default:
+		fmt.Printf("Ignored event: %s\n", e)
+	}
+}
+
+// TestConsumerFuncBasedAPI test function-based API, requires broker
+func TestConsumerFuncBasedAPI(t *testing.T) {
+	runConsumerFuncBasedAPI(t, "CommitMessage")
```
edenhill: If these tests rely on messages being available in the topic then that topic needs to be primed, maybe you can move this baby to testhelper and reuse it, it checks that the expected number of messages are available on the topic and if not produces to it:
https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/consumer_performance_test.go#L128

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+			switch e := ev.(type) {
+			case *Message:
+				fmt.Printf("%% Message on %s: key=(%s), value=(%s)\n",
+					e.TopicPartition, e.Key, e.Value)
+				savedMessages[msgcnt] = e
+				msgcnt++
+				if msgcnt >= maxMsgCnt {
+					run = false
+				}
+			case PartitionEOF:
```
edenhill: This isn't safe, it could reach EOF on any partition before having consumed expected number of messages from other partitions.
So simply ignore this event.

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+		}
+	}
+
+	switch commitMode {
+	case "CommitMessage":
+		// verify CommitMessage() API
+		for _, message := range savedMessages {
+			_, commit_err := c.CommitMessage(message)
+			if commit_err != nil {
+				t.Errorf("Cannot commit message\n")
```
edenhill: print err

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+		}
+	case "CommitOffsets":
+		// verify CommitOffset
+		partitions := make([]TopicPartition, len(savedMessages))
+		for index, message := range savedMessages {
+			partitions[index] = message.TopicPartition
+		}
+		_, commit_err := c.CommitOffsets(partitions)
+		if commit_err != nil {
+			t.Errorf("Failed to commit using CommitOffsets\n")
```
edenhill: print error

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+			}
+		}
+	}
+
+	switch commitMode {
+	case "CommitMessage":
+		// verify CommitMessage() API
+		for _, message := range savedMessages {
+			_, commit_err := c.CommitMessage(message)
+			if commit_err != nil {
```
edenhill: Verify that the committed offsets match the message.

kafka/integration_test.go
edenhill: verify

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+		_, commit_err := c.CommitOffsets(partitions)
+		if commit_err != nil {
+			t.Errorf("Failed to commit using CommitOffsets\n")
+		}
+	default:
+		// verify Commit() API
+		_, commit_err := c.Commit()
+		if commit_err != nil {
+			t.Errorf("Failed to commit: %s", commit_err)
+		}
```
edenhill: verify

kafka/integration_test.go
```diff
@@ -0,0 +1,529 @@
...
+		"enable.auto.commit":       false,
+		"session.timeout.ms":       6000,
+		"go.events.channel.enable": true,
+		"default.topic.config":     ConfigMap{"auto.offset.reset": "earliest"}})
+
+	if err != nil {
+		t.Fatalf("Failed to create consumer: %s\n", err)
+	}
+
+	t.Logf("Created Consumer %v\n", c)
```
edenhill: Could we avoid code duplication (and provide some verification in this one) by reusing the code above from runCon..?

edenhill:
When updating a PR with new commits please squash/fixup them to avoid "fix review comments" commits.
E.g.:
````
  $ git commit -m "review fixes" ....
   # repeat as necessary
  $ git rebase -i master
  # Move fixes down under the main feature commit and mark them as 'f' for fixup, this way they'll be silently joined with the parent commit.
````

**Comment by edenhill** [comment]
ping @hqin

File: kafka/integration_test.go [hunk 30]
```diff
@@ -0,0 +1,877 @@
...
+
+package kafka
+
+import (
+	"fmt"
+	"testing"
+)
+
+// producer test control
+type PC struct {
```
ewencp: Are we using shorthand for a reason? It makes the code harder to read if you're not already familiar with it.

kafka/integration_test.go
```diff
@@ -0,0 +1,877 @@
...
+
+	if testMsgsInit {
+		return
+	}
+	defer func() { testMsgsInit = true }()
+
+	testmsgs := make([]*testmsgType, 100)
+	i := 0
+
+	/*
```
ewencp: If this isn't needed anymore we should just get rid of it.

kafka/integration_test.go
```diff
@@ -0,0 +1,877 @@
...
+			Value:  []byte(fmt.Sprintf("value%d", i)),
+			Key:    []byte(fmt.Sprintf("key%d", i)),
+			Opaque: fmt.Sprintf("Op%d", i),
+		}}
+	i++
+
+	pAllTestMsgs = testmsgs[:i]
+}
+
+//Test LibraryVersion()
```
ewencp: Some of these seem like normal unit tests, maybe they should be relocated to the relevant `_test.go` files?

kafka/integration_test.go
```diff
@@ -0,0 +1,877 @@
...
+}
+
+// msgtrackerStart sets up a new message tracker
+func msgtrackerStart(t *testing.T, expectedCnt int) (mt msgtracker) {
+	mt = msgtracker{t: t}
+	mt.msgs = make([]*Message, expectedCnt)
+	return mt
+}
+
+var testMsgsInit  = false
```
ewencp: Need to run this file through `gofmt`, this line has an extra space.

kafka/integration_test.go
```diff
@@ -0,0 +1,877 @@
...
+	producerTest(t, "Channel producer (with DR)",
+		nil, PC{withDr: true},
+		func(p *Producer, m *Message, drChan chan Event) {
+			p.ProduceChannel() <- m
+		})
+
+}
+
+// test batch producer channel-based API without delivery report
+func xTestProducerBatchChannel(t *testing.T) {
```
ewencp: Are these two still WIP?

kafka/integration_test.go
hqin: Thanks for the feedback. I changed CC and PC to consumerCtrl and producerCtrl for readability.

ewencp:
This LGTM now. @edenhill you want to take another pass?

edenhill:
This looks like a test failure that wasn't properly detected:
https://travis-ci.org/confluentinc/confluent-kafka-go/jobs/179096796#L218

kafka/producer.go
```diff
@@ -57,6 +57,10 @@ func (p *Producer) gethandle() *handle {
 }
 
 func (p *Producer) produce(msg *Message, msgFlags int, deliveryChan chan Event) error {
+	if msg == nil || msg.TopicPartition.Topic == nil || len(*msg.TopicPartition.Topic) == 0 {
+		return newErrorFromString(ErrUnknownTopic, "")
```
edenhill: Should be ErrInvalidArg

kafka/message.go
```diff
@@ -209,7 +209,7 @@ func (h *handle) messageToC(msg *Message, cmsg *C.rd_kafka_message_t) {
 			valp = payload
 		}
 		if keyLen > 0 {
-			copy((*[1 << 30]byte)(payload)[valueLen:keyLen], msg.Key)
+			copy((*[1 << 30]byte)(payload)[valueLen:allocLen], msg.Key)
```
edenhill: Good find!

kafka/integration_test.go
```diff
@@ -787,7 +621,7 @@ func xTestProducerBatchChannelDR(t *testing.T) {
 // use opaque string to locate the matching test message for message verification
 func findExpectedMessage(expected []*testmsgType, opaque string) *testmsgType {
 	for i, m := range expected {
-		if expected[i].msg.Opaque.(string) == opaque {
+		if expected[i].msg.Opaque != nil && expected[i].msg.Opaque.(string) == opaque {
```
edenhill: I think we actually want to compare the Opaque pointer value to be equal, not the contents.
This isn't critical though, so you dont need to fix it if it means a larger test overhaul.

kafka/integration_test.go
```diff
@@ -747,6 +551,36 @@ func TestProducerFuncDR(t *testing.T) {
 		})
 }
 
+// test producer with bad messages
+func TestProducerWithBadMessages(t *testing.T) {
+	if !testconfRead() {
```
edenhill: You dont need a broker for these tests, so skip the testconf dependency so it runs on travis.

kafka/config_test.go
```diff
@@ -0,0 +1,88 @@
...
+	Port int
+}
+
+// implements String() interface
+func (hp HostPortType) String() string {
+	return fmt.Sprintf("%s:%d", hp.Host, hp.Port)
+}
+
+//Test config map APIs
+func TestConfigMapAPIs(t *testing.T) {
```
edenhill: great tests!

kafka/config_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	//test offset constant
+	err = config.SetKey("{topic}.auto.offset.reset", OffsetBeginning)
+	if err != nil {
+		t.Errorf("Failed to set key via SetKey(). Error: %s\n", err)
+	}
+
+	//test integer offset
+	err = config.SetKey("{topic}.auto.offset.reset", 10)
```
edenhill: Not that it is checked by SetKey, but 10 is actually an invalid value for "auto.offset.reset" and this might confuse people, so use some other property to test this, e.g. "{topic}.message.timeout.ms"

edenhill:
The event_test error is still happening on travis:
`	event_test.go:32: EOF at %!s(PANIC=runtime error: invalid memory address or nil pointer dereference)`

ewencp:
@edenhill @hqin This went into master, but should we consider also putting it on 3.1.x since it just adds tests + bug fixes?

hqin:
It is fine with me.

PR #11, opened 2016-11-08
Travis CI integration

edenhill:


.travis.yml
```diff
@@ -0,0 +1,18 @@
+language: go
+go:
```
ewencp: Are these the versions we want here given we are claiming support for more versions? It sounds like Travis supports many more versions.

edenhill: There was a cgo compilation failure on go 1.5 so I ripped it out, for now.
The Go community seems pretty up to date so I dont think we need to support older versions until people ask us to.

ewencp: Sounds fine. For some reason I thought we said something about supporting 1.2+, but I guess that's not in a README, just from a commit fixing it up.

PR #19, opened 2016-12-07
Producer timestamp

edenhill:


kafka/producer.go
```diff
@@ -27,11 +27,30 @@ import (
...
+rd_kafka_resp_err_t do_produce (rd_kafka_t *rk,
+          rd_kafka_topic_t *rkt, int32_t partition,
+          int msgflags,
+          void *val, size_t val_len, void *key, size_t key_len,
+          int64_t timestamp,
+          uintptr_t cgoid) {
+
+#if RD_KAFKA_VERSION >= 0x00090300
+  return rd_kafka_producev(rk,
+        RD_KAFKA_V_RKT(rkt),
```
ewencp: Any reason for using the rkt version here but in Python you passed the topic string?

kafka/producer.go
```diff
@@ -99,16 +118,23 @@ func (p *Producer) produce(msg *Message, msgFlags int, deliveryChan chan Event)
 		cgoid = p.handle.cgoPut(cgoDr{deliveryChan: deliveryChan, opaque: msg.Opaque})
 	}
 
-	r := int(C.do_produce(crkt, C.int32_t(msg.TopicPartition.Partition),
+	var timestamp int64
+	if !msg.Timestamp.IsZero() {
```
ewencp: Is this branch actually worth checking first just to avoid doing the division?

edenhill: Fun fact: Go's `time` default value is not 1970, but 0001-01-01..
Very convenient.

.travis.yml
```diff
@@ -6,14 +6,15 @@ go:
...
-  directories:
-    - tmp-build-${LIBRDKAFKA_VERSION}
+  matrix:
+    - LIBRDKAFKA_VERSION=v0.9.2
+    - LIBRDKAFKA_VERSION=master
+
 before_install:
-  - test -f tmp-build-${LIBRDKAFKA_VERSION}/lib/librdkafka.a || bash mk/bootstrap-librdkafka.sh ${LIBRDKAFKA_VERSION} tmp-build-${LIBRDKAFKA_VERSION}
+  - rm -rf tmp-build
+  - test -f tmp-build/lib/librdkafka.a || bash mk/bootstrap-librdkafka.sh ${LIBRDKAFKA_VERSION} tmp-build
```
ewencp: I guess it doesn't hurt, but given the above `rm -rf tmp-build`, won't the initial `test` here always fail?

edenhill: With all these distributed systems you can never be too sure

PR #28, opened 2017-01-13
safely handle slice message values

### LINKED ISSUE #24: json.Marshal results in "panic: runtime error: cgo argument has Go pointer to Go pointer"
Hi there!

Changing the producer_example.go to marshal a struct (see below) results in this error:

```
» go run producer_example.go localhost:9092 foo
Created Producer rdkafka#producer-1
panic: runtime error: cgo argument has Go pointer to Go pointer

goroutine 1 [running]:
panic(0x40d59a0, 0xc420ff0090)
	/usr/local/Cellar/go/1.7.4_1/libexec/src/runtime/panic.go:500 +0x1a1
github.com/confluentinc/confluent-kafka-go/kafka._cgoCheckPointer0(0xc42009a024, 0x0, 0x0, 0x0, 0x0)
	??:0 +0x59
github.com/confluentinc/confluent-kafka-go/kafka.(*Producer).produce(0xc4200a0000, 0xc42004deb0, 0x0, 0xc420098120, 0x0, 0xc42009a000)
	/Users/matthias/code/src/github.com/confluentinc/confluent-kafka-go/kafka/producer.go:130 +0x31b
github.com/confluentinc/confluent-kafka-go/kafka.(*Producer).Produce(0xc4200a0000, 0xc42004deb0, 0xc420098120, 0x11, 0x40)
	/Users/matthias/code/src/github.com/confluentinc/confluent-kafka-go/kafka/producer.go:154 +0x48
main.main()
	/Users/matthias/code/src/github.com/confluentinc/confluent-kafka-go/examples/producer_example/producer_example.go:53 +0x454
exit status 2
```

Here are the changes to producer_example.go:

```
diff --git a/examples/producer_example/producer_example.go b/examples/producer_example/producer_example.go
index f81f98a..1bf4ab4 100644
--- a/examples/producer_example/producer_example.go
+++ b/examples/producer_example/producer_example.go
@@ -18,9 +18,11 @@ package main
  */

 import (
+       "encoding/json"
        "fmt"
-       "github.com/confluentinc/confluent-kafka-go/kafka"
        "os"
+
+       "github.com/confluentinc/confluent-kafka-go/kafka"
 )

 func main() {
@@ -47,7 +49,7 @@ func main() {
        // .Events channel is used.
        deliveryChan := make(chan kafka.Event)

-       value := "Hello Go!"
+       value, _ := json.Marshal(struct{ M string }{M: "Hello Go!"})
        err = p.Produce(&kafka.Message{TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny}, Value: []byte(value)}, deliveryChan)

        e := <-deliveryChan
```

Should this work?

mcandre:
This should fix #24.

Note: Requires Go 1.7+ for `C.CBytes()` support. Some of the CI jobs fail due to attempting to build this patch for Go 1.6-.

I tried to rewrite the `C.CBytes()` call using `C.memcpy` for Go 1.6- support, but that's giving me errors. If someone knows cgo really well, maybe we can massage `valp` better.

### DIFF
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -84,22 +84,8 @@ func (p *Producer) produce(msg *Message, msgFlags int, deliveryChan chan Event) 
 
 	crkt := p.handle.getRkt(*msg.TopicPartition.Topic)
 
-	var keyp *byte
-	var empty byte
-	valLen := 0
-	keyLen := 0
-
-	if msg.Value != nil {
-		valLen = len(msg.Value)
-	}
-	if msg.Key != nil {
-		keyLen = len(msg.Key)
-		if keyLen > 0 {
-			keyp = &msg.Key[0]
-		} else {
-			keyp = &empty
-		}
-	}
+	valLen := len(msg.Value)
+	keyLen := len(msg.Key)
 
 	var cgoid int
 
@@ -122,7 +108,7 @@ func (p *Producer) produce(msg *Message, msgFlags int, deliveryChan chan Event) 
 		C.int32_t(msg.TopicPartition.Partition),
 		C.int(msgFlags)|C.RD_KAFKA_MSG_F_COPY,
 		C.CBytes(msg.Value), C.size_t(valLen),
-		unsafe.Pointer(keyp), C.size_t(keyLen),
+		C.CBytes(msg.Key), C.size_t(keyLen),
 		C.int64_t(timestamp),
 		(C.uintptr_t)(cgoid))
 	if cErr != C.RD_KAFKA_RESP_ERR_NO_ERROR {
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -104,11 +104,17 @@ func (p *Producer) produce(msg *Message, msgFlags int, deliveryChan chan Event) 
 		timestamp = msg.Timestamp.UnixNano() / 1000000
 	}
 
+	msgBytes := C.CBytes(msg.Value)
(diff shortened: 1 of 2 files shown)


matthias-margush:
@mcandre Thanks for looking into this (I'm not familiar enough with Go to have narrowed it down). I think that `msg.Key` will need the same fix as `msg.Value` to work correctly if a slice is used as the key.

ewencp:
@mcandre You can find the CLA [here](http://clabot.confluent.io/cla)

ewencp:
@mcandre @edenhill can probably respond better, but I think one of the things we want to keep an eye on here is performance -- I think you're now copying things twice, once in Go and once in librdkafka. I think you're also leaking the memory allocated by `CBytes`, which that method claims you are responsible for cleaning up after.

Also , dropping support for sufficiently old versions of Go is reasonable since people tend to pick up new versions pretty aggressively, but I'm not sure supporting only the most recent major version makes sense.

mcandre:
> dropping support for sufficiently old versions of Go is reasonable

Can we drop go1.2 from CI?

> I think you're now copying things twice, once in Go and once in librdkafka

Sadly, I don't know enough C, cGo, or Go to address this.

edenhill:
Fixed in https://github.com/confluentinc/confluent-kafka-go/pull/83

PR #34, opened 2017-01-18
Add Seek API to consumer

conorbranagan:
Using rd_kafka_seek implemented in the rdkafka library we just expose this on the consumer interface. I had this take a `TopicPartition` as the argument to match the way other methods worked but I can switch it to take topic/partition/offset args if that makes more sense.


edenhill:
Thanks for this!

PR #47, opened 2017-02-18
Updated librdkafka error codes

edenhill:


mhowlett:
LGTM. 

we don't have these in [confluent-kafka-dotnet](https://github.com/confluentinc/confluent-kafka-dotnet) either.

PR #50, opened 2017-03-01
updated readme

mhowlett:
@edenhill @gwenshap @ewencp

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,27 @@
-Confluent's Apache Kafka client for Golang
-==========================================
+Confluent's Golang Client for Apache Kafka<sup>TM</sup>
+=====================================================
 
-Confluent's Kafka client for Golang wraps the librdkafka C library, providing
-full Kafka protocol support with great performance and reliability.
+**confluent-kafka-go** is Confluent's Golang client for [Apache Kafka](http://kafka.apache.org/) and the
+[Confluent Platform](https://www.confluent.io/product/).
+
+Features:
+
+- **High performance** - confluent-kafka-go is a lightweight wrapper around
+[librdkafka](https://github.com/edenhill/librdkafka), a finely tuned C
+client.
+
+- **Reliability** - There are a lot of details to get right when writing an Apache Kafka
+client. We get them right in one place (librdkafka) and leverage this work
+across all of our clients (also [confluent-kafka-python](https://github.com/confluentinc/confluent-kafka-python)
+and [confluent-kafka-dotnet](https://github.com/confluentinc/confluent-kafka-dotnet)).
+
+- **Supported** - Commercial support is offered by
+[Confluent](https://confluent.io/).
+
+- **Future proof** - Confluent, founded by the
+creators of Kafka, is building a [streaming platform](https://www.confluent.io/product/)
+with Apache Kafka at its core. It's high priority for us that client features keep
+pace with core Apache Kafka and components of the [Confluent Platform](https://www.confluent.io/product/).
 
 The Golang bindings provides a high-level Producer and Consumer with support
 for the balanced consumer groups of Apache Kafka 0.9 and above.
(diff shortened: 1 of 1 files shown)


sriramsub:
I think it helps to talk about Confluent Open Source package somewhere and point to the downloads page. This helps us to track references on the download page from the client page.

README.md
```diff
@@ -1,8 +1,27 @@
...
+
+Features:
+
+- **High performance** - confluent-kafka-go is a lightweight wrapper around
+[librdkafka](https://github.com/edenhill/librdkafka), a finely tuned C
+client.
+
+- **Reliability** - There are a lot of details to get right when writing an Apache Kafka
+client. We get them right in one place (librdkafka) and leverage this work
+across all of our clients (also [confluent-kafka-python](https://github.com/confluentinc/confluent-kafka-python)
```
ewencp: Perhaps we can get a clients page on the main website that we can link to? Would make it less maintenance across the repos to keep these lists up to date (also, eventually they may become longer than we want to have included in each repo). The direct links here don't seem all that valuable since most people are looking for a specific language.

mhowlett: good point. for now we could leave as is or link to the confluent docs page: http://docs.confluent.io/3.1.0/clients/index.html
I have a slight preference for leaving as is until we re-work our client documentation.

README.md
```diff
@@ -1,8 +1,27 @@
-Confluent's Apache Kafka client for Golang
-==========================================
+Confluent's Golang Client for Apache Kafka<sup>TM</sup>
```
edenhill: -1 on this change, I think the previous version reads better. (this is an Apache Kafka client for Golang, not a Golang client (?) for Apache Kafka).

mhowlett:
@sriramsub - was just reviewing possible places to link to:

- https://www.confluent.io/product/compare/
- https://www.confluent.io/product/
- https://www.confluent.io/download/

I actually think the comparison page would be the best landing page as it very quickly and effectively gets the message across as to what the deal is - CP Enterprise is a superset of CP OSS (which is OS) which is a superset of AK. And it provides an obvious link to download. I think the download page doesn't provide enough context and people will be less inclined to download if they land straight on this page. Finally, I think the product page is too long winded for typical 5 second attention spans. It's linked to at the bottom of the comparison page ("learn more"), and i think provides an appropriate level of detail given people have clicked on that link.

we could put in brackets on the first line after Confluent Platform something like "OSS and Enterprise editions". I don't know on this one, but I my feeling is people will click on "confluent platform" anyway out of curiosity as to what CP is, and then it'd be immediately obvious from the comparison page that there is an OS version + additional context as to what this actually means (i.e. something that is a superset of Apache Kafka).

so in summary, i'm in favor of changing all https://www.confluent.io/product/ links to https://www.confluent.io/product/compare/

README.md
mhowlett: I anguished over this change for a good 30 minutes. The problem is i think the line below is very important (that it's a client for Kafka and Confluent Platform), and it's very difficult to write that if we put it the way around that you suggest (which I too think makes more sense). Worth noting that a lot of other people say it the way I changed it to (eg https://github.com/dpkp/kafka-python - python client for Apache Kafka).

mhowlett:
@edenhill you're not sure the entire PR adds value? I think it adds a lot of value - we need to spell things out - people are not going to think all this by default. Also, the message that CP is a big deal worthy of attention + number of people clicking through is going to be much more effective / much higher if CP is mentioned more than once.

README.md
ewencp: I'm a bit confused about the discussion because both versions read fine to me...

edenhill:
Oh, I reverted my uneducated complaints on the title in the PR.
LGTM

edenhill:
@mhowlett we're all happy with this, are you? If so, let's make sure it can be merged

mhowlett:
yep, i'll sort out the conflict and merge.

Incidentally, I don't like the "request changes" feature in github because people forget to accept the changes (e.g. in this case) and I think the vibe it gives is too authoritarian.

PR #53, opened 2017-03-17
Fix crash

thomas91310:


thomas91310:
sorry about that

kafka/producer.go
```diff
@@ -128,7 +121,7 @@ func (p *Producer) produce(msg *Message, msgFlags int, deliveryChan chan Event)
 	cErr := C.do_produce(p.handle.rk, crkt,
 		C.int32_t(msg.TopicPartition.Partition),
 		C.int(msgFlags)|C.RD_KAFKA_MSG_F_COPY,
-		unsafe.Pointer(valp), C.size_t(valLen),
+		C.CBytes(msg.Value), C.size_t(valLen),
```
edenhill: CBytes makes a copy of the payload, which we would want to avoid, or handle by using MSG_F_FREE.

It also requires Golang >= 1.7, a limitation we don't want to enforce yet.

kafka/producer_test.go
```diff
@@ -40,9 +45,16 @@ func TestProducerAPIs(t *testing.T) {
...
+	data, err := json.Marshal(envelope)
+
+	if err != nil {
+		t.Error(err)
+	}
+
 	// Produce with function, DR on passed drChan
 	err = p.Produce(&Message{TopicPartition: TopicPartition{Topic: &topic1, Partition: 0},
-		Value: []byte("Own drChan"), Key: []byte("This is my key")},
+		Value: data, Key: []byte("This is my key")},
```
edenhill: Please add a new test action instead of changing existing since we want to be sure to maintain previous functionality too.

edenhill:
Can you provide some details on how to crash the original code, a short usage snippet?

thomas91310:
Hey @edenhill, I'll try to do that soon.

I have another question and it might

edenhill:
Not specifying the key should create a null key, not an empty string.
I would assume the Java deserializer to handle this properly, I'll check it out.

ewencp:
Null key vs empty string are definitely distinguished in the deserializer.

But I think I'm a bit confused in the first place -- this seems to be touching on kafka-rest issues, but this client doesn't work with kafka-rest at all? What is the connection between go code and the REST proxy?

edenhill:
@ewencp I believe he is trying to consume messages with kafka-rest that were produced by confluent-kafka-go

ewencp:
Ah! That makes sense. The Java deserializer definitely handles this correctly (nulls are preserved as nulls), so if there is a problem it'd be in the REST proxy. It'd be helpful to understand the end-to-end setup so that if there is an underlying issue we can set up a reproduction.

edenhill:
@thomas91310 Can you describe your system setup, what client software is producing to the topic and what software is consuming from it, etc?

PR #63, opened 2017-05-24
Allow users to retrive parsed statistic

WaeCo:
This PR will expose a `Parse()` function on the `Stats` Event that returns a go struct `ParsedStats` containing all the parsed stats.

It is useful in case where the user does not want to forward/print the entire json string but rather extract certain parts. Before this had to be done by hand.
I think this should be done by the library rather then each app on its own. 

For performance reasons I did not want to parse it by default. Maybe we could introduce a config string (e.g. `go.statistics.parse: true`) for automated parsing.

One question left: Should I handle Unmarshal errors in `event.go:161`? And if so how? This should actually never happen, since it is covert by the tests and should only fail if librdkafka screwed up.

### DIFF
--- a/kafka/event.go
+++ b/kafka/event.go
@@ -17,6 +17,7 @@ package kafka
  */
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"unsafe"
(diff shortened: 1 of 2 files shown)


WaeCo:
[clabot:check]

PR #66, opened 2017-05-31
buffer signal channel

mhowlett:
the example works as is (though maybe only 99.99% of the time?), but this tripped me up when I extended the select statement to do more things. refer to: https://golang.org/pkg/os/signal/#Notify


edenhill:
good catch, I'd completely missed this:
> Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient.

A bit counter-intuitive behaviour when it comes to channels..

PR #68, opened 2017-05-31
readme: move the getting started up so will people see it

travisjeffery:
- Moves the getting started section up so people will see it
- Adds specific instructions for install the go client


mhowlett:
nice improvement :-) that reads well to me, but best wait for @edenhill's LGTM as I don't know if I have any authority here yet..

edenhill:
Thanks!

PR #72, opened 2017-07-10
Added Consumer.OffsetsStore()

ctrochalakis:
Expose librdkafka's `rd_kafka_offsets_store()` introduced in v0.11.0 (not yet released).

Consumer.OffsetsStore() allows manual control over the committed offsets without sacrificing auto-commit functionality.

I am not sure if returning the storedOffsets is the correct thing to do, we could stop returning them to keep the API cleaner, or return a list of the failed TopicPartitions.

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -194,6 +194,29 @@ func (c *Consumer) CommitOffsets(offsets []TopicPartition) ([]TopicPartition, er
 	return c.commit(offsets)
 }
 
+// OffsetsStore stores the provided list of offsets that will be committed
+// to the offset store according to `auto.commit.interval.ms` or manual
+// offset-less Commit().
+//
+// Returns the stored offsets on success. If at least one offset couldn't be stored,
+// an error and a list of offsets is returned. Each offset can be checked for
+// specific errors via its `.Error` member.
+func (c *Consumer) OffsetsStore(offsets []TopicPartition) (storedOffsets []TopicPartition, err error) {
+	coffsets := newCPartsFromTopicPartitions(offsets)
+	defer C.rd_kafka_topic_partition_list_destroy(coffsets)
+
+	cErr := C.rd_kafka_offsets_store(c.handle.rk, coffsets)
+
+	// coffsets might be annotated with an error
+	storedOffsets = newTopicPartitionsFromCparts(coffsets)
+
+	if cErr != C.RD_KAFKA_RESP_ERR_NO_ERROR {
+		return storedOffsets, newError(cErr)
+	}
+
+	return storedOffsets, nil
+}
+
 // Seek seeks the given topic partitions using the offset from the TopicPartition.
 //
 // If timeoutMs is not 0 the call will wait this long for the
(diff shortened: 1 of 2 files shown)


edenhill:
Thank you for this, will tend to it shortly.

kafka/consumer.go
```diff
@@ -194,6 +194,29 @@ func (c *Consumer) CommitOffsets(offsets []TopicPartition) ([]TopicPartition, er
...
 }
 
+// OffsetsStore stores the provided list of offsets that will be committed
+// to the offset store according to `auto.commit.interval.ms` or manual
+// offset-less Commit().
+//
+// Returns the stored offsets on success. If at least one offset couldn't be stored,
+// an error and a list of offsets is returned. Each offset can be checked for
+// specific errors via its `.Error` member.
+func (c *Consumer) OffsetsStore(offsets []TopicPartition) (storedOffsets []TopicPartition, err error) {
```
edenhill: Should probably be called StoreOffsets to be consistent with CommitOffsets (even though it makes it inconsistent with librdkafka, but that's a hidden gem).

kafka/consumer_test.go
```diff
@@ -54,7 +54,6 @@ func TestConsumerAPIs(t *testing.T) {
 	if err != nil {
 		t.Errorf("SubscribeTopics failed: %s", err)
 	}
-
```
edenhill: Remove whitespace diff

kafka/consumer_test.go
```diff
@@ -65,6 +64,20 @@ func TestConsumerAPIs(t *testing.T) {
...
+	topic := "gotest"
+	stored, err := c.OffsetsStore([]TopicPartition{{Topic: &topic, Partition: 0, Offset: 1}})
+	if err != nil && err.(Error).Code() != ErrUnknownPartition {
+		t.Errorf("OffsetsStore() failed: %s", err)
+		toppar := stored[0]
+		if toppar.Error.(Error).Code() == ErrUnknownPartition {
+			t.Errorf("OffsetsStore() TopicPartition error: %s", toppar.Error)
+		}
+	}
+	stored, err = c.OffsetsStore(nil)
```
edenhill: Is there a use for accepting a nil partition set?

ctrochalakis: `nil` is a valid value for a slice, in fact it is the default value. I wanted to make sure it is explicitly handled. In the fixup I changed that to an uninitialized (default) slice.

edenhill: Right, the data type allows it, but is there a point for the StoreOffsets API to allow it? Passing a nil slice should be considered a usage error and not something we need to explicitly handle (I do agree that error handling is necessary, but parameter checking in APIs is not as fruitful as it seems IMHO unless there is a security concern to it).

ctrochalakis: Let's call it an empty slice. I think that the API should allow passing an empty slice, whatever that means in Go. If someone chooses to pass an empty slice a proper error is returned. We don't explicitly handle it (as in `consumer.go` code) but we explicitly test that an empty `[]TopicPartition` is properly handled.

ctrochalakis: note that this part of the diff is now changed to:

```golang
        var empty []TopicPartition
        stored, err = c.StoreOffsets(empty)
```

which seems like a valid usage scenario.

PR #74, opened 2017-07-12
fix unsafe.Pointer use to be compatible with cgo rules

calebdoxsey:
According to https://github.com/golang/go/issues/14210 it's necessary to take the address of a byte slice in the same expression as the cgo function invocation.

IE do this:

    C.f(unsafe.Pointer(&dst[0]))

And not this:

    tmp := unsafe.Pointer(&dst[0])
    C.f(tmp)

This PR will fix https://github.com/confluentinc/confluent-kafka-go/issues/24 without resorting to `CBytes`, and therefore avoiding a copy.

### DIFF
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -57,6 +57,8 @@ rd_kafka_resp_err_t do_produce (rd_kafka_t *rk,
 */
 import "C"
 
+var emptyByteSlice = []byte{0}
+
 // Producer implements a High-level Apache Kafka Producer instance
 type Producer struct {
 	events         chan Event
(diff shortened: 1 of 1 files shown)


edenhill:
This is great stuff, will tend to this soon.
Thank you

kafka/producer.go
```diff
@@ -84,28 +86,15 @@ func (p *Producer) produce(msg *Message, msgFlags int, deliveryChan chan Event)
...
 
-	var valp *byte
-	var keyp *byte
-	var empty byte
-	valLen := 0
-	keyLen := 0
-
-	if msg.Value != nil {
-		valLen = len(msg.Value)
-		// allow sending 0-length messages (as opposed to null messages)
```
edenhill: There is a difference between an empty Vallue/Key field (length 0) and a Null message (Value/Key is nil).
Empty messages are treated like any other message, just empty, while Null messages values are considered tombstones and null message keys are considered key-less.
That's the reason why the original code initialized both valp and keyp to nil (NULL in C) and I'm guessing this will break the cast-on-same-line requirement of Go.
One way around this might be to have four different code paths depending on the nullity of key and value, e.g.:
```
 if  key && val {
    produce(..., key, val, ..)
 else if key && !val {
   produce(..., key, C.NULL, ..)
 ..and so on
```

edenhill:
Fixed in https://github.com/confluentinc/confluent-kafka-go/pull/83

Thank you for your PR, much appreciated!

PR #83, opened 2017-08-08
API catchup and some bug fixes

edenhill:


kafka/config.go
```diff
@@ -183,6 +183,14 @@ func (m ConfigMap) convert() (cConf *C.rd_kafka_conf_t, err error) {
 // If the key is not found defval is returned.
 // If the key is found but the type is mismatched an error is returned.
 func (m ConfigMap) get(key string, defval ConfigValue) (ConfigValue, error) {
+	if strings.HasPrefix(key, "{topic}.") {
```
mhowlett: is this the first client where we de-nest topic configuration?

have we considered (and is it worth having a plan for) how this might generalize to specific topic config? what you have could work I think, but i feel like I want to see this worked through before committing to anything. I'm not sure about the notation `{topic}` but not immediately having better ideas.

edenhill: librdkafka v0.11.0 added support for automatic fall-thru of global -> default.topic.config properties: any property set on the global object that turns out to be unknown will be attempted on the default topic config.
That makes this syntax obsolete, but since the Set() API already supports it we can't remove it, and it was better to add support to the corresponding Get() call as well.
I don't foresee anything but the simplest CLI utils to use this feature though.

kafka/consumer.go
```diff
@@ -395,6 +395,24 @@ func (c *Consumer) QueryWatermarkOffsets(topic string, partition int32, timeoutM
...
+//
+// The timestamps to query are represented as `.Offset` in the `times`
+// argument and the looked up offsets are represented as `.Offset` in the returned
+// `offsets` list.
+//
+// The function will block for at most timeoutMs milliseconds.
+//
+// Duplicate Topic+Partitions are not supported.
+// Per-partition errors may be returned in the `.Error` field.
+func (c *Consumer) OffsetsForTimes(times []TopicPartition, timeoutMs int) (offsets []TopicPartition, err error) {
```
mhowlett: Completely confusing having timestamps in a type called `TopicPartition`. 

The most natural data structure here is a map of TopicPartition : long, which is what the java client does. In dotnet I use an array also for consistency with the rest of the dotnet API, but on my todo list there is to change this where appropriate across the API.

mhowlett: but not suggesting any changes here as I expect the above is consistent with the rest of the API.

kafka/config.go
mhowlett: ahh, ok - sounds good. FWIW I'd be in favor of depreciating then removing this, but argument is good the other way, and not a big deal.

kafka/producer.go
```diff
@@ -405,3 +435,21 @@ func (p *Producer) GetMetadata(topic *string, allTopics bool, timeoutMs int) (*M
...
+//
+// The timestamps to query are represented as `.Offset` in the `times`
+// argument and the looked up offsets are represented as `.Offset` in the returned
+// `offsets` list.
+//
+// The function will block for at most timeoutMs milliseconds.
+//
+// Duplicate Topic+Partitions are not supported.
+// Per-partition errors may be returned in the `.Error` field.
+func (p *Producer) OffsetsForTimes(times []TopicPartition, timeoutMs int) (offsets []TopicPartition, err error) {
```
mhowlett: Looks good.... but since we're in a language that thinks is-a relationships are so bad it doesn't even support them :-), I'll give my usual speil:

`OffsetsForTimes` is not really a method I like having on something called a 'Producer', which I think should be a class very focussed stuff you want to do when producing. In an ideal world, I think it'd be great if librdkafka allowed for creating a client that could be used after construction for both consumption and production (though specific functionality for doing these tasks would be separate), this concept could then be exposed separately in clients, producer / consumer / admin classes could contain (has-a relationship) a client, and a specific client instance could be re-used.

FWIW, in dotnet we just put this on the consumer... but given the inheritance based object modeling, it does make more sense to have this on the producer as well, so this note is not a call to change anything.

kafka/producer.go
```diff
@@ -84,27 +89,52 @@ func (p *Producer) produce(msg *Message, msgFlags int, deliveryChan chan Event)
...
+	//     in the call to the C function.
+	//
+	// Solution:
+	//  Keep track of whether the Value or Key were nil (1), but let the valp and keyp pointers
+	//  point to a 1-byte slice (but the length to send is still 0) so that the dereference (2)
+	//  works.
+	//  Then perform the unsafe.Pointer() on the valp and keyp pointers (which now either point
+	//  to the original msg.Value and msg.Key or to the 1-byte slices) in the call to C (3).
+	//
+	var valp []byte
```
mhowlett: Worked through this logic and didn't spot any problems... It'd be good to have integration tests covering 0, 1, and 2 byte arrays (I think there is not good integration test coverage?), but probably not necessarily highest priority.

mhowlett:
travis seems unhappy, but otherwise LGTM...

PR #90, opened 2017-08-25
producer: Make events channel size configurable

### LINKED ISSUE #87: Make producer's event channel size configurable
In [rafka](https://github.com/skroutz/rafka), we tend to create a producer per client connection. We also sometimes create short-lived producers, that produce a single message and then are destroyed.

confluent-kafka-go sets the events channel buffer size for each producer to a fixed `1000000` (1M) and doesn't provide a way for the client to configure this: https://github.com/confluentinc/confluent-kafka-go/blob/9f5af4225f71044489783d9726429e8d911304f0/kafka/producer.go#L319

This, combined with https://github.com/golang/go/issues/16930 can be problematic in the setup mentioned above. In our case it results in about 2GB of RSS not being handled back to the OS. The problem could be mitigated, although not solved, if the client could control the channel size.

~~We could for example reuse the `go.produce.channel.size` value  when creating the `p.events` channel.~~

agis:
Closes #87

### DIFF
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -273,6 +273,7 @@ func (p *Producer) Close() {
 //                                     These batches do not relate to Kafka message batches in any way.
 //   go.delivery.reports (bool, true) - Forward per-message delivery reports to the
 //                                      Events() channel.
+//   go.events.channel.size (int, 1000000) - Events() channel size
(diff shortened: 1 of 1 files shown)


agis:
[clabot:check]

agis:
[clabot:check]

PR #101, opened 2017-10-06
move test helpers that need testing pkg to _test.go file

### LINKED ISSUE #51: 'testing' package is imported in non-test file
Given the following program:

```go
package main

import (
	"flag"

	"github.com/confluentinc/confluent-kafka-go/kafka"
)

func main() {
	flag.Parse()
	kafka.NewProducer(&kafka.ConfigMap{"bootstrap.servers": "foo"})
}
```

we get the following help output:

```shell
$ ./foo --help
Usage of ./foo:
  -test.bench regexp
        run only benchmarks matching regexp
  -test.benchmem
        print memory allocations for benchmarks
  -test.benchtime d
        run each benchmark for duration d (default 1s)
  -test.blockprofile file
        write a goroutine blocking profile to file
  -test.blockprofilerate rate
        set blocking profile rate (see runtime.SetBlockProfileRate) (default 1)
  -test.count n
        run tests and benchmarks n times (default 1)
  -test.coverprofile file
        write a coverage profile to file
  -test.cpu list
        comma-separated list of cpu counts to run each test with
  -test.cpuprofile file
        write a cpu profile to file
  -test.memprofile file
        write a memory profile to file
  -test.memprofilerate rate
        set memory profiling rate (see runtime.MemProfileRate)
  -test.mutexprofile string
        write a mutex contention profile to the named file after execution
  -test.mutexprofilefraction int
        if >= 0, calls runtime.SetMutexProfileFraction() (default 1)
  -test.outputdir dir
        write profiles to dir
  -test.parallel n
        run at most n tests in parallel (default 4)
  -test.run regexp
        run only tests and examples matching regexp
  -test.short
        run smaller test suite to save time
  -test.timeout d
        fail test binary execution after duration d (0 means unlimited)
  -test.trace file
        write an execution trace to file
  -test.v
        verbose: print additional output
```

This is because the `testing` package is imported by [`testhelpers.go`](https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/testhelpers.go#L23), so any program that imports the driver also imports the `testing` package effectively.

The solution would be for the testing helpers to be moved to a separate package which is then imported in the `_test.go` files.

confluent-kafka-go v0.9.4

gwilym:
intended to fix #51

### DIFF
--- a/kafka/testhelpers.go
+++ b/kafka/testhelpers.go
@@ -20,8 +20,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"testing"
-	"time"
 )
 
 /*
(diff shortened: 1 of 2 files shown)


gwilym:
[clabot:check]

edenhill:
Looks good, thanks!

What do you think of renaming testhelpers.go to testhelpers_test.go?

gwilym:
I got a compile error when I did that, something about using the `C` package in tests being disallowed. I might have missed something though, give it a shot perhaps.

PR #120, opened 2018-02-01
Message headers support

edenhill:


examples/consumer_example/consumer_example.go
```diff
@@ -74,6 +74,9 @@ func main() {
 			case *kafka.Message:
 				fmt.Printf("%% Message on %s:\n%s\n",
 					e.TopicPartition, string(e.Value))
+				if e.Headers != nil {
```
mhowlett: In C#, I currently return an empty collection rather than null when there are no headers, which has the advantage that it's a bit more convenient to use (null checks are not necessary) and consistency with Java (IIRC). downside is obviously memory usage / perf when there are no headers.

kafka/consumer.go
```diff
@@ -261,6 +263,58 @@ func (c *Consumer) Events() chan Event {
...
+// The call will block for at most `timeoutMs` waiting for
+// a new message or error. `timeoutMs` may be set to -1 for
+// indefinate wait.
+//
+// Timeout is returned as (nil, err) where err is `kafka.(Error).Code == Kafka.ErrTimedOut`.
+//
+// Messages are returned as (msg, nil),
+// while general errors are returned as (nil, err),
+// and partition-specific errors are returned as (msg, err) where
+// msg.TopicPartition provides partition-specific information (such as topic, partition and offset).
```
mhowlett: all the above seems nice and go like to my un-trained eye

kafka/consumer.go
```diff
@@ -261,6 +263,58 @@ func (c *Consumer) Events() chan Event {
...
+//
+// All other event types, such as PartitionEOF, AssignedPartitions, etc, are silently discarded.
```
mhowlett: this worries me a bit - it'd be preferable to have a single api that can be used for this purpose, but which you can also subscribe to the other event types if required. if that's possible (using channels or something?) i think you don't really want to introduce this.

kafka/consumer_test.go
```diff
@@ -32,9 +33,11 @@ func TestConsumerAPIs(t *testing.T) {
 	}
 
 	c, err = NewConsumer(&ConfigMap{
-		"group.id":           "gotest",
-		"socket.timeout.ms":  10,
-		"session.timeout.ms": 10})
+		"group.id":                 "gotest",
```
mhowlett: if you make this a guid you never have to worry about the state of the consumer group when re-running the test.

examples/consumer_example/consumer_example.go
edenhill: Noted.
I think nil checks are pretty typical for Go though, and collection methods, such as len(), work on nil slices as well.

kafka/consumer.go
edenhill: This is a high-level convenience interface.
The thing you are describing is what the current Poll() API combined with a switch statement already provides

kafka/consumer_test.go
edenhill: This is a unit-test, it does not interact with any brokers.

examples/consumer_example/consumer_example.go
mhowlett: sounds good to me

kafka/header.go
```diff
@@ -0,0 +1,67 @@
...
+// Message headers are made up of a list of Header elements, retaining their original insert
+// order and allowing for duplicate Keys.
+//
+// Key is a human readable string identifying the header.
+// Value is the key's binary value, Kafka does not put any restrictions on the format of
+// of the Value but it should be made relatively compact.
+// The value may be a byte array, empty, or nil.
+//
+// NOTE: Message headers are not available on producer delivery report messages.
+type Header struct {
```
mhowlett: it seems this is a current librdkafa limitation, right?

in the C# case, I have code in to retrieve the headers for delivery reports, but always get back an empty list.

edenhill: Yep, I didn't think it added much value to retain headers to the delivery report.
I think we're fine documenting this fact until a proper use-case emerges, what do you think?

mhowlett: I would have probably erred the other way since I usually prefer to optimize on having default behavior that doesn't surprise, but I don't think this is a big deal and I may be mis-prioritizing arguments the other way.

edenhill: Well, once functionality is added we can't remove or alter it, so I'd rather wait for a use-case.

mhowlett:
ok, done with the comments. still concerned `ReadMessage` represents something that isn't the best way of doing things, but lack greater context to really know or comment further without a fair bit of effort. fine for you to merge whenever.

edenhill:
ReadMessage was added after a discussion with @travisjeffery about our complex API, it was added as a way for new users to quickly get up and running without having to care or worry about the other event types.
As these users progress in their development cycle they will most likely migrate to the richer more complex API to be able to handle various events and errors in a meaningful way.

kafka/header.go
edenhill: Oh, to be clear, headers are available in the delivery report callback in librdkafka, it is just in the Go bindings that I've skipped them since there is a performance penalty to each C call.
We'll add support for this when a use-case emerges from the community.

mhowlett: interesting - i thought I saw this wasn't working. Yeah, I think excluding this is very justifiable. In C# I expect I'll consider the degree of overhead annoying but acceptable, and will likely include it (but will benchmark).

kafka/consumer.go
```diff
@@ -261,6 +263,58 @@ func (c *Consumer) Events() chan Event {
...
+//
+func (c *Consumer) ReadMessage(timeoutMs int) (*Message, error) {
```
travisjeffery: any reason not to take a `timeout time.Duration` instead of an int?

edenhill: I can't seem to find a way to specify an infinite timeout with a Duration, and it is not nillable.
Ideas?

mhowlett: should anyone really be using an infinite timeout in practice? (i.e. is this call cancellable by some mechanism?)

travisjeffery: The way it works now is 0 is infinite right? Could still work the same way. For example:

```
func main() {
	fn(0)
}

func fn(timeout time.Duration) {
	fmt.Printf("timeout: %v, %v", timeout, timeout == 0)
}

# >> timeout: 0s, true
```

edenhill: 0 is no timeout, -1 is indefinate, > 0 is a standard timeout.
Same semantics as poll(2)

travisjeffery: If people wanted the timeout they'd generally call it by doing something like time.Second * 5

edenhill: Okay, so we should keep the current semantics with -1, 0, > 0, but use time.Duration instead?

travisjeffery: Yeah

edenhill:
Please review the examples added to the README

README.md
```diff
@@ -31,6 +31,92 @@ See the [API documentation](http://docs.confluent.io/current/clients/confluent-k
...
+	})
+
+	if err != nil {
+		panic(err)
+	}
+
+	c.SubscribeTopics([]string{"myTopic", "^aRegex.*[Tt]opic"}, nil)
+
+	for true {
+		msg, err := c.ReadMessage(-1)
```
mhowlett: should this be a duration?

travisjeffery: go coerces it automatically

travisjeffery: if you'd had set the assigned the time `var i int = -1` you'd have to convert it yourself

travisjeffery:
lgtm

PR #123, opened 2018-02-08
Close event channel when consumer is closed

czchen:
This PR let `Consumer` to close `consumer.events` channel when calling `Consumer.Close()`. In this way, we can use the following code to consume data from `Consumer`, and exit automatically when `Consumer.Close()` is called.

```go
for ev := range consumer.Events() {
    // do something here
}
```


edenhill:
Good fix, thanks!

PR #125, opened 2018-02-09
explicit call out that you may need pkg-config

chstout:
'go get' will fail on mac if you don't have pkg-config installed.


edenhill:
Thank you!

PR #144, opened 2018-02-22
Simplify for loop

pavelnikolov:
In Golang `for {` would loop forever and is effectively the same as `for true {`.


pavelnikolov:
ping

edenhill:
[clabot:check]

edenhill:
Thank you for this!

PR #149, opened 2018-02-26
[WIP] Prototype of PollContext

pavelnikolov:
This is just a prototype implementation of `PollContext` method. 

**ASSUMPTION**:  The underlying library would work with `0ms` timeout if deadline is not set. If this is wrong the code needs to be adjusted.

If the underlying library cannot work without a timeout, then the code might panic if one is not set in the context. The method will return `nil` if the context is cancelled or expired.

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -17,6 +17,7 @@ package kafka
  */
 
 import (
+	"context"
 	"fmt"
 	"math"
 	"time"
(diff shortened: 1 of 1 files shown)


pavelnikolov:
@edenhill do you want me to work on this further?

edenhill:
I'm a bit swamped by release work, will get back to you on this within a couple of days.
Sorry for the delay.

KJTsanaktsidis:
@pavelnikolov You might don't care about this after two and a half years, but I think I implemented PollContext with cancellation in this PR: https://github.com/confluentinc/confluent-kafka-go/pull/518

pavelnikolov:
Wow, it's been 4 years since I opened this PR : )

PR #170, opened 2018-04-09
added confluent cloud example

mhowlett:


examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -0,0 +1,106 @@
...
+	value := "golang test value"
+	topic := "go-test-topic"
+	p.Produce(&kafka.Message{
+		TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny},
+		Value:          []byte(value),
+	}, nil)
+
+	e := <-p.Events()
+	
+	m := e.(*kafka.Message)
```
mhowlett: the need for this type assertion seems both ugly and not n00b friendly. wondering why it's necessary. without really looking, I'm guessing there can be more than one type of event. without knowing much about go, wondering why these couldn't be made available on different channels.

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		fmt.Printf("failed to create consumer: %s\n", err)
+		os.Exit(1)
+	}
+
+	topics := []string { topic }
+	c.SubscribeTopics(topics, nil)
+
+	for {
+		msg, err := c.ReadMessage(100 * time.Millisecond)
```
mhowlett: the examples use -1 here, but is there a clean way to break of this method in that case if there are no new events? if not, I think we shouldn't even allow -1 to be specified.

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -0,0 +1,106 @@
...
+		if err == nil {
```
mhowlett: `err` seems to be set even when `ReadMessage` times out. this doesn't seem right

examples/confluent_cloud_example/confluent_cloud_example.go
edenhill: This is how you implement common interfaces in Go, supporting multiple types.
Type switching or type assertions are the way to go, but you need to check that the assertion was ok:

```golang
m, ok := e.(*kafka.Message)
if !ok {
     // not  a message..
```

We've been through this before, but the reason messages and other events are on the same channel is ordering.
The ordering of, let's say PartitionEOF and messages, are significant.

examples/confluent_cloud_example/confluent_cloud_example.go
edenhill: Ctrl-C didn't work?

examples/confluent_cloud_example/confluent_cloud_example.go
edenhill: What seems more right?

README.md
```diff
@@ -114,7 +114,7 @@ func main() {
 }
 ```
 
-More elaborate examples are available in the [examples](examples) directory.
+More elaborate examples are available in the [examples](examples) directory, including [how to configure](examples/confluent_cloud_example) the go client for use with [Confluent Cloud](https://www.confluent.io/confluent-cloud/).
```
edenhill: Capital Go, or Golang

edenhill: Try to keep column width around 80 chars to make it readable in terminals (e.g., split line before "including" and at "for use"

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -0,0 +1,106 @@
+// This is a simple example demonstrating how to produce a message to 
+// Confluent Cloud then read it back again.
+//     
+// https://www.confluent.io/confluent-cloud/
+// 
+// Confluent Cloud does not auto-create topics. You will need to use the ccloud
```
edenhill: "Auto-creation of topics is disabled in Confluent Cloud"

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -0,0 +1,106 @@
+// This is a simple example demonstrating how to produce a message to 
+// Confluent Cloud then read it back again.
+//     
+// https://www.confluent.io/confluent-cloud/
+// 
+// Confluent Cloud does not auto-create topics. You will need to use the ccloud 
+// cli to create the go-test-topic topic before running this example. The
```
edenhill: an example ccloud command line to do this would be great

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -0,0 +1,106 @@
+// This is a simple example demonstrating how to produce a message to 
+// Confluent Cloud then read it back again.
+//     
+// https://www.confluent.io/confluent-cloud/
+// 
+// Confluent Cloud does not auto-create topics. You will need to use the ccloud 
+// cli to create the go-test-topic topic before running this example. The 
+// <ccloud bootstrap servers>, <ccloud key> and <ccloud secret> parameters are
+// available via the confluent cloud web interface. For more information,
```
edenhill: Capital Confluent Cloud

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"time"
+	"fmt"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
+	"os"
+)
+
+func main() {
+
+	p, err := kafka.NewProducer(&kafka.ConfigMap{
```
edenhill: same comments as for the Python client, so not repeating here.

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -0,0 +1,106 @@
...
+		"broker.version.fallback": "0.10.0.0",
+		"api.version.fallback.ms": 0,
+		"sasl.mechanisms": "PLAIN",
+		"security.protocol": "SASL_SSL",
+		"ssl.ca.location": "/usr/local/etc/openssl/cert.pem",
+		"sasl.username": "<ccloud key>",
+		"sasl.password": "<ccloud secret>",})
+
+	if err != nil {
+		fmt.Printf("failed to create producer: %s\n", err)
```
edenhill: Can use `panic("Failed to create producer"...)` to avoid os import

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	value := "golang test value"
+	topic := "go-test-topic"
+	p.Produce(&kafka.Message{
+		TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny},
+		Value:          []byte(value),
+	}, nil)
+
+	e := <-p.Events()
```
edenhill: Add comment what this does, e.g., "Wait for delivery report"

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -0,0 +1,106 @@
...
+	
+	m := e.(*kafka.Message)
+	if m.TopicPartition.Error != nil {
+		fmt.Printf("failed to deliver message: %v\n", m.TopicPartition.Error)
```
edenhill: I think you should print just "m.TopicPartition" to get the TopicPartition info as well.

examples/confluent_cloud_example/confluent_cloud_example.go
mhowlett: ok sounds good

examples/confluent_cloud_example/confluent_cloud_example.go
mhowlett: IIRC, yes it does. I guess i'm sensitive to this since in C# you shouldn't have a method where potentially the only way out of it is via a signal. If you think that's idiomatic golang, that's fine by me (but maybe it's just idiomatic C?). I'm just sort of saying what comes to mind in case it turns out to be useful :-)

examples/confluent_cloud_example/confluent_cloud_example.go
mhowlett: sounds good

examples/confluent_cloud_example/confluent_cloud_example.go
mhowlett: `msg` being `nil`? 
having `err` set under non-error program flow seems counter-intuitive to me. 
again, first reactions (err, well second reactions, seem to remember reviewing that). haven't analyzed closely.

examples/confluent_cloud_example/confluent_cloud_example.go
mhowlett: i was purposefully leaving the `api.version.request` in for safety since i was thinking librdkafka version was not guaranteed and as far as I could tell isn't enforced. but i'm also just as happy to take it out.

examples/confluent_cloud_example/confluent_cloud_example.go
edenhill: :+1: In the long run we want to support Contexts for timeouts, which allows cancellation.

examples/confluent_cloud_example/confluent_cloud_example.go
edenhill: Well, there are two ways to do it and this is one of them. Not sure if there is a right answer.

examples/confluent_cloud_example/confluent_cloud_example.go
edenhill: It's been the default since v0.11.0, so I think we're fine without specifying it explicitly. We'll want people to use latest librdkafka anyway.

PR #173, opened 2018-04-17
Topic Admin API

edenhill:
Note: The incremental option is still WIP since the upstream KIP-248 has not been finalized yet.


kafka/integration_test.go
```diff
@@ -1084,3 +1088,364 @@ func TestProducerConsumerHeaders(t *testing.T) {
...
+		t.Fatalf("Create topic request failed: %v", err)
+	}
+
+	if topicResult[0].Error.Code() != ErrNoError {
+		t.Fatalf("Failed to create topic %s: %s", topic, topicResult[0].Error)
+	}
+
+	// Read back config to validate
+	configResources := []ConfigResource{{Type: ResourceTopic, Name: topic}}
+	describeRes, err := a.DescribeConfigs(ctx, configResources, nil)
```
mhowlett: It would be good to test that `DescribeConfigs` on a topic that doesn't exist behaves as expected (I don't think it does).

kafka/adminapi_test.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	confStrings := map[string]string{
+		"some.topic.config":  "unchecked",
+		"these.are.verified": "on the broker",
+		"and.this.is":        "just",
+		"a":                  "unit test"}
+
+	// Correct input, fail with timeout
+	ctx, cancel := context.WithTimeout(context.Background(), expDuration)
```
travisjeffery: Just curious: Why does this time out? Cause the admin client's broker isn't set or defaults to something like localhost:9092 where no broker is running during this test and there's no response? You'd think the client would fail faster in that case if so.

kafka/adminapi_test.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	if ctx.Err() != context.DeadlineExceeded {
+		t.Fatalf("Expected DeadlineExceeded, not %v", ctx.Err())
+	}
+}
+
+// TestAdminAPIs dry-tests most Admin APIs, no broker is needed.
+func TestAdminAPIs(t *testing.T) {
+
+	a, err := NewAdminClient(&ConfigMap{})
```
travisjeffery: Could DRY up these tests a bit going table-style: https://github.com/golang/go/wiki/TableDrivenTests

kafka/integration_test.go
```diff
@@ -1084,3 +1088,364 @@ func TestProducerConsumerHeaders(t *testing.T) {
...
+
+	validateTopicResult(t, result2, expError)
+}
+
+func validateConfig(t *testing.T, results []ConfigResourceResult, expResults []ConfigResourceResult, checkConfigEntries bool) {
+
+	_, file, line, _ := runtime.Caller(1)
+	caller := fmt.Sprintf("%s:%d", path.Base(file), line)
+
+	if len(results) != len(expResults) {
```
travisjeffery: Could you do all the following using [reflect.DeepEqual](https://golang.org/pkg/reflect/#DeepEqual)?

kafka/integration_test.go
travisjeffery: 👍

kafka/integration_test.go
```diff
@@ -1084,3 +1088,364 @@ func TestProducerConsumerHeaders(t *testing.T) {
...
+		if (i % 1) == 0 {
+			newTopics[i].ReplicationFactor = len(fewReplicas)
+		} else {
+			newTopics[i].ReplicationFactor = len(manyReplicas)
+		}
+
+		expError[newTopics[i].Topic] = Error{} // No error
+
+		var useReplicas []int32
+		if i == 2 {
```
travisjeffery: I'd probably use a switch for > 2 cases but not a big deal.

travisjeffery:
Would be good to have a ListTopics API so you see which topics exist already.

travisjeffery:
Whoops

kafka/adminapi_test.go
edenhill: While the Go API does not expose it, librdkafka allows manually adding brokers after client instance inception (such as learnt from ZK or other external system), so even if no brokers are available at the time the admin API is called the admin state machine will wait for up to AdminOptions.RequestTimeout to send the request, handling broker connectivity issues, controller changes, and what not.

So yes, in this test we rely on the context to timeout before the actual underlying admin operation in librdkafka times out.

kafka/adminapi_test.go
edenhill: I'll add that to the Future Fun list :)

kafka/integration_test.go
edenhill: expResult.Config only contains the config entries we changed, while result.Config contains the full broker/topic config - so we would have to prepopulate expResult.Config and overwrite with our changes in that case, that would be more error prone.

kafka/integration_test.go
edenhill: Yah, this is tested in librdkafka and it exposed a bug in the broker where it would return the default config for unknown topics, rather than an error.

edenhill:
@travisjeffery We have GetMetadata(), which is the same thing with a legacy name.
I'm adding it to the AdminClient for now, we do want to add ListTopics() too but it requires some work in librdkafka to make an extensible API.

edenhill:
@travisjeffery Are you satisfied with the public API?

examples/admin_create_topic/admin_create_topic.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// Contexts are used to abort or limit the amount of time
+	// the Admin call blocks waiting for a result.
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Create topics on cluster.
+	// Set Admin options to wait for the operation to finish (or at most 60s)
+	maxDur, err := time.ParseDuration("60s")
+	if err != nil {
```
mhowlett: i wouldn't check for / handle bugs in standard go libraries

examples/admin_create_topic/admin_create_topic.go
```diff
@@ -0,0 +1,93 @@
...
+		panic("ParseDuration(60s)")
+	}
+	results, err := a.CreateTopics(
+		ctx,
```
mhowlett: Just wanted to check how sure you are that a blocking call with `ctx` as the first argument is the most appropriate API for this in golang. This is a n00b question, just calling out something I was wondering about: having a context parameter makes usage of the api more verbose than it otherwise needs to be for those who don't want it. On the other hand, it does look like it's a new feature and may be an impending standard sort of thing to do?. But standard file io stuff doesn't seem to do this, and there seem to be standard patterns for using go routines to asyncify synchronous methods. Is that not good enough?

kafka/adminapi.go
```diff
@@ -0,0 +1,964 @@
...
+
+	return a.cConfigResourceToResult(cResults, cCnt)
+}
+
+// GetMetadata queries broker for cluster and topic metadata.
+// If topic is non-nil only information about that topic is returned, else if
+// allTopics is false only information about locally used topics is returned,
+// else information about all topics is returned.
+// GetMetadata is equivalent to listTopics, describeTopics and describeCluster in the Java API.
+func (a *AdminClient) GetMetadata(topic *string, allTopics bool, timeoutMs int) (*Metadata, error) {
```
mhowlett: no context parameter here. is there a plan for consistency going forward?

examples/admin_create_topic/admin_create_topic.go
edenhill: The introduction of contexts was after a discussion with @travisjeffery, it seems to be the way forward for the Go community and we should follow suite.

Contexts provides a standardised way to limit or cancel long running blocking operations.

kafka/adminapi.go
edenhill: GetMetadata() is an existing API on the Producer and Consumer, so it is copied verbatim to the AdminClient for consistency.

In the long run we'll want to add a ListTopics(), which should then take a context.

examples/admin_create_topic/admin_create_topic.go
edenhill: Right, my intention was for this example to be a nice cut'n'paste template for our users, which might want to pass the timeout as user input.
And there is no harm in a bit of extra error checking

mhowlett: there's harm in that it makes the example longer and some people may 'learn' that they should do this in this scenario. my opinion is still the other way, but minor detail.

examples/admin_create_topic/admin_create_topic.go
mhowlett: it's a pity that in golang this convention necessarily add verbosity - in other languages this is not the case. but it seems the right choice as far as i can tell.

kafka/testhelpers.go
```diff
@@ -125,3 +125,18 @@ func getMessageCountInTopic(topic string) (int, error) {
 
 	return cnt, nil
 }
+
+// getBrokerList returns a list of brokers (ids) in the cluster
+func getBrokerList(H Handle) (brokers []int32, err error) {
+	md, err := getMetadata(H, nil, true, 15*1000)
```
mhowlett: this timeout is not a `Duration`... although arguably a small thing, people will use detail like this as a proxy for their opinion of the entire library - getting them right is very important (yep, i know this is an existing decision, but pointing it out anyway). And using an int for a timeout is a bad idea (though writing it as 15*1000 is a great idea).

kafka/adminapi.go
```diff
@@ -0,0 +1,964 @@
...
+		return
+	}
+
+	a.handle.cleanup()
+
+	C.rd_kafka_destroy(a.handle.rk)
+}
+
+// NewAdminClient creats a new AdminClient instance with a new underlying client instance
+func NewAdminClient(conf *ConfigMap) (*AdminClient, error) {
```
mhowlett: I don't understand why a pointer to a `ConfigMap` is used here (which is a pointer to a runtime.hmap structure), rather than just `ConfigMap`.

mhowlett: More commentary on `ConfigMap`:

`SetKey` is a confusing name for that method. Also, in addition to the `SetKey` method, the map indexer is public and I would expect them to do the same thing, but they don't (right?).

`Get` would be better named `GetOrDefault` or something. Again, different operation is different from the map indexer.

Having the specific use case centric `Set` method on `ConfigMap` sets a precedent for other miscellaneous config helper methods being put on this type, which I think is questionable. It should also be called something more descriptive of what it does (one would expect set to be something else).

kafka/adminapi.go
```diff
@@ -0,0 +1,964 @@
...
+		a.handle = &handle{}
+		return
+	}
+
+	a.handle.cleanup()
+
+	C.rd_kafka_destroy(a.handle.rk)
+}
+
+// NewAdminClient creats a new AdminClient instance with a new underlying client instance
```
mhowlett: creats doesn't have a good ring to it

examples/admin_create_topic/admin_create_topic.go
```diff
@@ -0,0 +1,93 @@
...
+		// Multiple topics can be created simultaneously
+		// by providing more NewTopic structs here.
+		[]kafka.NewTopic{{
+			Topic:             topic,
+			NumPartitions:     numParts,
+			ReplicationFactor: replicationFactor}},
+		// Admin options
+		kafka.SetAdminOperationTimeout(maxDur))
```
mhowlett: why not just keep it simple and just use a plain old `CreateTopicsOptions` struct for this? the way this works atm is intricate enough that I need to go investigate `SetAdminOperationTimeout` to understand what it means to be 'setting' something in the context of a method argument. I don't think there's any benefit to trying to have fancy modeling of the the options structure, but there is benefit to having it straightforward.

examples/admin_create_topic/admin_create_topic.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		fmt.Printf("Failed to create topic: %v\n", err)
+		os.Exit(1)
+	}
+
+	// Print results
```
mhowlett: ...and there i was thinking this was calculating the value of Pi.

examples/admin_create_topic/admin_create_topic.go
mhowlett: i should say that it is an interesting and quite nice solution you've come up with here, which works effectively, but if it's not familiar to people using the api, the extra complexity is a detractor. if it's idiomatic go, you can ignore the above comment.

edenhill: The problem with using options structs is that uninitialized fields will be set to their "zero value", which is 0 for most data types, but 0 might not actually be the default value of certain options and may have specific meaning.
One way around that is to use a factory function, but since there are no default-args for Go functions those factory functions would need to take all options as arguments and then we're back to the same problem with zero value.
Another way is to use pointers, but that will require temporary variables, allocations or a set of helper functions to provide int/bool/whatnot arguments in pointer form, and that just becomes messy.

I discussed this with @travisjeffery and he suggested the solution that is used here, which I think also reads quite nice.

As for UX:  this solution should actually be easier on the user since if they dont want options (which most people probably wont start with) then they don't have to care about options at all but simply omit that argument.

examples/admin_create_topic/admin_create_topic.go
edenhill: compartmentalizing the code layout!

kafka/adminapi.go
edenhill: I don't remember the historical reason why ConfigMap is a pointer, but we need to be consistent with the existing Producer and Consumer client types.

As for the ConfigMap methods: The Set() was added to make it compatible and simple to use with a popular command line parsing module, which was a questionable decision.

Since Set() was already taken, SetKey() was added.

Anyway, these are part of the existing API so they will not be changed here.

kafka/testhelpers.go
edenhill: Contexts did not exist when getMetadata() was added.

examples/admin_create_topic/admin_create_topic.go
edenhill: Yeah, Golang seems to be quite explicit and verbose in these regards.

examples/admin_create_topic/admin_create_topic.go
edenhill: I disagree, generally people should check the error value, it is good practice.

As for example length, we could skip this option completely in this example to save some lines, but I believe a proper example of an option is important.

examples/admin_create_topic/admin_create_topic.go
mhowlett: thanks for the commentary - yes, this looks good to me. I'm increasingly of the opinion that golang makes the wrong tradeoffs, but that's a separate issue.

kafka/testhelpers.go
mhowlett: `Duration` did though...

kafka/adminapi.go
```diff
@@ -0,0 +1,964 @@
...
+	//        settings in the future (which happened with AlterConfigs)
+	//        we must make a breaking change or have two fields:
+	//        Config, ConfigEntries.
+	//        The StringMapToConfigEntries(, AlterOperationSet) can be
+	//        used in the meantime.
+	Config []ConfigEntry
+}
+
+// DeleteTopic holds parameters for deleting a topic.
+type DeleteTopic struct {
```
mhowlett: is this struct necessary? the java api uses just a string. In addition to thinking this is over-future-proofing, I don't really like how this reads when using the api.

examples/admin_create_topic/admin_create_topic.go
```diff
@@ -0,0 +1,93 @@
...
+	// Set Admin options to wait for the operation to finish (or at most 60s)
+	maxDur, err := time.ParseDuration("60s")
+	if err != nil {
+		panic("ParseDuration(60s)")
+	}
+	results, err := a.CreateTopics(
+		ctx,
+		// Multiple topics can be created simultaneously
+		// by providing more NewTopic structs here.
+		[]kafka.NewTopic{{
```
mhowlett: Not only would `TopicSpecification` be a better name for `NewTopic`, that name also wouldn't create confusion with what seems to be a golang convention of having constructor methods named `New<Something>`. The name `NewTopic` is somewhat susceptible to similar confusion in other languages, but in golang it actually conflicts with an actual convention.

kafka/adminapi.go
mhowlett: my argument is the java people are very paranoid about this as well, thought it through, and chose to use string.

examples/admin_create_topic/admin_create_topic.go
mhowlett: so to confirm, we're avoiding possibly confusing a minority of users who have exposure to the java api (by doing something that isn't at all confusing) by doing something that clashes with a golang convention. awesome

kafka/adminapi.go
edenhill: That's a good point. I'll verify with the core people that they are still happy with this choice

examples/admin_create_topic/admin_create_topic.go
edenhill: It is not mainly about confusing users, but making existing documentation useful and providing the same API semantics, naming, etc, regardless of language - i.e., a coherent product experience.

I do agree that kafka.NewTopic is suboptimal from a Go perspective.

kafka/adminapi.go
```diff
@@ -0,0 +1,964 @@
...
+	// Value of configuration entry.
+	Value string
+	// Source indicates the configuration source.
+	Source ConfigSource
+	// IsReadOnly indicates whether the configuration entry can be altered.
+	IsReadOnly bool
+	// IsSensitive indicates whether the configuration entry contains sensitive information, in which case the value will be unset.
+	IsSensitive bool
+	// IsSynonym indicates whether the configuration entry is a synonym for another configuration property.
+	IsSynonym bool
```
mhowlett: can this be derived by comparing the size of the `Synonyms` map to 1 (i.e. if a is a synonym of b then technically b is also a synonym of a) or is it the case that there are canonical names for which this property is never set to true (even if there is a synonym for the canonical name). If the former, I think this property should be removed. If the latter, i think this needs to be explicit in the property docs.

kafka/adminapi.go
```diff
@@ -0,0 +1,964 @@
...
+		Syn := configEntryResultFromC(cSyn)
+		entry.Synonyms[Syn.Name] = Syn
+	}
+
+	return entry
+}
+
+// ConfigResourceResult provides the result for a resource from a AlterConfigs or
+// DescribeConfigs request.
+type ConfigResourceResult struct {
```
mhowlett: In java, alter configs doesn't include a config entry map in the result (only describe configs does). i'd be inclined to have two structs in the public api rather than reusing the one common one. this would match java more closely, and the naming is more straightforward.

kafka/adminapi.go
mhowlett: maybe an integration test for this would be good.

mhowlett:
ok, finished with my comments - feel free to do with them as you see fit, then merge.

kafka/adminapi.go
```diff
@@ -0,0 +1,964 @@
...
+	// if an explicit ReplicaAssignment is set.
+	ReplicationFactor int
+	// (Optional) Explicit replica assignment. The outer array is
+	// indexed by the partition number, while the inner per-partition array
+	// contains the replica broker ids. The first broker in each
+	// broker id list will be the preferred replica.
+	ReplicaAssignment [][]int32
+	// Topic configuration.
+	// FIXME: Not sure about this one, a simple map[string]string
+	//        would be more convenient, but if the protocol adds per-entry
```
mhowlett: i can't seem to find any info on per-entry settings. the only thing i found is the incremental flag in KIP-248. i currently have this as `map[string]string` in dotnet as well so very interested to have thought through the future a bit more. in java it's a string string map as well, but it's a function call which could be overloaded. golang, and the property convention i'm using in dotnet don't have that luxury.

kafka/adminapi.go
```diff
@@ -0,0 +1,964 @@
...
+}
+
+// String returns a human-readable representation of a ConfigResource
+func (c ConfigResource) String() string {
+	return fmt.Sprintf("Resource(%s, %s)", c.Type, c.Name)
+}
+
+// AlterOperation specifies the operation to perform on the ConfigEntry.
+// One of AlterOperationAdd, AlterOperationSet, AlterOperationDelete.
+type AlterOperation int
```
mhowlett: i don't understand why this is necessary (or desirable) for implementing KIP-248. is there something else I'm missing?

kafka/adminapi.go
edenhill: Synonyms is kind of a bad word for this, synonyms not only depicts actual synonyms (e.g., aliases) but also the same configuration property configured through other means. So when you set a property it will get a synonym pointing to the default-config property and value.

I'm not really sure what the value of these are from a user perspective, so I'm just going to stick with what the Java API does.

kafka/adminapi.go
edenhill: Another related question, all these APIs take and return a list of objects, rather than a map which Py and Java does: should we change it to a map to make it easier (well.., sometimes) to look up the result for each element?

kafka/adminapi.go
edenhill: You're probably right, map[string]string should be sufficient. I'll change it.

kafka/adminapi.go
mhowlett: in the .NET API, I took the view that where you're likely to want to iterate through the entire collection (which you're likely to want to do with the result classes), a list was better. In the case of individual config entries returned from describe configs, a dictionary is better as you're more likely to want to do specific lookups.

kafka/adminapi.go
mhowlett: yeah, consistency (re ConfigMap) is most important here. But I wanted to point out that for me, this was a significant point of friction. I needed to dig deeper since it's something that went against my intuition - is there a reason for this? if not, why is this like this? i see it as very important that new users to the library aren't immediately presented with something that seems off, or they need to question (and this is one of the first things new users of the library will be exposed to).

kafka/adminapi.go
```diff
@@ -193,28 +193,19 @@ func (c ConfigResource) String() string {
...
 	AlterOperationSet = iota
-	// AlterOperationDelete deletes the configuration setting,
-	// possibly reverting it to default.
-	AlterOperationDelete = iota
 )
 
 // String returns the human-readable representation of an AlterOperation
 func (o AlterOperation) String() string {
 	switch o {
-	case AlterOperationAdd:
```
mhowlett: I think you should take the AlterOperation property out of the ConfigEntry struct altogether.

edenhill: I was tempted to do so too, but Colin's upcoming KIP-339 actually adds a pretty much identical thing, and keeping this type in the Go client makes the ConfigEntry interface forwards compatible.

edenhill:
@mhowlett @rnpridgeon 
Please review this commit:
https://github.com/confluentinc/confluent-kafka-go/pull/173/commits/9e3abddb55fd67d13e1696596c9a9d82c25eda95

mhowlett:
https://github.com/confluentinc/confluent-kafka-go/commit/9e3abddb55fd67d13e1696596c9a9d82c25eda95 lgtm. a suggested grammatical change and some commentary on `TotalCount` vs `IncreaseTo`, but either is fine by me.

PR #179, opened 2018-04-30
Use same scheme as the HTML document in docs

dtjm:
Some browsers may not pull unencrypted resources that are referenced from encrypted pages. Which causes the styling to be broken on https://docs.confluent.io/current/clients/confluent-kafka-go/index.html in Chrome.

https://www.dropbox.com/s/zzcj5axukisyvk6/Screenshot%202018-04-30%2011.01.03.png?dl=0


dtjm:
[clabot:check]

edenhill:
Thank you!

Issue #183, opened 2018-05-10
Memory leak when the producer uses Headers


lintang0:
Description
===========
When the producer sends kafka message with "Headers", the memory usage keeps going up. No memory leak if no Headers inside the message.

How to reproduce
================
run the following code, which is modified from the sample producer code, memory leak can be reprodued.

    package main

    /**
      Memory leak detected when Headers inside message
    */

    import (
	"fmt"
	"os"

	"github.com/confluentinc/confluent-kafka-go/kafka"
    )

    func main() {

	if len(os.Args) != 3 {
		fmt.Fprintf(os.Stderr, "Usage: %s <broker> <topic>\n",
			os.Args[0])
		os.Exit(1)
	}

	broker := os.Args[1]
	topic := os.Args[2]

	p, err := kafka.NewProducer(&kafka.ConfigMap{"bootstrap.servers": broker})

	if err != nil {
		fmt.Printf("Failed to create producer: %s\n", err)
		os.Exit(1)
	}

	fmt.Printf("Created Producer %v\n", p)
	value := "Hello Go!"

	for {
		err = p.Produce(&kafka.Message{
			TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny},
			Value:          []byte(value),
			Headers:        []kafka.Header{{"myTestHeader", []byte("header values are binary")}},
		}, nil)
	}
}
Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-go and librdkafka version (`LibraryVersion()`): 0.11.4
 - [x] Apache Kafka broker version: 2.11 confluent-3.2.2
 - [ ] Client configuration: `ConfigMap{...}`
 - [x] Operating system: Ubuntu 14.04
 - [ ] Provide client logs (with `"debug": ".."` as necessary)
 - [ ] Provide broker log excerpts
 - [x] Critical issue

billygout:
Hi @edenhill. I'm working with @lintang0 here. I just wanted to check that the information here is sufficient for debugging purposes. Thanks!

mhowlett:
makes me wonder - have we ever tried testing librdkafka with valgrind?

billygout:
@mhowlett I see valgrind mentioned here: https://github.com/edenhill/librdkafka/tree/master/tests
[...]

PR #184, opened 2018-05-14
Fix headers memory leak #183

billygout:


billygout:
[clabot:check]

edenhill:
Great stuff, thanks!

PR #202, opened 2018-06-12
defer p.Close() in producer, avoid memory leak

dewtx29:
Fixed bug memory leak (important bug !)

by closing Producer


edenhill:
Thank you!

PR #219, opened 2018-08-07
Fix so v0.9.2.x library will build

yacn:
### changes

Our kafka infrastructure is currently on 0.10.1, thus I needed to use the older version of this library.

This changeset fixes a type error so that when you run `go build` using this library on the `0.9.2.x` branch, it actually builds.

### testing

Running `go build` before this patch:

```
➜ dep ensure # to overwrite my local changes
➜ go build
# github.wouldntyouliketoknow.com/eng/kafkacat-go/vendor/github.com/confluentinc/confluent-kafka-go/kafka
vendor/github.com/confluentinc/confluent-kafka-go/kafka/config.go:131: cannot use (*_Ctype_struct_rd_kafka_conf_s)(ctopicConf) (type *_Ctype_struct_rd_kafka_conf_s) as type *_Ctype_struct_rd_kafka_topic_conf_s in argument to func literal
```

After the patch:

```
➜ cat confluent-kafka-go.patch
--- vendor/github.com/confluentinc/confluent-kafka-go/kafka/config.go	2018-08-07 11:30:50.000000000 -0700
+++ vendor/github.com/confluentinc/confluent-kafka-go/kafka/config.go.updated	2018-08-07 11:31:19.000000000 -0700
@@ -128,7 +128,7 @@
 }

 func (ctopicConf *rdkTopicConf) set(cKey *C.char, cVal *C.char, cErrstr *C.char, errstrSize int) C.rd_kafka_conf_res_t {
-	return C.rd_kafka_topic_conf_set((*C.rd_kafka_conf_t)(ctopicConf), cKey, cVal, cErrstr, C.size_t(errstrSize))
+	return C.rd_kafka_topic_conf_set((*C.rd_kafka_topic_conf_t)(ctopicConf), cKey, cVal, cErrstr, C.size_t(errstrSize))
 }

 func configConvertAnyconf(m ConfigMap, anyconf rdkAnyconf) (err error) {
➜ patch -b -p0 < confluent-kafka-go.patch
➜ go build -a
➜ # built successfully
```

### DIFF
--- a/kafka/config.go
+++ b/kafka/config.go
@@ -128,7 +128,7 @@ func (cConf *rdkConf) set(cKey *C.char, cVal *C.char, cErrstr *C.char, errstrSiz
 }
 
 func (ctopicConf *rdkTopicConf) set(cKey *C.char, cVal *C.char, cErrstr *C.char, errstrSize int) C.rd_kafka_conf_res_t {
-	return C.rd_kafka_topic_conf_set((*C.rd_kafka_conf_t)(ctopicConf), cKey, cVal, cErrstr, C.size_t(errstrSize))
+	return C.rd_kafka_topic_conf_set((*C.rd_kafka_topic_conf_t)(ctopicConf), cKey, cVal, cErrstr, C.size_t(errstrSize))
 }
 
 func configConvertAnyconf(m ConfigMap, anyconf rdkAnyconf) (err error) {


yacn:
[clabot:check]

edenhill:
librdkafka is backwards compatible with any broker version >= 0.8, so there is no need to use an older librdkafka with an older cluster.

The Go client has a rather strict version coupling with librdkafka, requiring an up to date version.

yacn:
You'd rather have a broken version hanging around (why bother keeping the branch then?) than merge a one line fix?

yacn:
Why even keep a broken branch around then?

edenhill:
There is little point in supporting old versions of librdkafka, as a user you would want to use the latest version to make sure you get all the functionality and fixes.
v0.9.2, for instance, is almost two years old and completely unsupport at this point.
https://github.com/edenhill/librdkafka/releases

yacn:
As far as I was aware due to our infra here we were needing to use librdkafka 0.9.2 explicitly.

I've since tested out the latest librdkafka (from `brew install`) and this library (from master) and everything seems to work just fine. This makes things _a lot_ simpler (I have a gist w/ the old homebrew formula for librdkafka, pinning it, having to build inside docker for specific versioning, etc), so thank you for the TIL.

edenhill:
Happy to hear that!
For broker compat, see this: https://github.com/edenhill/librdkafka/wiki/Broker-version-compatibility

Issue #225, opened 2018-09-03
order-dependent configuration failure with plugin.library.paths monitoring-interceptor


dtheodor:
Any plugin-specific configuration properties fail to be set if the plugin has not been registered with the property `"plugin.library.paths"`, with an error that looks like this

```sh
panic: No such configuration property: "confluent.monitoring.interceptor.icdebug"
```

However, there's no way to specify ordering of configuration properties through a `ConfigMap`, which is just a map with keys that are returned in random order. When instantiating a client, there's a loop over the map's keys, and each of them is set through `C.rd_kafka_conf_set` which will fail immediately for any unrecognized property

https://github.com/confluentinc/confluent-kafka-go/blob/5e4d04e05fc319ce5996a867aafef29059f26862/kafka/config.go#L130-L132

That means that about 50% of the time I am starting my program and I set any `confluent.monitoring.interceptor.*` property the program crashes with `panic: No such configuration property` because the `confluent.monitoring.interceptor.*` came before the `"plugin.library.paths":                     "monitoring-interceptor"` property during the iteration over the config keys.

Running librdkafka 0.11.5 and confluent-kafka-go version v0.11.4

rnpridgeon:
Thanks for reporting this @dtheodor, we actually ran into [this issue](https://github.com/confluentinc/confluent-kafka-python/commit/d6c85393e3df2a1c95fa04161a67f03cc398443f) with the python client recently as well. 

To fix this the property `plugin.library.paths` is intercepted and processed prior to handling any additional properties. I can get a fix in for this sometime this week or if you'd like you are more than welcome to submit a PR yourself. 

Let me know, 
Ryan

dtheodor:
I can do a PR. Thinking of extending the following snippet with something like this

https://github.com/confluentinc/confluent-kafka-go/blob/5e4d04e05fc319ce5996a867aafef29059f26862/kafka/config.go#L130-L133

```go
func configConvertAnyconf(m ConfigMap, anyconf rdkAnyconf) (err error) {
  // set the plugin.library.paths first
  paths, exist := m["plugin.library.paths"]
  if exist {
    anyconfSet(anyconf, "plugin.library.paths", paths)
    delete(m, "plugin.library.paths")
  } 
  for k, v := range m {
    // .. the config property loop unchanged
```

dtheodor:
[...]

PR #226, opened 2018-09-04
Make sure plugins are set before other configuration options

dtheodor:
Fixes https://github.com/confluentinc/confluent-kafka-go/issues/225

### DIFF
--- a/kafka/config_test.go
+++ b/kafka/config_test.go
@@ -124,3 +124,21 @@ func TestConfigMapAPIs(t *testing.T) {
 	}
 
 }
+
+// Test that plugins will always be configured before their config options
+func TestConfigPluginPaths(t *testing.T) {
+	config := &ConfigMap{
+		"plugin.library.paths":                     "monitoring-interceptor",
+		"confluent.monitoring.interceptor.icdebug": true,
+	}
+
+	// convert() would fail randomly due to random order of ConfigMap key iteration.
+	// running convert() once gave the test case a 50% failure chance,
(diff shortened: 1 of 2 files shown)


dtheodor:
[clabot:check]

kafka/config_test.go
```diff
@@ -124,3 +124,21 @@ func TestConfigMapAPIs(t *testing.T) {
 	}
 
 }
+
+// Test that plugins will always be configured before their config options
+func TestConfigPluginPaths(t *testing.T) {
+	config := &ConfigMap{
+		"plugin.library.paths":                     "monitoring-interceptor",
```
edenhill: This will fail if the monitoring-interceptors are not installed.
I'm not sure what the best solution is here, maybe we should first try to set only plugin.library.paths to verify that the interceptor is available, and if not just skip the test.

kafka/config_test.go
```diff
@@ -124,3 +124,29 @@ func TestConfigMapAPIs(t *testing.T) {
...
 }
+
+// Test that plugins will always be configured before their config options
+func TestConfigPluginPaths(t *testing.T) {
+	config := &ConfigMap{
+		"plugin.library.paths": "monitoring-interceptor",
+	}
+	_, err := config.convert()
+	if err != nil {
+		t.Skipf("Missing monitoring-interceptor.so, %s", err)
```
edenhill: nit: drop .so suffix since many developers are on osx and they might get confused

PR #229, opened 2018-09-22
mirror consumer.Close()

yacn:
not necessarily required and maybe I'm just doing something wrong. Feel free to close.

### changes

Updates `kafka.Producer.Close()` to return an `error` like `kafka.Consumer.Close()` does. This way, I can write something like this:

```go
type ResourceCloser interface {
	Close() error
}

func CloseResourceWithTimeout(r ResourceCloser, name string, t time.Duration) {
	fmt.Fprintf(os.Stderr, "%% Closing %s\n", name)
	closeDone := make(chan struct{})
	go func() {
		r.Close()
		close(closeDone)
	}()
	select {
	case <-closeDone:
		fmt.Fprintf(os.Stderr, "%% Closed %s\n", name)
		break
	case <-time.After(t):
		fmt.Fprintf(os.Stderr, "%% Timed out waiting for %s to close, moving on\n", name)
		break
	}
}

CloseResourceWithTimeout(c, "consumer", 45 * time.Second)
CloseResourceWithTimeout(p, "producer", 45 * time.Second)
```

instead of having to have two identical functions for both `kafka.Consumer` and `kafka.Producer`.

Maybe I'm thinking about this incorrectly / am not understanding the correct way to do this in Go.

### DIFF
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -346,7 +346,7 @@ func (p *Producer) Flush(timeoutMs int) int {
 
 // Close a Producer instance.
 // The Producer object or its channels are no longer usable after this call.
-func (p *Producer) Close() {
+func (p *Producer) Close() error {
 	// Wait for poller() (signaled by closing pollerTermChan)
 	// and channel_producer() (signaled by closing ProduceChannel)
 	close(p.pollerTermChan)
@@ -358,6 +358,7 @@ func (p *Producer) Close() {
 	p.handle.cleanup()
 
 	C.rd_kafka_destroy(p.handle.rk)
+	return nil
 }
 
 // NewProducer creates a new high-level Producer instance.


rnpridgeon:
:+1:  I'm actually all for making these handles more interface friendly. Might I make a suggestion that we move `Close() error` to the `Handle`  interface in [handle.go]( https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/handle.go#L33). 

There are actually a few methods which could be added to the `Handle` interface :

1. `Events() chan Event`
2. `GetMetadata(topic *string, allTopics bool, timeoutMs int) (*Metadata, error)` 
3. `QueryWatermarkOffsets(topic string, partition int32, timeoutMs int) (low, high int64, err error)`. 

While you're in there do you mind adding those to the `Handle` interface. 

Then you could optionally use the `Handle` interface

```
func CloseResourceWithTimeout(h Handle, name string, t time.Duration)
```

Or even embed the interface in your own 

```
type ResourceCloser interface {
	Handle
}
```

yacn:
@rnpridgeon can do, I'll handle those changes tomorrow morning when I get into the office.

yacn:
Not sure what the CI failure is about:

```
package github.com/golang/lint/golint: code in directory /Users/travis/gopath/src/github.com/golang/lint/golint expects import "golang.org/x/lint/golint"
```

yacn:
Okay, now we're getting somewhere. 

According to https://github.com/golang/tools/commit/a53bc13a6341e9847bbcf8c0a0691e60e18a87c0

Support for GoLang versions < 1.10 in `gcimporter` was removed.

So basically, any builds for go versions < 1.10 are going to fail.

That being addressed, I can now get to the actual code failures 😄

yacn:
@rnpridgeon I removed `Events()` and `QueryWatermarkOffsets()` from `Handle` since the new Admin API client doesn't implement either but uses the `Handle` interface.

yacn:
@rnpridgeon can I get another review?

edenhill:
Let's revisit this.

yacn:
Please let me know if there's anything I can do to assist.

PR #241, opened 2018-10-23
Fix use-after-free in AdminClient when context is cancelled

edenhill:


.travis.yml
```diff
@@ -25,14 +25,15 @@ before_install:
   - if [[ $TRAVIS_OS_NAME == linux ]]; then sudo apt-get install confluent-librdkafka-plugins -y ; fi
   - rm -rf tmp-build
   - bash mk/bootstrap-librdkafka.sh ${LIBRDKAFKA_VERSION} tmp-build
-  - go get -u github.com/golang/lint/golint
+  # golint requires Go >= 1.9
+  - if [[ ! $TRAVIS_GO_VERSION =~ ^1\.[78] ]] ; then go get -u golang.org/x/lint/golint && touch .do_lint ; fi
```
rnpridgeon: It might be worth investigating if `dep` can handle this sort of thing for us.

kafka/adminapi.go
```diff
@@ -331,8 +331,10 @@ func (a *AdminClient) waitResult(ctx context.Context, cQueue *C.rd_kafka_queue_t
 				}
 
 			default:
-				// Wait for result event for at most 100ms
-				rkev := C.rd_kafka_queue_poll(cQueue, 100)
+				// Wait for result event for at most 50s
```
rnpridgeon: I believe this should be in milliseconds not seconds. A 50 second delay in context cancellation would likely induce saddness.

edenhill:
```suggestion
				// Wait for result event for at most 50ms
```

Issue #59, opened 2017-04-28
go build -tags static


nicbaz:
Hi,

I'm having trouble building a binary with **librdkafka** as a static library.

```
$ go build -tags static -ldflags -v bin/blu.go
a# command-line-arguments
HEADER = -H4 -T0x401000 -D0x0 -R0x1000
searching for runtime.a in $WORK/runtime.a
searching for runtime.a in /go/pkg/linux_amd64/runtime.a
searching for runtime.a in /usr/local/go/pkg/linux_amd64/runtime.a
 0.00 deadcode
 0.03 pclntab=1408982 bytes, funcdata total 187752 bytes
 0.04 dodata
 0.05 dwarf
 0.08 symsize = 0
 0.12 reloc
 0.13 asmb
 0.13 codeblk
 0.14 datblk
 0.15 sym
 0.15 symsize = 256320
 0.16 symsize = 258288
 0.16 dwarf
 0.18 headr
 0.19 host link: "gcc" "-m64" "-gdwarf-2" "-o" "/tmp/go-build416718997/command-line-arguments/_obj/exe/a.out" "-rdynamic" "/tmp/go-link-567465902/go.o" "/tmp/go-link-567465902/000000.o" "/tmp/go-link-567465902/000001.o" "/tmp/go-link-567465902/000002.o" "-g" "-O2" "-g" "-O2" "-Wl,-Bstatic" "-lrdkafka" "-Wl,-Bdynamic" "-lrdkafka" "-lpthread" "-lz" "-lcrypto" "-lssl" "-lsasl2" "-lrt" "-g" "-O2" "-lpthread"
 0.29 cpu time
94449 symbols
84360 liveness data
$ ldd blu
	linux-vdso.so.1 (0x00007ffda22d0000)
	librdkafka.so.1 => /usr/lib/x86_64-linux-gnu/librdkafka.so.1 (0x00007f3b0fc16000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3b0f9f9000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f3b0f7dd000)
	libcrypto.so.1.0.0 => /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.0 (0x00007f3b0f3e1000)
[...]

PR #249, opened 2018-11-26
Fix static build

pragus:
Based on https://github.com/confluentinc/confluent-kafka-go/issues/59

### DIFF
--- a/kafka/build_static.go
+++ b/kafka/build_static.go
@@ -3,5 +3,6 @@
 
 package kafka
 
-// #cgo pkg-config: rdkafka-static
+// #cgo pkg-config: rdkafka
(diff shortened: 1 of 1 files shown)


kafka/build_static.go
```diff
@@ -3,5 +3,6 @@
 
 package kafka
 
-// #cgo pkg-config: rdkafka-static
+// #cgo pkg-config: rdkafka
+// #cgo LDFLAGS: -Wl,-Bstatic -lrdkafka -llz4 -lssl -lcrypto -lz -lgcc -Wl,-Bdynamic -lsasl2 -lrt -ldl -lm -lpthread
```
edenhill: This list of libraries depends on the librdkafka build options, which will vary and is thus not general applicable.
That's why we use pkg-config, it tells us the libraries librdkafka was built with

PR #271, opened 2019-01-21
Fix deprecation format

edenhill:


kafka/config.go
```diff
@@ -39,15 +39,15 @@ type ConfigValue interface{}
...
 // topic configuration properties shall be specified in the standard ConfigMap.
 // For backwards compatibility, "default.topic.config" (if supplied)
 // takes precedence.
 type ConfigMap map[string]ConfigValue
 
 // SetKey sets configuration property key to value.
 //
-// Deprecated: For user convenience a key prefixed with {topic}. will be
+// For user convenience a key prefixed with {topic}. will be
 // set on the "default.topic.config" sub-map, this use is deprecated.
```
rnpridgeon: Since `default.topic.config` has been deprecated this statement is no longer true right?

rnpridgeon: The  code below needs to be removed now that `default.topic.config` is no longer used.

```
	if strings.HasPrefix(key, "{topic}.") {
		_, found := m["default.topic.config"]
		if !found {
			m["default.topic.config"] = ConfigMap{}
		}
		m["default.topic.config"].(ConfigMap)[strings.TrimPrefix(key, "{topic}.")] = value
	} else {
		m[key] = value
	}
```

edenhill: Since both {topic} and default.topic.config are deprecated I think we can leave this as-is and remove them both in v2.0

rnpridgeon: Fair enough

PR #281, opened 2019-02-07
Add support for fatal errors

edenhill:


kafka/config.go
```diff
@@ -68,7 +68,7 @@ func (m ConfigMap) SetKey(key string, value ConfigValue) error {
 func (m ConfigMap) Set(kv string) error {
 	i := strings.Index(kv, "=")
 	if i == -1 {
-		return Error{ErrInvalidArg, "Expected key=value"}
+		return newErrorFromString(ErrInvalidArg, "Expected key=value")
```
rnpridgeon: :+1: I like this much better.

kafka/event.go
```diff
@@ -231,6 +231,18 @@ out:
...
+				// A fatal error has been raised.
+				// Extract the actual error from the client
+				// instance and return a new Error with
+				// fatal set to true.
+				cFatalErrstrSize := C.size_t(512)
+				cFatalErrstr := (*C.char)(C.malloc(cFatalErrstrSize))
+				defer C.free(unsafe.Pointer(cFatalErrstr))
+				cFatalErr := C.rd_kafka_fatal_error(h.rk, cFatalErrstr, cFatalErrstrSize)
+				fatalErr := newErrorFromCString(cFatalErr, cFatalErrstr)
+				fatalErr.fatal = true
```
rnpridgeon: newErrorFromCString is not exported anyway so why not add a isFatal argument and use that? Alternatively you could use and enum for log severity to make it more robust.

edenhill: I didn't want to change the 10 other places that don't need to set the fatal thing.

PR #298, opened 2019-03-05
Add GetWatermarkOffsets API

kevinconaway:
This PR exposes the [`rd_kafka_get_watermark_offsets`](https://github.com/edenhill/librdkafka/blob/8ea4c10549d5fab2d417e46f9a6e89c852c6e7b7/src/rdkafka.h#L2456) API as `GetWatermarkOffsets`

This is similar to `QueryWatermarkOffsets` except that it does not query the broker for the offsets, it will return the values that are internally cached on the client.  For the purposes of calculating the consumer lag, this is sufficient because the "high watermark" is returned with every fetch request so it will be available while consuming.  

Having this API allows consumers to calculate the lag while avoiding the overhead of either querying the broker or generating and parsing the statistics JSON

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -482,16 +482,15 @@ func (c *Consumer) GetMetadata(topic *string, allTopics bool, timeoutMs int) (*M
 	return getMetadata(c, topic, allTopics, timeoutMs)
 }
 
-// QueryWatermarkOffsets queries the broker for the low and high offsets for the given topic
-// and partition.
+// QueryWatermarkOffsets queries the broker for the low and high offsets for the given topic and partition.
 func (c *Consumer) QueryWatermarkOffsets(topic string, partition int32, timeoutMs int) (low, high int64, err error) {
 	return queryWatermarkOffsets(c, topic, partition, timeoutMs)
 }
 
 // GetWatermarkOffsets returns the cached low and high offsets for the given topic
 // and partition.  The high offset is populated on every fetch response or via calling QueryWatermarkOffsets.
-// The low offset is populated only when calling QueryWatermarkOffsets.  0 will be returned if
-// there is no cached offset for either value.
+// The low offset is populated every statistics.interval.ms if that value is set.
+// OffsetInvalid will be returned if there is no cached offset for either value.
 func (c *Consumer) GetWatermarkOffsets(topic string, partition int32) (low, high int64, err error) {
 	return getWatermarkOffsets(c, topic, partition)
 }
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -482,10 +482,18 @@ func (c *Consumer) GetMetadata(topic *string, allTopics bool, timeoutMs int) (*M
 	return getMetadata(c, topic, allTopics, timeoutMs)
 }
 
-// QueryWatermarkOffsets returns the broker's low and high offsets for the given topic
+// QueryWatermarkOffsets queries the broker for the low and high offsets for the given topic
 // and partition.
 func (c *Consumer) QueryWatermarkOffsets(topic string, partition int32, timeoutMs int) (low, high int64, err error) {
 	return queryWatermarkOffsets(c, topic, partition, timeoutMs)
+}
+
+// GetWatermarkOffsets returns the cached low and high offsets for the given topic
+// and partition.  The high offset is populated on every fetch response or via calling QueryWatermarkOffsets.
+// The low offset is populated only when calling QueryWatermarkOffsets.  0 will be returned if
+// there is no cached offset for either value.
+func (c *Consumer) GetWatermarkOffsets(topic string, partition int32) (low, high int64, err error) {
+	return getWatermarkOffsets(c, topic, partition)
 }
 
 // OffsetsForTimes looks up offsets by timestamp for the given partitions.
--- a/kafka/integration_test.go
+++ b/kafka/integration_test.go
@@ -445,9 +445,7 @@ func TestConsumerGetWatermarkOffsets(t *testing.T) {
 		"bootstrap.servers":        testconf.Brokers,
 		"group.id":                 testconf.GroupID,
 		"session.timeout.ms":       6000,
-		"api.version.request":      "true",
 		"enable.auto.commit":       false,
-		"debug":                    ",",
 		"auto.offset.reset":        "earliest",
 	}
 	_ = config.updateFromTestconf()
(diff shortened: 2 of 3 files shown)


edenhill:
Thank you! Will review this as soon as v1.0.0 is out (a couple of weeks)

kevinconaway:
@edenhill With 1.0.0 out, can this be revisited please?

Thank you

kafka/consumer.go
```diff
@@ -482,12 +482,20 @@ func (c *Consumer) GetMetadata(topic *string, allTopics bool, timeoutMs int) (*M
 	return getMetadata(c, topic, allTopics, timeoutMs)
 }
 
-// QueryWatermarkOffsets returns the broker's low and high offsets for the given topic
+// QueryWatermarkOffsets queries the broker for the low and high offsets for the given topic
```
edenhill: can you add " and partition." to the end of this sentence?

kafka/consumer.go
```diff
@@ -482,12 +482,20 @@ func (c *Consumer) GetMetadata(topic *string, allTopics bool, timeoutMs int) (*M
...
 // and partition.
 func (c *Consumer) QueryWatermarkOffsets(topic string, partition int32, timeoutMs int) (low, high int64, err error) {
 	return queryWatermarkOffsets(c, topic, partition, timeoutMs)
 }
 
+// GetWatermarkOffsets returns the cached low and high offsets for the given topic
+// and partition.  The high offset is populated on every fetch response or via calling QueryWatermarkOffsets.
+// The low offset is populated only when calling QueryWatermarkOffsets.  0 will be returned if
```
edenhill: The low offset is populated every statistics.interval.ms or when calling QueryWatermarkOffsets.

kafka/integration_test.go
```diff
@@ -433,6 +433,68 @@ func TestConsumerQueryWatermarkOffsets(t *testing.T) {
...
+		t.Skipf("Missing testconf.json")
+	}
+
+	// Create consumer
+	config := &ConfigMap{
+		"go.events.channel.enable": true,
+		"bootstrap.servers":        testconf.Brokers,
+		"group.id":                 testconf.GroupID,
+		"session.timeout.ms":       6000,
+		"api.version.request":      "true",
```
edenhill: api.version.request not required.

kafka/integration_test.go
```diff
@@ -433,6 +433,68 @@ func TestConsumerQueryWatermarkOffsets(t *testing.T) {
...
+		"enable.auto.commit":       false,
+		"debug":                    ",",
```
edenhill: remove "debug"

kafka/metadata.go
```diff
@@ -156,3 +156,29 @@ func queryWatermarkOffsets(H Handle, topic string, partition int32, timeoutMs in
...
+
+	var cLow, cHigh C.int64_t
+
+	e := C.rd_kafka_get_watermark_offsets(h.rk, ctopic, C.int32_t(partition),
+		&cLow, &cHigh)
+	if e != C.RD_KAFKA_RESP_ERR_NO_ERROR {
+		return 0, 0, newError(e)
+	}
+
+	if cLow != C.RD_KAFKA_OFFSET_INVALID {
```
edenhill: I don't think these two if statements should be there, since it will convert an Invalid offset into a proper one (0).
Just do the cast.

kevinconaway: If I remove them, my current test will fail with

```
integration_test.go:487: QueryWatermarkOffsets low[0] does not equal GetWatermarkOffsets low[-1001]
```

`QueryWatermarkOffsets` correctly returns 0 for the low watermark but `GetWatermarkOffsets` will return `RD_KAFKA_OFFSET_INVALID` (1001).

Is that the behavior that we want ?

kafka/consumer.go
kevinconaway: Actually, going through this now, it looks like _nothing_ populates the low offset (`rktp->rktp_lo_offset`) that is returned by `get_watermark_offsets`

https://github.com/edenhill/librdkafka/search?q=rktp_lo_offset&unscoped_q=rktp_lo_offset

edenhill: It is only updated when statistics is enabled.

edenhill: Changed the wording in my comment above.

kafka/metadata.go
edenhill: Yes, that's the correct behaviour.
It must not default to offset 0, which is a valid offset, but not necessarily the low offset of the partition (after log segment deletion).

edenhill: To fix your test:
Add statistics.interval.ms=100 and a short sleep (200ms) after the first message has been receieved to make sure the low offset is populated.

kevinconaway: I tried that by doing something like

```
	seenMessage, seenStats := false, false
	// Wait for messages to be received so that we know the watermark offsets have been delivered
	// with the fetch response
	for ev := range c.Events() {
		if _, ok := ev.(*Message); ok {
			seenMessage = true
		}
		if _, ok := ev.(*Stats); ok {
			seenStats = true
		}
		if seenMessage && seenStats {
			break
		}
	}
```

But the low watermark is still invalid.  Even sleeping for 5s doesn't help.

Per my comment above and based on a quick glance through the code, I couldn't find anything that actually sets the specific `rktp_lo_offset` field

edenhill: It is set here, in the Offset consumer_lag response callback, which is triggered every statistics.interval.ms.
https://github.com/edenhill/librdkafka/blob/d306f5f7ef2aea04cdb9e7146796efd4a28d49f1/src/rdkafka_partition.c#L106

But it only fires for assigned partitions, so the sleep needs to be done after the group assignment has kicked in, and the safest way is to sleep after seeing the first message.

kevinconaway: OK I believe that figured out why it wasn't working.  It looks like minimum interval to begin querying the lower watermark offset is 10s per [this block](https://github.com/edenhill/librdkafka/blob/d306f5f7ef2aea04cdb9e7146796efd4a28d49f1/src/rdkafka_partition.c#L225-L236).

If I sleep for at least 10s, the code works as expected but this isn't ideal.  I'm inclined to avoid checking the low offset in this test.  Thoughts ?

edenhill: You're right, 10s! :)

I'm fine with skipping the low offset check, yeah.

edenhill:
Thanks for your contribution!

PR #300, opened 2019-03-06
WIP: Add KIP-255 SASL/OAUTHBEARER support

rondagostino:
Adds SASL/OAUTHBEARER support.

Depends on https://github.com/edenhill/librdkafka/pull/2189
Depends on merge of https://github.com/edenhill/librdkafka/tree/singlekinit

### DIFF
--- a/examples/README
+++ b/examples/README
@@ -7,7 +7,7 @@ Examples:
   producer_channel_example - Channel based producer
   producer_example - Function based producer
 
-  go-kafkacat - Channel based kafkacat Go clone
+  go-kafkacat - Channel based kafkacat Go clone, also provides unsecured SASL/OAUTHBEARER example
 
 
 Usage example:
--- a/examples/README
+++ b/examples/README
@@ -7,7 +7,9 @@ Examples:
   producer_channel_example - Channel based producer
   producer_example - Function based producer
 
-  go-kafkacat - Channel based kafkacat Go clone, also provides unsecured SASL/OAUTHBEARER example
+  go-kafkacat - Channel based kafkacat Go clone
+
+  oauthbearer_example - Provides unsecured SASL/OAUTHBEARER example
 
 
 Usage example:
--- a/examples/go-kafkacat/go-kafkacat.go
+++ b/examples/go-kafkacat/go-kafkacat.go
@@ -113,22 +113,27 @@ func retrieveUnsecuredToken(e kafka.OAuthBearerTokenRefresh, config *kafka.Confi
 	if len(oauthbearerConfigMap) > 3 {
 		return kafka.OAuthBearerToken{}, fmt.Errorf("ignoring event %T: unrecognized key(s): %s=%s", e, saslDotOAuthBearerDotConfig, v)
 	}
-	now := time.Now().Unix()
-	lifetimeSeconds := now + int64(lifeSeconds)
+
+	now := time.Now()
+	nowSecondsSinceEpoch := now.Unix()
+
+	expiration := now.Add(time.Second * time.Duration(lifeSeconds))
+	expirationSecondsSinceEpoch := expiration.Unix()
+
 	oauthbearerMapForJSON := map[string]interface{}{
 		principalClaimName: principal,
-		"iat":              now,
-		"exp":              lifetimeSeconds,
+		"iat":              nowSecondsSinceEpoch,
+		"exp":              expirationSecondsSinceEpoch,
 	}
 	claimsJSON, _ := json.Marshal(oauthbearerMapForJSON)
 	encodedClaims := base64.RawURLEncoding.EncodeToString(claimsJSON)
 	jwsCompactSerialization := joseHeaderEncoded + "." + encodedClaims + "."
 	extensions := map[string]string{}
 	oauthBearerToken := kafka.OAuthBearerToken{
-		TokenValue:           jwsCompactSerialization,
-		LifetimeMilliseconds: lifetimeSeconds * 1000,
-		Principal:            principal,
-		Extensions:           extensions,
+		TokenValue: jwsCompactSerialization,
+		Expiration: expiration,
+		Principal:  principal,
+		Extensions: extensions,
 	}
 	return oauthBearerToken, nil
 }
--- a/examples/go-kafkacat/go-kafkacat.go
+++ b/examples/go-kafkacat/go-kafkacat.go
@@ -17,6 +17,12 @@ package main
  * limitations under the License.
  */
 
+/*
+#include <stdlib.h>
+#include <librdkafka/rdkafka.h>
+*/
+import "C"
+
 import (
 	"bufio"
 	"encoding/base64"
@@ -63,8 +69,15 @@ const (
 // It must be invoked whenever kafka.OAuthBearerTokenRefresh appears on the client's event channel,
 // which will occur whenever the client requires a token (i.e. when it first starts and when the
 // previously-received token is 80% of the way to its expiration time).
-func handleOAuthBearerTokenRefreshEvent(client kafka.Handle, e kafka.OAuthBearerTokenRefresh, config *kafka.ConfigMap) {
-	oauthBearerToken, retrieveErr := retrieveUnsecuredToken(e, config)
+func handleOAuthBearerTokenRefreshEvent(client kafka.Handle, e kafka.OAuthBearerTokenRefresh) {
+	val, err := client.GetConfigValue(saslDotOAuthBearerDotConfig)
+	if err != nil {
+		// illegal config value or unknown config name; should generally not happen
+		fmt.Fprintf(os.Stderr, "Unknown/llegal type for %s\n", saslDotOAuthBearerDotConfig)
+		os.Exit(1)
+	}
+
+	oauthBearerToken, retrieveErr := retrieveUnsecuredToken(e, val)
 	if retrieveErr != nil {
 		fmt.Fprintf(os.Stderr, "%% Token retrieval error: %v\n", retrieveErr)
 		client.SetOAuthBearerTokenFailure(retrieveErr.Error())
@@ -77,13 +90,9 @@ func handleOAuthBearerTokenRefreshEvent(client kafka.Handle, e kafka.OAuthBearer
 	}
 }
 
-func retrieveUnsecuredToken(e kafka.OAuthBearerTokenRefresh, config *kafka.ConfigMap) (kafka.OAuthBearerToken, error) {
-	v, err := config.Get(saslDotOAuthBearerDotConfig, "")
-	if err != nil {
-		return kafka.OAuthBearerToken{}, fmt.Errorf("ignoring event %T due to incorrect type for %s: %v", e, saslDotOAuthBearerDotConfig, err)
-	}
-	if !oauthbearerConfigRegex.MatchString(v.(string)) {
-		return kafka.OAuthBearerToken{}, fmt.Errorf("ignoring event %T due to malformed %s: %v", e, saslDotOAuthBearerDotConfig, v)
+func retrieveUnsecuredToken(e kafka.OAuthBearerTokenRefresh, v string) (kafka.OAuthBearerToken, error) {
+	if !oauthbearerConfigRegex.MatchString(v) {
+		return kafka.OAuthBearerToken{}, fmt.Errorf("ignoring event %T due to malformed %s: %s", e, saslDotOAuthBearerDotConfig, v)
 	}
 	// set up initial map with default values
 	oauthbearerConfigMap := map[string]string{
@@ -91,7 +100,7 @@ func retrieveUnsecuredToken(e kafka.OAuthBearerTokenRefresh, config *kafka.Confi
 		lifeSecondsKey:        "3600",
 	}
 	// parse the provided config and store name=value pairs in the map
-	for _, kv := range oauthbearerNameEqualsValueRegex.FindAllStringSubmatch(v.(string), -1) {
+	for _, kv := range oauthbearerNameEqualsValueRegex.FindAllStringSubmatch(v, -1) {
 		oauthbearerConfigMap[kv[1]] = kv[2]
 	}
 	principalClaimName := oauthbearerConfigMap[principalClaimNameKey]
@@ -155,7 +164,7 @@ func runProducer(config *kafka.ConfigMap, topic string, partition int32) {
 			if !ok {
 				switch e := ev.(type) {
 				case kafka.OAuthBearerTokenRefresh:
-					handleOAuthBearerTokenRefreshEvent(p, e, config)
+					handleOAuthBearerTokenRefreshEvent(p, e)
 				}
 				continue
 			}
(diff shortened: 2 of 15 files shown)


rondagostino:
@edenhill Can you take a look?

examples/go-kafkacat/go-kafkacat.go
```diff
@@ -37,6 +42,63 @@ var (
...
+	mdLifetimeSeconds := now + int64(lifeSeconds)
+	oauthbearerMapForJSON := map[string]interface{}{
+		principalClaimName: mdPrincipal,
+		"iat":              now,
+		"exp":              mdLifetimeSeconds,
+	}
+	claimsJSON, _ := json.Marshal(oauthbearerMapForJSON)
+	encodedClaims := base64.RawURLEncoding.EncodeToString(claimsJSON)
+	jwsCompactSerialization := joseHeaderEncoded + "." + encodedClaims + "."
+	return jwsCompactSerialization, mdLifetimeSeconds, mdPrincipal, map[string]string{}, nil
```
mkaminski1988: Maybe return a struct here? It's a bit difficult to understand what's being returned by looking at the method signature.

examples/go-kafkacat/go-kafkacat.go
```diff
@@ -37,6 +42,63 @@ var (
...
+	principalClaimName := oauthbearerConfigMap[principalClaimNameKey]
+	mdPrincipal := oauthbearerConfigMap[principalKey]
+	lifeSeconds, lifeSecondsErr := strconv.Atoi(oauthbearerConfigMap[lifeSecondsKey])
+	// regexp is such that principalClaimName and lifeSeconds cannot end up blank,
+	// so check for a blank principal (which will happen if it isn't specified)
+	if mdPrincipal == "" {
+		return "", 0, "", nil, fmt.Errorf("ignoring event %T: no %s: %s=%s", e, principalKey, saslDotOAuthBearerDotConfig, v)
+	}
+	// sanity-check the provided lifeSeconds value, which must be a positive integer
+	if lifeSecondsErr != nil || lifeSeconds == 0 {
```
mkaminski1988: This error check appears pretty far from the assignment. I'd recommend doing this check right after the call to `strconv.Atoi(...)`

examples/go-kafkacat/go-kafkacat.go
```diff
@@ -37,6 +42,63 @@ var (
 	sigs         chan os.Signal
 )
 
+var oauthbearerConfigRegex = regexp.MustCompile("^(\\s*(\\w+)\\s*=\\s*(\\w+))+\\s*$")
```
mkaminski1988: Maybe group these together? 
```
var (
    oauthbearerConfigRegex = regexp.MustCompile("^(\\s*(\\w+)\\s*=\\s*(\\w+))+\\s*$")
    oauthbearerNameEqualsValueRegex = regexp.MustCompile("(\\w+)\\s*=\\s*(\\w+)")
)

File: examples/go-kafkacat/go-kafkacat.go (hunk 2 above)
Comment (rondagostino) [reply examples/go-kafkacat/go-kafkacat.go]: 👍

File: examples/go-kafkacat/go-kafkacat.go (hunk 3 above)
Comment (rondagostino) [reply examples/go-kafkacat/go-kafkacat.go]: 👍

File: examples/go-kafkacat/go-kafkacat.go (hunk 1 above)
Comment (rondagostino) [reply examples/go-kafkacat/go-kafkacat.go]: 👍 Added `type OAuthBearerToken struct {...}` in `handle.go`

File: examples/go-kafkacat/go-kafkacat.go [hunk 4]
```diff
@@ -177,6 +265,18 @@ func runConsumer(config *kafka.ConfigMap, topics []string) {
 				if verbosity >= 2 {
 					fmt.Fprintf(os.Stderr, "%% %v\n", e)
 				}
+			case kafka.OAuthBearerTokenRefresh:
+				oauthBearerToken, retrieveErr := retrieveUnsecuredToken(e, config)
```
jamesd: The token refresh handling case for the consumer is basically the same as the producer. Can we use the `Handle` interface to make this a generic helper that both can use?

kafka/handle.go
```diff
@@ -205,3 +246,45 @@ func (h *handle) cgoGet(cgoid int) (cg cgoif, found bool) {
...
+	for key, value := range oauthBearerToken.Extensions {
+		cExtensions[extensionSize] = C.CString(key)
+		defer C.free(unsafe.Pointer(cExtensions[extensionSize]))
+		extensionSize++
+		cExtensions[extensionSize] = C.CString(value)
+		defer C.free(unsafe.Pointer(cExtensions[extensionSize]))
+		extensionSize++
+	}
+	var cExtensionsToUse **C.char
+	if extensionSize == 0 {
```
jamesd: `cExtensionsToUse` is initialized to nil so this can be simplified.
```suggestion
	if extensionSize > 0 {
		cExtensionsToUse = (**C.char)(unsafe.Pointer(&cExtensions[0]))
	}
```

rondagostino: 👍

examples/go-kafkacat/go-kafkacat.go
rondagostino: @jamesd I noticed this DRY issue, too.  I looked at it and noticed that the `fmt.Fprintf(os.Stderr,...)` statement is specific to `go-kafkacat`, and the retrieval is definitely something that will differ from one installation to another.  There is also desire to be able to differentiate between a retrieval error (which might be caused by the token endpoint being down) and an error setting the token (which is likely caused by a bad configuration -- for example an illegal extension key or value).  So the only code I can think of that might work is something like this:

```
oauthBearerToken, retrieveErr := retrieveUnsecuredToken(e, config)
if retrieveErr != nil {
	fmt.Fprintf(os.Stderr, "%% Token retrieval error: %v\n", retrieveErr)
}
setTokenError := theHandle.processOAuthBearerTokenRetrieval(oauthBearerToken, retrieveErr)
if setTokenError != nil {
	fmt.Fprintf(os.Stderr, "%% Error setting token and extensions: %v\n", setTokenError)
}
```

I'm not sure if declaring `func (h *Handle) processOAuthBearerTokenRetrieval(oauthBearerToken OAuthBearerToken, retrieveErr error) error` is worth it.  It _could_ eliminate the `SetOAuthBearerToken/SetOAuthBearerTokenFailure()` methods, though I hesitate to remove these building blocks.

Thoughts on this or another, possibly better way to adhere to DRY?

jamesd: That makes sense. At a minimum, how about just a helper in `go-kafkacat`? Something like:
```suggestion
func handleOAuthBearerTokenRefreshEvent(client Handle, e kafka.OAuthBearerTokenRefresh, config *kafka.ConfigMap) {
	oauthBearerToken, retrieveErr := retrieveUnsecuredToken(e, config)
	if retrieveErr != nil {
		fmt.Fprintf(os.Stderr, "%% Token retrieval error: %v\n", retrieveErr)
		client.SetOAuthBearerTokenFailure(retrieveErr.Error())
	} else {
		setTokenError := client.SetOAuthBearerToken(oauthBearerToken)
		if setTokenError != nil {
			fmt.Fprintf(os.Stderr, "%% Error setting token and extensions: %v\n", setTokenError)
			client.SetOAuthBearerTokenFailure(setTokenError.Error())
		}
	}
}
```

rondagostino: 👍 Done

rondagostino:
All set, @jamesd -- thanks for the review!

rondagostino:
[clabot:check]

rondagostino:
@edenhill Can you review?

examples/README
```diff
@@ -7,7 +7,7 @@ Examples:
   producer_channel_example - Channel based producer
   producer_example - Function based producer
 
-  go-kafkacat - Channel based kafkacat Go clone
+  go-kafkacat - Channel based kafkacat Go clone, also provides unsecured SASL/OAUTHBEARER example
```
edenhill: Can we move this to its own minimal example?
I think something that simply lists the brokers in the cluster (using GetMetadata()) is sufficient to show-case the OAUTHBEARER use.

I've done the mistake of trying to add everything to the same example before, and it just turns out to be a bad example (looking at you, [rdkafka_example.c](https://github.com/edenhill/librdkafka/blob/master/examples/rdkafka_example.c)).

examples/go-kafkacat/go-kafkacat.go
```diff
@@ -37,6 +42,89 @@ var (
 	sigs         chan os.Signal
 )
 
+var (
+	oauthbearerConfigRegex          = regexp.MustCompile("^(\\s*(\\w+)\\s*=\\s*(\\w+))+\\s*$")
```
edenhill: Add a comment of what format it is trying to match.

examples/go-kafkacat/go-kafkacat.go
```diff
@@ -37,6 +42,89 @@ var (
...
+
+const (
+	saslDotOAuthBearerDotConfig = "sasl.oauthbearer.config"
+	principalClaimNameKey       = "principalClaimName"
+	principalKey                = "principal"
+	lifeSecondsKey              = "lifeSeconds"
+	joseHeaderEncoded           = "eyJhbGciOiJub25lIn0" // {"alg":"none"}
+)
+
+func handleOAuthBearerTokenRefreshEvent(client kafka.Handle, e kafka.OAuthBearerTokenRefresh, config *kafka.ConfigMap) {
```
edenhill: docstring what this does, when, and why

examples/go-kafkacat/go-kafkacat.go
```diff
@@ -52,6 +140,10 @@ func runProducer(config *kafka.ConfigMap, topic string, partition int32) {
 		for ev := range drs {
 			m, ok := ev.(*kafka.Message)
 			if !ok {
+				switch e := ev.(type) {
+				case kafka.OAuthBearerTokenRefresh:
+					handleOAuthBearerTokenRefreshEvent(p, e, config)
```
edenhill: I think we'll want a better way to provide the config here so the application does not need to hang on to it.

In librdkafka we should either pass the "sasl.oauthbearer.config" value as an argument to the refresh callback, which is the more direct approach, or provide an API to extract config values from an rd_kafka_t handle, which is the more generic and flexible approach.

I'm voting for the latter and if we agree on that being the solution I'll make the PR for librdkafka.

What do you think?

kafka/handle.go
```diff
@@ -28,9 +28,50 @@ import (
...
+type OAuthBearerToken struct {
+	// Token value, often (but not necessarily) a JWS compact serialization
+	// as per https://tools.ietf.org/html/rfc7515#section-3.1; it must meet
+	// the regular expression for a SASL/OAUTHBEARER value defined at
+	// https://tools.ietf.org/html/rfc7628#section-3.1
+	TokenValue string
+	// Metadata about the token indicating when it expires
+	// in terms of the number of milliseconds since the epoch; it must
+	// represent a time in the future
+	LifetimeMilliseconds int64
```
edenhill: Shouldn't we use the Time type here?
Is there any value/correctness in using int64 instead?

kafka/handle.go
```diff
@@ -28,9 +28,50 @@ import (
...
+	// to a broker during SASL/OAUTHBEARER authentication. It will return nil
+	// on success, otherwise an error if:
+	// 1) the token data is invalid (meaning an expiration time in the past
+	// or either a token value or an extension key or value that does not meet
+	// the regular expression requirements as per
+	// https://tools.ietf.org/html/rfc7628#section-3.1);
+	// 2) SASL/OAUTHBEARER is not supported by the underlying librdkafka build;
+	// 3) SASL/OAUTHBEARER is supported but is not configured as the client's
+	// authentication mechanism.
+	SetOAuthBearerToken(oauthBearerToken OAuthBearerToken) error
```
edenhill: Since these are exposed on the Consumer, Producer, Admin.. clients, do these methods really need to be public on the Handle?

kafka/handle.go
```diff
@@ -205,3 +246,43 @@ func (h *handle) cgoGet(cgoid int) (cg cgoif, found bool) {
...
+
+func (h *handle) setOAuthBearerToken(oauthBearerToken OAuthBearerToken) error {
+	cTokenValue := C.CString(oauthBearerToken.TokenValue)
+	defer C.free(unsafe.Pointer(cTokenValue))
+	cPrincipal := C.CString(oauthBearerToken.Principal)
+	defer C.free(unsafe.Pointer(cPrincipal))
+	cErrstrSize := C.size_t(512)
+	cErrstr := (*C.char)(C.malloc(cErrstrSize))
+	defer C.free(unsafe.Pointer(cErrstr))
+	cExtensions := make([]*C.char, 2*len(oauthBearerToken.Extensions))
```
edenhill: nit: I wouldn't mind some more air in this function. Blank lines between sections

kafka/kafka.go
```diff
@@ -125,6 +125,15 @@
 // These errors are normally just informational since the
 // client will try its best to automatically recover (eventually).
 //
+// * `OAuthBearerTokenRefresh` - time to retrieve a new SASL/OAUTHBEARER token.
```
edenhill: "time to" might be interpreted as this is an actual time(stamp), might be better with "request to" or similar.

examples/go-kafkacat/go-kafkacat.go
rondagostino: IIUC, you would prefer the call to be simply `handleOAuthBearerTokenRefreshEvent(p, e)` rather than passing in the config explicitly as a third parameter, and then inside that function the implementation could retrieve the `sasl.oauthbearer.config` value via this new API.  If that is a correct understanding, then yes, I think it makes sense.  Now that you mention it, actually, it does seem strange that there is no way to interrogate the underlying configuration.

> I'm voting for the latter and if we agree on that being the solution I'll make the PR for librdkafka.

That would be great

edenhill: Come to think of it, I think we should do both:
 1. pass the value of sasl.oauthbearer.config to the refresh callback, since most implementations will need configuration.
 2. provide a way to query any configuration value, for more complex implementations that perhaps rely on other client settings as well, whatever that is.

kafka/handle.go
rondagostino: Good question.  If they are not public then the code cannot be generalized to work with any client instance, and it is a Do-Not-Repeat-Yourself issue.  I had originally had repeated code as per the review comments on March 14th by jamesd, and I added these methods to `Handle` to allow consolidation of the mostly-repeated code into a single method.

kafka/handle.go
rondagostino: 👍

kafka/kafka.go
rondagostino: 👍

examples/go-kafkacat/go-kafkacat.go
rondagostino: 👍

examples/go-kafkacat/go-kafkacat.go
rondagostino: 👍

examples/go-kafkacat/go-kafkacat.go
rondagostino: OK, sounds good.  Will you do the PR to take care of both, or should I do the refresh callback one?

kafka/handle.go
rondagostino: 👍

examples/go-kafkacat/go-kafkacat.go
rondagostino: All set, the call is now simply `handleOAuthBearerTokenRefreshEvent(p, e)`

examples/README
rondagostino: Thanks for adding `consumer_test.go` and `producer_test.go` while I was away.

edenhill:
Fantastic work yet again @rondagostino , thank you!

PR #301, opened 2019-03-06
Added rd_kafka_list_groups API to get consumer group subscription metadata

simnotes:
I added new GetConsumerGroupsMetadata function to consumer, producer and adminClient API, which calls rd_kafka_list_groups from librdkafka to retrieve current consumer group status and consumer group subscriptions from broker. Via the new API it is possible to query subscription-status like in "kafka-consumer-groups.sh"

Also added integration and dry-tests to everything.


edenhill:
Thank you! Will review this as soon as v1.0.0 is out (a couple of weeks)

edenhill:
Thank you! Will review this as soon as v1.0.0 is out (a couple of weeks)

ruseinov:
@edenhill Any idea if this ever gets merged into master? Looking for a similar functionality to support dynamic kafka monitoring for all the consumers

ruseinov:
@edenhill Any idea if this ever gets merged into master? Looking for a similar functionality to support dynamic kafka monitoring for all the consumers and this would be a lot of help.

kimba74:
@edenhill I second @ruseinov. I'm as well looking for the possibility of dynamic Kafka monitoring and this would be **very** helpful indeed.

sourcec0de:
@edenhill my team is also looking for this functionality.
Confluent Cloud doesn't make this information available from its metrics API.
This feature seems high value.

PR #307, opened 2019-03-14
v1.0.0 preps, fatal error fixes and an idempotent producer example

edenhill:


fmd:
@edenhill  can i try to add an example for exposing of `read_uncommitted=false` or `isolation.level=read_committed`? do you know which rev of librdkafka i'd need to test those features? i know it's going to be available in the 1.0.0 release but currently i'm on RC8 and `confluent-kafka-go` doesn't understand `isolation.level`

edenhill:
@fmd That'd be great, but isolation.level is not yet added to librdkafka (or its daughter clients) unfortunately.

fmd:
@edenhill Thanks for the quick response! what about `read_uncommitted=false`? i can't find any documentation for it and a quick search of librdkafka isn't returning results; is it a currently-existing workaround?

edenhill:
I'm not sure where read_uncommited=false is coming from, but it sounds like another name for the isolation.level.

fmd:
https://github.com/edenhill/librdkafka/issues/1308#issuecomment-329717547
^ i saw rumblings about it here. Just to confirm then, is librdkafka 1.0.0 likely to have that feature (`isolation.level`)?

edenhill:
librdkafka v1.0.0 will not support isolation.level/read_committed, that will come later (within a month or two).

examples/idempotent_producer_example/idempotent_producer_example.go
```diff
@@ -0,0 +1,178 @@
+// Idempotent Producer example.
+//
+// The idempotent producer provides strict ordering and
+// exactly-once producing guarantees.
+//
+// From the application developer's perspective, the only difference
+// from a standard producer is the enabling of the feature by setting
+// the `enable.idempotence` configuration property to true, and
+// handling fatal fatal errors (Error.IsFatal()) which are raised when the
```
mhowlett: a fatal fatal error sounds really bad.

examples/idempotent_producer_example/idempotent_producer_example.go
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			fmt.Printf("Failed to produce message: %v\n", err)
+		}
+
+		msgcnt++
+
+		// Since fatal errors can't be triggered in practice,
+		// use the test API to trigger a fabricated error after some time.
+		if msgcnt == 13 {
+			p.TestFatalError(kafka.ErrOutOfOrderSequenceNumber, "Testing fatal errors")
```
mhowlett: Although you've convinced me of the value of exposing this test method, I don't think it should be on the producer instance. My first instinct is to have a new test class with a method 'CauseFatalError' on it which takes a producer instance. A side effect of this (but not the primary driver, which is structural) is I also like having a different verb in there 'cause' or 'emit', since test could be considered an adjective. but we definitely don't want to drop the word test.

examples/idempotent_producer_example/idempotent_producer_example.go
edenhill: These things should not be underestimated, I'll add a third fatal for clarity

examples/idempotent_producer_example/idempotent_producer_example.go
edenhill: There are no classes in Go, the method needs to be super generic or be a Producer-typed interface - since this method is only relevant for the producer I think it resides in the right place.

I think Test should be in there, which is a verb in this case, so I dont think an additional verb (Cause) adds anything.
Anything named Test should stand out anyway and will not be confused with production code.

examples/idempotent_producer_example/idempotent_producer_example.go
```diff
@@ -0,0 +1,178 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import (
+	"fmt"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
```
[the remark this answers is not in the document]
edenhill: Good input, but since we have so few non-stdlib imports I think this might be premature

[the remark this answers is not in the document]
rnpridgeon: that's fair

PR #309, opened 2019-03-16
SASL_SSL why?

qiannianshaonian:
Invalid value "SASL_SSL" for configuration property "security.protocol"

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ func main() {
 		"bootstrap.servers": "localhost",
 		"group.id":          "myGroup",
 		"auto.offset.reset": "earliest",
+		"security.protocol": "SASL_SSL",
 	})
 
 	if err != nil {


qiannianshaonian:
@edenhill Can you take a look?

README.md
```diff
@@ -49,6 +49,7 @@ func main() {
 		"bootstrap.servers": "localhost",
 		"group.id":          "myGroup",
 		"auto.offset.reset": "earliest",
+		"security.protocol": "SASL_SSL",
```
edenhill: This is environment-specific, some clusters require SASL, SSL, or both.
And this configuration is not complete with SASL_SSL, also needs to set mechanisms, usernames, kerberos stuff, etc.

The README examples are aimed to be minimal, any environment-specific configuration is outside their scope.

qiannianshaonian: I have set up the sasl.username,sasl.password sasl.mechanisms, security.protocol and so on locally.  the demo is  https://github.com/Azure/azure-event-hubs-for-kafka/blob/master/quickstart/go/consumer/consumer.go Can you take a look?

edenhill: Looks good, apart from the sasl.username $ConnectionString reference.

PR #326, opened 2019-04-14
Support setting partitioner by name

the4thamigo-uk:
This is a PR for discussion.

### DIFF
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -124,6 +124,13 @@ rd_kafka_resp_err_t err;
       return RD_KAFKA_RESP_ERR_NO_ERROR;
 #endif
 }
+typedef int32_t (*partitioner_cb) (
+						const rd_kafka_topic_t *rkt,
+						const void *keydata,
+						size_t keylen,
+						int32_t partition_cnt,
+						void *rkt_opaque,
+						void *msg_opaque);
 */
 import "C"
 
@@ -422,6 +429,12 @@ func NewProducer(conf *ConfigMap) (*Producer, error) {
 	}
 	produceChannelSize := v.(int)
 
+	v, err = confCopy.extract("go.produce.partitioner", "")
+	if err != nil {
+		return nil, err
+	}
+	partitionerName := v.(string)
+
 	if int(C.rd_kafka_version()) < 0x01000000 {
 		// produce.offset.report is no longer used in librdkafka >= v1.0.0
 		v, _ = confCopy.extract("{topic}.produce.offset.report", nil)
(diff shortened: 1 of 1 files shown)


the4thamigo-uk:
> [clabot:check]

[clabot:check]

mhowlett:
the `partitioner` config property should achieve this already, right?

the4thamigo-uk:
> the `partitioner` config property should achieve this already, right?

Ah, right, so I simply set `"partitioner"="murmur2"` in the call to `NewProducer()`? Does that automatically get mapped to https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_conf.h#L467 in the topic configuration then? I couldn't quite see how this works.

the4thamigo-uk:
> > the `partitioner` config property should achieve this already, right?
> 
> Ah, right, so I simply set `"partitioner"="murmur2"` in the call to `NewProducer()`? Does that get mapped to https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_conf.h#L467 in the topic configuration then? I couldn't quite see how this works.

We have just tried this 

> > the `partitioner` config property should achieve this already, right?
> 
> Ah, right, so I simply set `"partitioner"="murmur2"` in the call to `NewProducer()`? Does that get mapped to https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_conf.h#L467 in the topic configuration then? I couldn't quite see how this works.



> > the `partitioner` config property should achieve this already, right?
> 
> Ah, right, so I simply set `"partitioner"="murmur2"` in the call to `NewProducer()`? Does that get mapped to https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_conf.h#L467 in the topic configuration then? I couldn't quite see how this works.

Ah I should use `topic.partitioner` it seems? https://github.com/confluentinc/confluent-kafka-go/blob/aba1d4b9ad6802580e1fa7015a7137af6a3e2893/kafka/config.go#L50

edenhill:
You'll want to set `"partitioner": "murmur2_random"` in the outer global ConfigMap that you pass to NewProducer(), but make sure that you don't have a "default.topic.config" entry in there since it will overwrite any topic-level configuration properties set on the global ConfigMap.
(default.topic.config is being deprecated in favour of setting topic properties on the global ConfigMap)

the4thamigo-uk:
> You'll want to set `"partitioner": "murmur2_random"` in the outer global ConfigMap that you pass to NewProducer(), but make sure that you don't have a "default.topic.config" entry in there since it will overwrite any topic-level configuration properties set on the global ConfigMap.
> (default.topic.config is being deprecated in favour of setting topic properties on the global ConfigMap)

Setting `{topic}.partitioner` it seems, using `ConfigMap.SetKey()`?

edenhill:
You could use SetKey(), but don't use the "{topic}." prefix, just "partitioner"

the4thamigo-uk:
> You could use SetKey(), but don't use the "{topic}." prefix, just "partitioner"

Thanks, I tried this, but it didnt seem to work. I'm quite confused now, I thought `partitioner` was a topic level setting and the logic in [here](https://github.com/confluentinc/confluent-kafka-go/blob/460e8e43b282a1a68219df600ef63442b81faf5f/kafka/config.go#L47) suggests to prefix with `{topic}.`. Maybe this is due to the fact we are still using https://github.com/edenhill/librdkafka/releases/tag/v0.11.6?

edenhill:
This works for me:
```golang
	cm := &kafka.ConfigMap{"bootstrap.servers": broker}
        // or just specifiy the partitioner right in the ConfigMap instantiation above
	cm.SetKey("partitioner", "murmur2_random")
	p, err := kafka.NewProducer(cm)
```

the4thamigo-uk:
> This works for me:
> 
> ```go
> 	cm := &kafka.ConfigMap{"bootstrap.servers": broker}
>         // or just specifiy the partitioner right in the ConfigMap instantiation above
> 	cm.SetKey("partitioner", "murmur2_random")
> 	p, err := kafka.NewProducer(cm)
> ```



> This works for me:
> 
> ```go
> 	cm := &kafka.ConfigMap{"bootstrap.servers": broker}
>         // or just specifiy the partitioner right in the ConfigMap instantiation above
> 	cm.SetKey("partitioner", "murmur2_random")
> 	p, err := kafka.NewProducer(cm)
> ```

Not sure what Im doing differently, but `{topic}.partitioner` does seems to work, in that I can join my streams, without having to fork this library so thats great. Thanks for your help.

PR #349, opened 2019-07-03
Improve the Confluent Cloud Go example for better Developer Experience

riferrei:
This PR essentially improves the developer experience of users trying this example with Confluent Cloud. Here is the link of changes made:

- The import has been changed from 'github.com/confluentinc/confluent-kafka-go/kafka' to 'gopkg.in/confluentinc/confluent-kafka-go.v1/kafka' to be in line with what the docs suggest.

- Instructions regarding how the user can retrieve information about the bootstrap server and API key has been provided, according to the behavior of the new CCloud CLI.

- The information about the bootstrap server and the API key have been made available in the code as constants, and now they are reused across the Admin API, the producer, and the consumer.

- A new function called 'createTopic()' was created to automatically create the topic for the user. This reduces the steps from him/her to see the example working.

### DIFF
--- a/examples/confluent_cloud_example/confluent_cloud_example.go
+++ b/examples/confluent_cloud_example/confluent_cloud_example.go
@@ -153,6 +153,7 @@ func createTopic(topic string) {
 		fmt.Printf("Failed to create Admin client: %s\n", err)
 		os.Exit(1)
 	}
+
 	// Contexts are used to abort or limit the amount of time
 	// the Admin call blocks waiting for a result.
 	ctx, cancel := context.WithCancel(context.Background())
--- a/examples/confluent_cloud_example/confluent_cloud_example.go
+++ b/examples/confluent_cloud_example/confluent_cloud_example.go
@@ -32,7 +32,7 @@ import (
 	"gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
 )
 
-// In order to set the constants below, you're going to need
+// In order to set the constants below, you are going to need
 // to log in into your Confluent Cloud account. If you choose
 // to do this via the Confluent Cloud CLI, follow these steps.
 
--- a/examples/confluent_cloud_example/confluent_cloud_example.go
+++ b/examples/confluent_cloud_example/confluent_cloud_example.go
@@ -1,19 +1,10 @@
 // This is a simple example demonstrating how to produce a message to
-// Confluent Cloud then read it back again.
+// a topic, and then reading it back again using a consumer. The topic
+// belongs to a Apache Kafka cluster from Confluent Cloud. For more
+// information about Confluent Cloud, please visit:
 //
 // https://www.confluent.io/confluent-cloud/
-//
-// Auto-creation of topics is disabled in Confluent Cloud. You will need to
-// use the ccloud cli to create the go-test-topic topic before running this
-// example.
-//
-// $ ccloud topic create go-test-topic
-//
-// The <ccloud bootstrap servers>, <ccloud key> and <ccloud secret> parameters
-// are available via the Confluent Cloud web interface. For more information,
-// refer to the quick-start:
-//
-// https://docs.confluent.io/current/cloud-quickstart.html
+
 package main
 
 /**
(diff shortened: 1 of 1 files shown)


riferrei:
[clabot:check]

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -90,14 +123,59 @@ func main() {
...
+		"sasl.mechanisms":         "PLAIN",
+		"security.protocol":       "SASL_SSL",
+		"sasl.username":           ccloudAPIKey,
+		"sasl.password":           ccloudAPISecret})
+
+	if err != nil {
+		fmt.Printf("Failed to create Admin client: %s\n", err)
+		os.Exit(1)
+	}
+	// Contexts are used to abort or limit the amount of time
```
edenhill: Add empty line

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -90,14 +123,59 @@ func main() {
...
+	if err != nil {
+		panic("ParseDuration(60s)")
+	}
+	results, err := adminClient.CreateTopics(ctx,
+		[]kafka.TopicSpecification{{
+			Topic:             topic,
+			NumPartitions:     1,
+			ReplicationFactor: 3}},
+		kafka.SetAdminOperationTimeout(maxDuration))
+	if err != nil {
```
edenhill: This error only catches request-level errors, not per-topic errors.
You will need to check the status of the topic in `results` to get per-topic creation errors - the reason is that a CreateTopics() call may succeed for some topics and fail for others.
So something like:
`if results[0].Error.Code() != nil { .. topic creation failed for results[0].Topic }`

examples/confluent_cloud_example/confluent_cloud_example.go
riferrei: Done!

examples/confluent_cloud_example/confluent_cloud_example.go
riferrei: Done. I wrote the following check:

`	// Check for specific topic errors
	for _, result := range results {
		if result.Error.Code() != kafka.ErrNoError {
			fmt.Printf("Topic creation failed for %s: %v",
				result.Topic, result.Error.String())
		}
	}
`

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -90,14 +125,69 @@ func main() {
...
+		kafka.SetAdminOperationTimeout(maxDuration))
+
+	if err != nil {
+		fmt.Printf("Problem during the topic creation: %v\n", err)
+		os.Exit(1)
+	}
+
+	// Check for specific topic errors
+	for _, result := range results {
+		if result.Error.Code() != kafka.ErrNoError {
```
edenhill: What we want to do is exit on error, unless the error is ErrTopicAlreadyExists, in which case we treat it as success.

riferrei: Gotcha! I understand what you're saying. Just pushed the change.

examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -182,9 +182,10 @@ func createTopic(topic string) {
 
 	// Check for specific topic errors
 	for _, result := range results {
-		if result.Error.Code() != kafka.ErrNoError {
+		if result.Error.Code() != kafka.ErrTopicAlreadyExists {
```
edenhill: It needs to check ErrNoError and ErrTopicAlreadyExists :)

riferrei: Can both situation happen? Like, fail because the topic is already there but still... cause an unexpected error during the process? (i.e.: network call error?)

edenhill: There are three possible error states that we care about here:
 1. ErrNoError - topic was created - ✅ 
 2. ErrTopicAlreadyExists - topic already exists - ✅ 
 3. Any other error - fatal error - 👎 

So this if statement should be something like: `if ..Code() != kafka.ErrNoError && ..Code() != ..TopicAlreadyExists { fail.. `

riferrei: Never mind, you're right. In the event of an error (as long this error is not TopicAlreadyExists) we should exit the program because the producer and the consumer will fail anyway.

PR #355, opened 2019-07-15
Scripts for importing librdkafka static bundles

edenhill:
Once this and the [librdkafka PR](https://github.com/edenhill/librdkafka/pull/2413) has been reviewed, we'll make an additional PR with the actual librdkafka import.

The CI tests will not pass until the libs are imported.


README.md
```diff
@@ -161,13 +161,23 @@ Golang import:
 import "gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
 ```
 
-**Note:** that the development of librdkafka and the Go client are kept in synch.
-If you use the master branch of the Go client, then you need to use the master branch of
-librdkafka.
+librdkafka is included with the Go client and does not have to be installed
+separately on the build, nor target, system.
```
mhowlett: "and does not need to be installed separately on the build or target system" i think is better

README.md
```diff
@@ -161,13 +161,23 @@ Golang import:
...
+
+
+**Dynamic linking:** If you prefer to link librdkafka dynamically, which requires
```
mhowlett: maybe: "If you prefer to link librdkafka dynamically you should build your application with `go build -tags dynamic`. This requires librdkafka to be installed on your system.  ..."

mhowlett:
> Once this and the librdkafka PR has been reviewed, we'll make an additional PR with the actual librdkafka import.

are you planning on committing all librdkafka binary builds to the master branch over time (i.e. is that branch going to get very big over time)? I recall you were concerned about size increases if you made a mistake, but i don't understand the plan in the happy case.

i'm not a go practitioner and don't know the specifics, but i wanted to ask if this is an issue, and if there is anything that can be done about it, even if that is un-orthodox.

edenhill:
Each librdkafka import will be on its own branch that is merged, hopefully this will help in keeping the repo download size down when using shallow clones or snapshots, this is actually documented as part of this PR ;) 
https://github.com/confluentinc/confluent-kafka-go/pull/355/files#diff-04fd5d79cdf652f9aeb3198432cfbe44R8

kafka/librdkafka/import.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+#
+#
+# Import a new version of librdkafka based on a librdkafka static bundle.
+# This will create a separate branch, import librdkafka, make a commit,
+# and then ask you to push the branch to github, have it reviewed,
+# and then later merged (NOT squashed or rebased).
+# Having a merge per import allows future shallow clones to skip and ignore
+# older imports, hopefully reducing the amount of git history data 'go get'
+# needs to download.
```
mhowlett: seems like shallow clone isn't going to be a default or option with go get any time soon: https://github.com/golang/go/issues/13078#issuecomment-152062051

(no actionable feedback here, more noting what i'm reading)

mhowlett:
lg from my perspective. minor grammar suggestions

kafka/librdkafka/import.sh
```diff
@@ -0,0 +1,88 @@
...
+
+echo "Checking for existing commits that match this version (should be none)"
+git log --oneline | grep "^librdkafka static bundle $version\$" && exit 1
+
+
+echo "Creating import branch $import_branch"
+git checkout -b $import_branch
+
+echo "Importing bundle $bundle"
+./bundle-import.sh "$bundle" || error_cleanup
```
rnpridgeon: Why not just trap the error?

edenhill: I considered it, but I only wanted this cleanup for these specific errors 🤷‍♂

kafka/librdkafka/import.sh
edenhill: Right, I think there are Go plans to use snapshots rather than clones in the future, which would be equivalent of `--depth 1` minus the .git index (so even smaller).

`git clone --depth 1 .. `  is still an option for people that highly care of import sizes.

I've done some explorative testing and these are the resulting numbers:
 * A standard librdkafka three platform build import is ~8MB per platform, a total of 24MB.
 * This consumes 24MB in the .gitindex and 24MB in the checkout itself = 48MB
 * Compressing the index (which is done by github and by manual `git gc`) brings the index size down to about 8MB
 * Importing 3 more releases, to a total of 4 releases, brings the index size up to 42MB, uncompressed.
 * Compression that index again brings it down to 8MB again. This makes sense since these releases are nearly identical.
 * Performing a shallow clone (`--depth 1`) gives a clone that consumes 17MB: 8MB in the git index, 7MB in the checkout

PR #375, opened 2019-09-26
Simplify Confluent Cloud example config

mhowlett:
these config settings are not needed post v1.0.


examples/confluent_cloud_example/confluent_cloud_example.go
```diff
@@ -74,8 +74,6 @@ func main() {
 	// Produce a new record to the topic...
 	producer, err := kafka.NewProducer(&kafka.ConfigMap{
 		"bootstrap.servers":       bootstrapServers,
-		"broker.version.fallback": "0.10.0.0",
-		"api.version.fallback.ms": 0,
 		"sasl.mechanisms":         "PLAIN",
```
edenhill: Where I have seen this diff before, such deja vu

PR #384, opened 2019-10-18
Fix typo

jeffwidman:


jeffwidman:
Oops, looks like this was already handled in https://github.com/confluentinc/confluent-kafka-go/pull/345

PR #386, opened 2019-10-21
Migrate confluent-kafka-go to Go modules.

rigelbm:
Followed instructions in https://blog.golang.org/migrating-to-go-modules.

Importing as "gopkg.in/confluentinc/confluent-kafka-go.v1/kafka" instead of
the canonical "github.com/confluentinc/confluent-kafka-go/kafka" should
still work as normal. [1]

[1] I am basing the above on the fact that we used to import
"gopkg.in/alecthomas/kingpin.v2" even though it has also been migrated to
modules.


README.md
```diff
@@ -149,21 +149,18 @@ Build from source:
 Install the client
 -------------------
 
-We recommend that you version pin the confluent-kafka-go import to v1:
-
 Manual install:
 ```bash
-go get -u gopkg.in/confluentinc/confluent-kafka-go.v1/kafka
+go get -u github.com/confluentinc/confluent-kafka-go/kafka
```
edenhill: For people on older versions of Go where go mods are not supported we'll still want them to use the versioned import, right?

So I think we'll want to maintain the current instructions but also add a section on using modules.

README.md
```diff
@@ -76,7 +73,7 @@ Producer
 ```golang
 import (
 	"fmt"
-	"gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
```
edenhill: As mentioned in the previous comments; it is too early to assume the user is using Go modules and have that as the default. When not using go modules this import will reference the master branch, which is dangerous.

README.md
```diff
@@ -116,180 +113,170 @@ func main() {
...
 
-Getting Started
-===============
+## Installation
 
-Installing librdkafka
----------------------
+### Installing librdkafka
 
-This client for Go depends on librdkafka v1.1.0 or later, so you either need to install librdkafka
```
edenhill: Don't remove/change this chapter, we shouldn't require people to visit some other page to get full instructions.

README.md
```diff
@@ -116,180 +113,170 @@ func main() {
...
+Install the package from [gopkg.in](https://gopkg.in).
+```bash
+go get -u gopkg.in/confluentinc/confluent-kafka-go.v1/kafka
+```
 
-Install the client
--------------------
+Import the `kafka` package in your code:
+```golang
+import "gopkg.in/confluentinc/confluent-kafka-go/kafka"
```
edenhill: If user does not explicitly call 'go get -u gopkg..' but just 'go get' they will end up with the master branch.

README.md
```diff
@@ -116,180 +113,170 @@ func main() {
...
 ```golang
-import "gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
+import "github.com/confluentinc/confluent-kafka-go/kafka"
 ```
 
-**Note:** that the development of librdkafka and the Go client are kept in synch.
-If you use the master branch of the Go client, then you need to use the master branch of
-librdkafka.
+**Note**: the development of confluent-kafka-go and librdkafka are kept in sync. If you install
+confluent-kafka-go from master, then you will also need to install librdkafka from master.
```
edenhill: "install librdkafka from master" is a bit vague, the original word is clearer.

README.md
```diff
@@ -116,180 +113,170 @@ func main() {
...
 
-See the [examples](examples) for usage details.
+#### Using Go 1.13+ Modules
```
edenhill: This chapter is good, I recommend reverting all the versioning changes above and just adding this chapter.

README.md
```diff
@@ -116,180 +113,170 @@ func main() {
...
-Function Based Consumer
------------------------
+### Function Based Consumer
 
 Messages, errors and events are polled through the consumer.Poll() function.
 
 Pros:
 
- * More direct mapping to underlying librdkafka functionality.
+*   More direct mapping to underlying librdkafka functionality.
```
edenhill: Not sure about this whitespace change, seems pointless

README.md
```diff
@@ -116,180 +113,170 @@ func main() {
...
 
-To link your application statically with librdkafka append `-tags static` to
-your application's `go build` command, e.g.:
+To link your application statically with librdkafka append `-tags static` to your application's
+`go build` command, e.g.:
 
-    $ cd kafkatest/go_verifiable_consumer
-    $ go build -tags static
+```bash
+cd kafkatest/go_verifiable_consumer
```
edenhill: keep the prompt "$ "

README.md
```diff
@@ -116,180 +113,170 @@ func main() {
...
-After a succesful static build verify the dependencies by running
-`ldd ./your_program` (or `otool -L ./your_program` on OSX), librdkafka should not be listed.
+After a succesful static build verify the dependencies by running `ldd ./your_program` (or
+`otool -L ./your_program` on OSX), librdkafka should not be listed.
 
+## Testing
 
-Tests
-=====
+See [kafka/README/Testing](kafka/README.md#testing).
```
edenhill: Not thrilled about this link name

rigelbm:
Friendly Ping @edenhill

go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/confluentinc/confluent-kafka-go
+
+go 1.13
+
+require (
+	github.com/alecthomas/kingpin v2.2.6+incompatible
```
edenhill: These are not needed for the kafka/ module itself, only for the examples.
Can we put the go.mod in kafka/ instead?

examples/admin_create_topic/admin_create_topic.go
```diff
@@ -20,7 +20,7 @@ package main
 import (
 	"context"
 	"fmt"
-	"github.com/confluentinc/confluent-kafka-go/kafka"
+	"gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
```
edenhill: This is a separate, and troublesome change.

Standalone applications should use the gopkg import, but the examples in the repo needs to use the github module url to make sure in-tree examples actually use the in-tree kafka/ package, and not a gopkg.in clone.

edenhill:
> Go documentation recommends single go.mod for entire repository for simplicity. I will follow that.

But we really don't want our users to depend on random packages just because our examples do.

If possible I'd like the go module to reside in kafka/, and leave the rest un-mod-ified (he-he)

kafkatest/go_verifiable_consumer/go_verifiable_consumer.go
```diff
@@ -20,8 +20,8 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/alecthomas/kingpin"
```
edenhill: Why this change? We definitely want to pin the version on this one.

kafkatest/go_verifiable_producer/go_verifiable_producer.go
```diff
@@ -20,8 +20,8 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/alecthomas/kingpin"
```
edenhill: dito

rigelbm:
> > Go documentation recommends single go.mod for entire repository for simplicity. I will follow that.
> 
> But we really don't want our users to depend on random packages just because our examples do.
> 
> If possible I'd like the go module to reside in kafka/, and leave the rest un-mod-ified (he-he)

See https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository.

TL;DR: There's no harm on the very small extra dependencies we have for examples, there's possible harm in having a complicated module structure.

I would very much like to keep it this way.

rigelbm:
PTAL

go.mod
travisjeffery: I'd make the examples their own module to separate them from the top-level module so their dependencies are separate. This is a perfect use case for it. If I was using the client's module and saw that it imported kingpin but didn't seem to use it, I wouldn't like it.

rigelbm: Okay. I will do:

```
confluent-kafka-go/
    go.mod
    examples/
        go.mod
    kafkatest/
        go.mod
``` 

This way, the import path of `kafka` will stay the same `github.com/confluent-kafka-go/vX/kafka`, instead of changing to `github.com/confluent-kafka-go/kafka/vX` (what would happen if we put the go.mod in `kafka` instead of top-level.

rigelbm:
PTAL

README.md
```diff
@@ -167,6 +167,26 @@ librdkafka.
...
+Starting with Go 1.13, you can use [Go Modules](https://blog.golang.org/using-go-modules) to install
+confluent-kafka-go.
+
+Import the `kafka` package from GitHub in your code:
+
+```golang
+import "github.com/confluentinc/confluent-kafka-go/kafka"
+```
+
+Compile:
```
edenhill: Should probably say "Build your project:", since it is somewhat ambigious now what needs to be compiled ("kafka" or my thing?)

go.mod
edenhill: Is that correct though?
What we're interested in is the api version 1 of kafka/, so in my mind "github.com/confluent-kafka-go/kafka/v1" makes more sense, rather than the kafka api of version 1.

README.md
rigelbm: Done.

go.mod
rigelbm: As discussed offline, the module has the version, not the package. `github.com/confluent-kafka-go/vX/kafka` means "The `kafka` package of the `vX` version of the `confluent-kafka-go` module".

rigelbm:
PTAL

Issue #319, opened 2019-03-31
Question - determine current controller broker id


fritchie:
I see that I can get a broker list with GetMetadata but is there a way to determine which broker is the current controller?

edenhill:
There is a librdkafka API, [rd_kafka_controllerid](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L2295), that can be used to acquire the controller id, but it is not (yet) exposed as a method in the Go client.

edenhill:
Add ClusterId() and ControllerId()

rigelbm:
Added the APIs for ClusterID and ControllerID. They should be available in v1.2.0.

edenhill:
v1.2.2 :)

PR #390, opened 2019-10-30
#319: Add ClusterID and ControllerID to AdminClient.

rigelbm:
Issue: https://github.com/confluentinc/confluent-kafka-go/issues/319

### DIFF
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -421,7 +421,7 @@ func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t
 }
 
 
-// Returns the ClusterId as reported in broker metadata.
+// ClusterId returns the cluster ID as reported in broker metadata.
 //
 // Requires broker version >=0.10.0 and api.version.request=true.
 func (a *AdminClient) ClusterID(timeout time.Duration) (clusterID string, err error) {
@@ -437,7 +437,7 @@ func (a *AdminClient) ClusterID(timeout time.Duration) (clusterID string, err er
 	return clusterID, nil
 }
 
-// Returns the current ControllerId as reported in broker metadata.
+// ControllerID returns the current broker ID of the controller as reported in broker metadata.
 //
 // Requires broker version >=0.10.0 and api.version.request=true.
 func (a *AdminClient) ControllerID(timeout time.Duration) (controllerID int32, err error) {
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -421,30 +421,35 @@ func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t
 }
 
 
+
 // ClusterID returns the cluster ID as reported in broker metadata.
 //
+// Returns an error C.RD_KAFKA_RESP_ERR__TIMED_OUT if the operation times out.
+//
 // Requires broker version >=0.10.0 and api.version.request=true.
-func (a *AdminClient) ClusterID(timeout time.Duration) (clusterID string, err error) {
-	cClusterID := C.rd_kafka_clusterid(a.handle.rk, C.int(timeout/time.Millisecond))
-	defer C.free(unsafe.Pointer(cClusterID))
+func (a *AdminClient) ClusterID(ctx context.Context) (clusterID string, err error) {
+	cClusterID := C.rd_kafka_clusterid(a.handle.rk, C.int(Timeout(ctx)/time.Millisecond))
 
 	if cClusterID == nil {
-		err = newError(C.RD_KAFKA_RESP_ERR_REQUEST_TIMED_OUT)
+		err = newError(C.RD_KAFKA_RESP_ERR__TIMED_OUT)
 		return "", err
 	}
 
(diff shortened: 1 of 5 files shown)


kafka/integration_test.go
```diff
@@ -1559,3 +1559,57 @@ func TestAdminGetMetadata(t *testing.T) {
...
+	}
+	defer admin.Close()
+
+	controllerID, err := admin.ControllerID(time.Second)
+	if err != nil {
+		t.Errorf("Failed to get ControllerID: %s\n", err)
+		return
+	}
+
+	t.Logf("ControllerID: %d\n", controllerID)
```
edenhill: Verify that controllerID is valid (>=0)

kafka/integration_test.go
```diff
@@ -1559,3 +1559,57 @@ func TestAdminGetMetadata(t *testing.T) {
...
+	}
+	defer admin.Close()
+
+	clusterID, err := admin.ClusterID(time.Second)
+	if err != nil {
+		t.Errorf("Failed to get ClusterID: %s\n", err)
+		return
+	}
+
+	t.Logf("ClusterID: %s\n", clusterID)
```
edenhill: verify that clusterid is valid (printable non-empty string)

kafka/adminapi.go
```diff
@@ -420,6 +420,37 @@ func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t
...
+// ClusterID returns the cluster ID as reported in broker metadata.
+//
+// Requires broker version >=0.10.0 and api.version.request=true.
+func (a *AdminClient) ClusterID(timeout time.Duration) (clusterID string, err error) {
+	cClusterID := C.rd_kafka_clusterid(a.handle.rk, C.int(timeout/time.Millisecond))
+	defer C.free(unsafe.Pointer(cClusterID))
+
+	if cClusterID == nil {
+		err = newError(C.RD_KAFKA_RESP_ERR_REQUEST_TIMED_OUT)
+		return "", err
```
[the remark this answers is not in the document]
rigelbm: Empty string is the zero value of type string, per language specification.

kafka/adminapi.go
```diff
@@ -421,30 +421,35 @@ func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t
 }
 
 
+
 // ClusterID returns the cluster ID as reported in broker metadata.
 //
+// Returns an error C.RD_KAFKA_RESP_ERR__TIMED_OUT if the operation times out.
```
edenhill: Use the Go type name, we don't want to expose librdkafka or its types.
See generated_errors.go

kafka/context.go
```diff
@@ -0,0 +1,30 @@
...
+
+package kafka
+
+import (
+	"context"
+	"time"
+)
+
+// Timeout returns the remaining timeout in the given context, or 0 if no deadline/timeout was set.
+func Timeout(ctx context.Context) (timeout time.Duration) {
```
edenhill: How about calling this TimeoutMs and do the ms conversion in it since that's how it will be used?

Also, there are three different types of timeout in librdkafka interfaces:
 -1 - indefinite timeout (block forever)
  0 - no timeout, non-blocking
 >0 - an actual timeout

This function should handle all three cases.

And passing a non-deadline context should convert to -1 (indeifnite timeout)

kafka/integration_test.go
```diff
@@ -1559,3 +1559,57 @@ func TestAdminGetMetadata(t *testing.T) {
...
+		t.Skipf("Missing testconf.json")
+	}
+
+	config := &ConfigMap{"bootstrap.servers": testconf.Brokers}
+	if err := config.updateFromTestconf(); err != nil {
+		t.Errorf("Failed to update test configuration: %s\n", err)
+		return
+	}
+
+	admin, err := NewAdminClient(config)
```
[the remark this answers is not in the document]
edenhill: Not addressed.
See NewAdminClientFromProducer()

kafka/context.go
rigelbm: Created a separate cTimeoutFromContext function.

kafka/adminapi.go
```diff
@@ -420,6 +420,42 @@ func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t
 	return result, nil
 }
 
+
+
+// ClusterID returns the cluster ID as reported in broker metadata.
+//
+// Returns an error C.RD_KAFKA_RESP_ERR__TIMED_OUT if the operation times out.
+//
```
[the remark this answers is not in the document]
rigelbm: Added note and better handling. Right now I'm blocking on cancellation. Another option is to wait for it asynchronously. I.e.:

```
case <-ctx.Done():
	go func() {
		if cClusterID := <-responseChan; cClusterID != nil {
			C.rd_kafka_mem_free(a.handle.rk, unsafe.Pointer(cClusterID))
		}
	}()
	return "", ctx.Err()
```

That would be "behave" like it cancelled, even if it is waiting in the background. Let me know what you think.

PR #397, opened 2019-11-20
Add batch consume method to consumer

tburmeister:
This still needs some work, but wanted to get thoughts on the general idea. Goal here is to add a batch consume method, already available in the C and Python clients, for more efficient consumption of messages. I've included an example of usage, as well as two examples for benchmarking the batch consumer against the current `Poll` method. In my benchmarks, Saw the following speedups for consuming 1M x 512B messages from a local Kafka server:
```
poll:
real    0m4.138s
user    0m1.471s
sys    0m0.307s

batch consume:
real    0m0.721s
user    0m0.871s
sys    0m0.208s
```

There are currently two pretty big limitations to my implementation:
1. You must turn off auto offsets commits;
2. You must use the `Assign` method to manually assign topic-partitions; you can't use the high level `Subscribe` method.
Basically, the commit and assignment callbacks cause a panic.


lexasoft123:
@edenhill might be this PR can resolve CGO bottleneck? :)

syepes:
Is this still in active work? it would be nice to also have this on the Go client

PR #412, opened 2019-12-16
Prepare v1.3.0 release.

rigelbm:
I had to adapt the doc-gen.py script, since godoc does not work properly with Go Modules (https://github.com/golang/go/issues/26827).


kafka/README.md
```diff
@@ -126,6 +126,8 @@ Also make sure the examples in README.md work.
 Convert any examples using `github.com/confluentinc/confluent-kafka-go/kafka` to use
 `gopkg.in/confluentinc/confluent-kafka-go.v1/kafka` import path.
 
+    $ find examples/ -type f -name "*\.go" -exec sed -i -e "s/github\.com\/confluentinc\/confluent-kafka-go\/kafka/gopkg\.in\/confluentinc\/confluent-kafka-go\.v1\/kafka/" {} +
```
edenhill: tip: use another regex substituion separator when working with paths, e.g::
`sed -i -e 's,github.com/confluentinc/confluent-kafka-go/....,some-replacement,g'`

micro nit: use single-quotes instead of double-quotes
nit: should replace all occurrences since we can't know if the import is the first one, so use `s,...,..,g`

edenhill: what's the trailing + for?

kafka/generated_errors.go
```diff
@@ -1,6 +1,6 @@
 package kafka
 // Copyright 2016 Confluent Inc.
-// AUTOMATICALLY GENERATED BY /Users/magnus/gocode/bin/go_rdkafka_generr ON 2019-07-15 15:16:46.48088 +0200 CEST m=+0.000708503 USING librdkafka 1.1.0
+// AUTOMATICALLY GENERATED BY /home/rigelbm/go/bin/go_rdkafka_generr ON 2019-12-16 13:32:50.652785506 +0000 GMT m=+0.000071676 USING librdkafka 1.2.2-RC1-35-g4ffe54
```
edenhill: Looks like you did not generate this with librdkafka v1.3.0, or your librdkafka was not a tagged build (`git fetch --tags && ./configure --reconfigure && make clean && make -j && sudo make install`)

mk/doc-gen.py
```diff
@@ -34,4 +37,5 @@
         t['src'] = '//golang.org' + t['src']
 
     # Write updated HTML to stdout
-    print(soup.prettify().encode('utf-8'))
+    print(soup.prettify())
```
edenhill: Why'd you drop the utf8 conversion?
Does this still work on both py2 and p3?

kafka/README.md
rigelbm: `-exec cat {} +` results in `cat foo bar`, while `-exec cat {} \;` results in `cat foo; cat bar`, i.e. + uses a single sed instead of multiple. Does not matter.

mk/doc-gen.py
rigelbm: `print(string.encode('utf-8'))` prints a bytes literal in Python 3. It works in Python 3 now (I think it won't in Python 2 though).

rigelbm:
> Need to update README and in-code dependencies to librdkafka v1.3.0 since the Go client relies on new error codes (I missed this last time we had this discussion).

Done.

PR #430, opened 2020-02-21
Added "go.logs.channel" property for passing an existing channel to log to

edenhill:
.. and some refactoring.
.. and include the client instance name in the LogEvent.
.. and cache the client instance name to avoid CGo call for each stringification.

### DIFF
--- a/kafka/config.go
+++ b/kafka/config.go
@@ -237,6 +237,32 @@ func (m ConfigMap) extract(key string, defval ConfigValue) (ConfigValue, error) 
 	return v, nil
 }
 
+// extractLogConfig extracts generic go.logs.* configuration properties.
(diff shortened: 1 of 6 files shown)


kafka/config.go
```diff
@@ -237,6 +237,32 @@ func (m ConfigMap) extract(key string, defval ConfigValue) (ConfigValue, error)
 	return v, nil
 }
 
+// extractLogConfig extracts generic go.logs.* configuration properties.
+func (m ConfigMap) extractLogConfig() (logsChanEnable bool, logsChan chan LogEvent, err error) {
```
rigelbm: I would call this `createLogChannel` as it is not only extracting the log config, but also has the side effect of creating the channel itself. Side effects should always be explicit, I think.

edenhill: It doesn't, the logs channel is created later from setupLogQueue()

PR #431, opened 2020-02-21
Transactional API

edenhill:
This should be complete short of doc strings for the public API (since they will need to be finalized in librdkafka first).

Depends on https://github.com/edenhill/librdkafka/pull/2728


examples/transactions_example/generator.go
```diff
@@ -0,0 +1,124 @@
...
+	"fmt"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+// Intersections this application will process.
+var Intersections = [...]string{
+	"Stockholmsvagen,Uppsalavagen",
```
mhowlett: very diplomatic

examples/transactions_example/processor.go
```diff
@@ -0,0 +1,354 @@
...
+	if istate.currGreen != "" {
+		// Let all cars on the green road pass thru
+		istate.carCnt -= istate.carsWaiting[istate.currGreen]
+		istate.carsWaiting[istate.currGreen] = 0
+	}
+
+	return true
+}
+
+// FIXME
```
mhowlett: fixme

examples/transactions_example/processor.go
```diff
@@ -0,0 +1,354 @@
...
+		fatal(err)
+	}
+
+	addLog(fmt.Sprintf("Processor: waiting for messages on topic %s",
+		inputTopic))
+	for !doTerm {
+		select {
+
+		case <-ticker.C:
+			// Run intersection state machine(s) every second.
```
mhowlett: every second -> periodically (?). the ticker is 500ms.

though pedantically, periodically is not really right either because of the consumer poll with a timeout.

examples/transactions_example/processor.go
```diff
@@ -0,0 +1,354 @@
...
+			for partition := range partitionsToCommit {
+				commitTransactionForInputPartition(partition)
+			}
+
+		case <-termChan:
+			doTerm = true
+
+		default:
+			// Poll consumer for new messages or rebalance events.
+			ev := processorConsumer.Poll(100)
```
mhowlett: seems fine, but would it be a bit better to have poll have a timeout of `0` and have the ticker interval drive everything?

examples/transactions_example/processor.go
```diff
@@ -0,0 +1,354 @@
...
+		}
+	}
+
+	return Roads[0]
+}
+
+// processIngressCarMessage processes an input message of a car coming
+// into an intersection.
+func processIngressCarMessage(msg *kafka.Message) {
+	if len(producers) == 0 {
```
mhowlett: is this really possible at the application level?
if so, why is this possible? if not, this should be fatal.

examples/transactions_example/processor.go
```diff
@@ -0,0 +1,354 @@
...
+		// Outdated message from last rebalance;
+		// no producers are available.
+		return
+	}
+
+	if msg.Key == nil || msg.Value == nil {
```
mhowlett: why check for this? seems a bit arbitrary

examples/transactions_example/processor.go
```diff
@@ -0,0 +1,354 @@
...
+	partition   int32             // Input partition
+	lightState  map[string]string // Current light states, indexed by road
+	carsWaiting map[string]int    // Numbers of cars waiting, indexed by road
+	carCnt      int               // Total number of cars waiting
+	currGreen   string            // Currently green road
+	lastChange  time.Time         // Time of last light change
+}
+
+// intersectionStates maintains state per intersection
+var intersectionStates map[string]*intersectionState
```
mhowlett: this state is not persisted, but that seems unwise because (if i understood the code logic correctly) in the worst case it could result in some v. unhappy cars waiting at intersections forever.

examples/transactions_example/visualizer.go
```diff
@@ -0,0 +1,456 @@
...
+		}
+	}
+
+	drawText(s, drawPos{pos1.x + 2, pos1.y}, name, false)
+}
+
+// drawRoads draws the roads in the intersection and returns the
+// light and eligible car positions keyed by road.
+func drawRoads(s tcell.Screen, pos drawPos) (lightPos map[string]drawPos,
+	lanes map[string][]drawPos) {
```
mhowlett: wasn't going to bother compiling and running this, and then I saw this method.

kafka/producer.go
```diff
@@ -687,3 +688,95 @@ func (p *Producer) SetOAuthBearerToken(oauthBearerToken OAuthBearerToken) error
 func (p *Producer) SetOAuthBearerTokenFailure(errstr string) error {
 	return p.handle.setOAuthBearerTokenFailure(errstr)
 }
+
+// Transactional API
+
+// InitTransactions ..
+func (p *Producer) InitTransactions(ctx context.Context) error {
```
mhowlett: I think you forgot to add docs here and on the other txn methods.

examples/transactions_example/processor.go
edenhill: We want to consume messages as quickly as possible, blocking on the ticker would only consume when the ticker fires, which I think is atypical.

examples/transactions_example/processor.go
edenhill: You're right, the example used the channel based consumer before which has this problem of outdated messages, but now it uses the polling consumer and this is no longer a problem.

examples/transactions_example/processor.go
edenhill: To avoid a null dereference, we don't control what's coming on the input topic.

examples/transactions_example/processor.go
edenhill: Persisted across rebalances you mean? Correct, I wanted to provide a simple state-less processor example.

In the case no new input messages (cars) are coming in, the state machine will arbiter over the traffic lights anyway and traffic will pass, just not optimized for queue depth.

examples/transactions_example/txnhelpers.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		fatal(err)
+	}
+
+	return position
+}
+
+// commitTransactionForInputPartition sends the consumer offsets for
+// the given input partition and commits the current transaction.
+// A new transaction will be started when done.
```
rigelbm: nit: Add an extra line between function short-description and long-description.

examples/transactions_example/txnhelpers.go
```diff
@@ -0,0 +1,221 @@
...
+func commitTransactionForInputPartition(partition int32) {
+
```
rigelbm: nit: Remove extra new line.

examples/transactions_example/txnhelpers.go
```diff
@@ -0,0 +1,221 @@
...
+	if !found || producer == nil {
+		fatal(fmt.Sprintf("BUG: No producer for input partition %v", partition))
+	}
+
+	position := getConsumerPosition(partition)
+	consumerMetadata := processorConsumer.GetConsumerGroupMetadata()
+	defer consumerMetadata.Close()
+
+	err := producer.SendOffsetsToTransaction(nil, position, consumerMetadata)
+	if err != nil {
```
rigelbm: Can you limit the scope of `err`? Example:

```
if err := producer.SendOffsetsToTransaction(nil, position, consumerMetadata); err != null {
        // ...
}
```

examples/transactions_example/txnhelpers.go
```diff
@@ -0,0 +1,221 @@
...
+		addLog(fmt.Sprintf(
+			"Processor: Failed to send offsets to transaction for input partition %v: %s: aborting transaction",
+			partition, err))
+
+		err = producer.AbortTransaction(nil)
+		if err != nil {
```
rigelbm: Limit scope of `err`.

```
if err := producer.AbortTransaction(nil); err != nil {
        // ...
}
```

examples/transactions_example/txnhelpers.go
```diff
@@ -0,0 +1,221 @@
...
+			fatal(err)
+		}
+
+		// Rewind this input partition to the last committed offset.
+		rewindConsumerPosition(partition)
+	} else {
+		err = producer.CommitTransaction(nil)
+		if err != nil {
```
rigelbm: Limit scope of `err`.

```
if err := producer.CommitTransaction(nil); err != nil {
        // ...
}
```

examples/transactions_example/txnhelpers.go
```diff
@@ -0,0 +1,221 @@
...
+			addLog(fmt.Sprintf(
+				"Processor: Failed to commit transaction for input partition %v: %s",
+				partition, err))
+
+			err = producer.AbortTransaction(nil)
+			if err != nil {
```
rigelbm: Limit scope of `err`.

examples/transactions_example/txnhelpers.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+
+			// Rewind this input partition to the last committed offset.
+			rewindConsumerPosition(partition)
+		}
+	}
+
+	// Start a new transaction
+	err = producer.BeginTransaction()
+	if err != nil {
```
rigelbm: Limit scope of `err`.

examples/transactions_example/visualizer.go
```diff
@@ -0,0 +1,456 @@
...
+	"github.com/confluentinc/confluent-kafka-go/kafka"
+	"github.com/gdamore/tcell"
+	"os"
+	"sort"
+	"sync"
+	"time"
+)
+
+// height and width per intersection frame
+const hpi = 20
```
rigelbm: nit: Maybe rename `hpi` to `height`.

examples/transactions_example/visualizer.go
```diff
@@ -0,0 +1,456 @@
...
+	"fmt"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
+	"github.com/gdamore/tcell"
+	"os"
+	"sort"
+	"sync"
+	"time"
+)
+
+// height and width per intersection frame
```
rigelbm: nit: include measurement unit (e.g. `pixels`).

examples/transactions_example/visualizer.go
```diff
@@ -0,0 +1,456 @@
...
+const hpi = 20
+const wpi = 39
```
rigelbm: nit: Maybe rename `wpi` to `width`.

examples/transactions_example/visualizer.go
```diff
@@ -0,0 +1,456 @@
...
+
+var screen tcell.Screen
+
+type drawPos struct {
```
rigelbm: nit: s/`drawPos`/`drawPosition`/ (or just `position`)

rigelbm: Add documentation clarifying if this is left-right or right-left and bottom-top or top-bottom.

examples/transactions_example/visualizer.go
```diff
@@ -0,0 +1,456 @@
...
+	x int
+	y int
+}
+
+const showLogCnt = 20
```
rigelbm: nit: s/`showLogCnt`/`showLogCount`/

examples/transactions_example/visualizer.go
```diff
@@ -0,0 +1,456 @@
...
+
+var collectedLogs [showLogCnt]string
+var logIndex int
+var logPos drawPos
```
rigelbm: nit: s/`logPos`/`logPosition`/

examples/transactions_example/visualizer.go
```diff
@@ -0,0 +1,456 @@
...
+var logMutex sync.Mutex
+
+// drawText draws the text string at the given position
```
rigelbm: nit: Missing period at the end.

examples/transactions_example/visualizer.go
```diff
@@ -0,0 +1,456 @@
...
+func drawText(s tcell.Screen, pos drawPos, text string, clearPad bool) {
```
rigelbm: nit: s/`s`/`screen`/ s/`pos`/`position`/

kafka/consumer.go
```diff
@@ -576,6 +576,23 @@ func (c *Consumer) Committed(partitions []TopicPartition, timeoutMs int) (offset
 	return newTopicPartitionsFromCparts(cparts), nil
 }
 
+// Position returns the current consume position for the given partitions.
+// Typical use is to call Assignment() to get the partition list
+// and then pass it to Position() to get the current consume position for
+// each of the assigned partitions.
+// The conusme position is the next message to read from the partition.
```
rigelbm: nit: s/`conusme`/`consume`/

kafka/consumer.go
```diff
@@ -576,6 +576,23 @@ func (c *Consumer) Committed(partitions []TopicPartition, timeoutMs int) (offset
...
+// i.e., the offset of the last message seen by the application + 1.
```
rigelbm: I would rephrase as `last message consumed + 1` instead of the `last message seen by the application + 1`, since there's no guarantee the last message of the given partition was read by this application (e.g. if this application just started).

kafka/consumer.go
```diff
@@ -576,6 +576,23 @@ func (c *Consumer) Committed(partitions []TopicPartition, timeoutMs int) (offset
...
+func (c *Consumer) Position(partitions []TopicPartition) (offsets []TopicPartition, err error) {
+	cparts := newCPartsFromTopicPartitions(partitions)
+	defer C.rd_kafka_topic_partition_list_destroy(cparts)
+	cerr := C.rd_kafka_position(c.handle.rk, cparts)
+	if cerr != C.RD_KAFKA_RESP_ERR_NO_ERROR {
```
rigelbm: nit: Limit scope of `cerr`.

kafka/consumer.go
```diff
@@ -626,3 +643,37 @@ func (c *Consumer) SetOAuthBearerToken(oauthBearerToken OAuthBearerToken) error
...
+
+// ConsumerGroupMetadata reflects the current consumer group member metadata.
+type ConsumerGroupMetadata struct {
+	cgmd *C.rd_kafka_consumer_group_metadata_t
+}
+
+// GetConsumerGroupMetadata returns the consumer's current group metadata.
+// This object should be passed to the transactional producer's
+// SendOffsetsToTransaction() API and then destroyed with metadata.Close().
+func (c *Consumer) GetConsumerGroupMetadata() *ConsumerGroupMetadata {
```
rigelbm: It sounds strange that you need to do a `close()` on something you do `get()` to acquire. Getters usually return things owned by something else, instead of transferring ownership of it to the caller. What is the semantics of multiple `GetConsumerGroupMetadata()` calls? Would you need call `Close()` for each one? Do they return the same object (assuming nothing changed underneath)?

kafka/consumer.go
```diff
@@ -626,3 +643,37 @@ func (c *Consumer) SetOAuthBearerToken(oauthBearerToken OAuthBearerToken) error
...
+	if cg.cgmd != nil {
+		C.rd_kafka_consumer_group_metadata_destroy(cg.cgmd)
+		cg.cgmd = nil
+	}
+}
+
+// NewConsumerGroupMetadata creates a new consumer group metadata instance
+// for testing use.
+// The returned object must be destroyed with metadata.Close().
+func NewConsumerGroupMetadata(groupID string) *ConsumerGroupMetadata {
```
rigelbm: I would rename this function as `NewTestConsumerGroupMetadata` to make it even more explicit this is only for tests. If I were using an IDE with code-completion (e.g. Goland) and I was trying to find out how do I get a `ConsumerGroupMetadata`, `NewConsumerGroupMetadata` would be the first thing I would try.

kafka/error.go
```diff
@@ -33,22 +33,24 @@ import (
...
 // Error provides a Kafka-specific error container
 type Error struct {
-	code  ErrorCode
-	str   string
-	fatal bool
+	code         ErrorCode
+	str          string
+	fatal        bool
+	retriable    bool
+	txnAbortable bool
```
rigelbm: nit: s/`txnAbortable`/`transactionAbortable`/

kafka/log.go
```diff
@@ -48,22 +48,14 @@ func (h *handle) newLogEvent(cEvent *C.rd_kafka_event_t) LogEvent {
...
 		case <-doneChan:
-			if cEvent := <-pollChan; cEvent != nil {
-				C.rd_kafka_event_destroy(cEvent)
-			}
 			return
 
-		case cEvent := <-pollChan:
-			if cEvent == nil { // C timeout
+		default:
+			cEvent := C.rd_kafka_queue_poll(h.logq, C.int(timeoutMs))
```
rigelbm: Rumbling: I must say I prefer handling doneChan signal during native call explicitly, but this will lead to the same behaviour.

kafka/consumer.go
edenhill: Position() returns the last message +1 passed to this application instance, I think you are referring to Committed() which is the last committed message + 1.

kafka/consumer.go
edenhill: Aye, fixed.

kafka/log.go
edenhill: I'm not sure I understand what you are proposing?

PR #433, opened 2020-02-26
Improve Readme Examples with error handling

silviolleite:


README.md
```diff
@@ -55,15 +55,18 @@ func main() {
 		panic(err)
 	}
 
-	c.SubscribeTopics([]string{"myTopic", "^aRegex.*[Tt]opic"}, nil)
+	err = c.SubscribeTopics([]string{"myTopic", "^aRegex.*[Tt]opic"}, nil)
+        if err != nil {
+            fmt.Printf("Error subscribing to topic: %v\n", err)
```
edenhill: This should be a terminal error, maybe use panic() or a return(1) instead?

README.md
```diff
@@ -105,10 +108,13 @@ func main() {
 	// Produce messages to topic (asynchronously)
 	topic := "myTopic"
 	for _, word := range []string{"Welcome", "to", "the", "Confluent", "Kafka", "Golang", "client"} {
-		p.Produce(&kafka.Message{
+		err = p.Produce(&kafka.Message{
 			TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny},
 			Value:          []byte(word),
 		}, nil)
+        if err != nil {
```
edenhill: indent should be same as the Produce() call

silviolleite: Done!

README.md
```diff
@@ -105,10 +108,13 @@ func main() {
 	// Produce messages to topic (asynchronously)
 	topic := "myTopic"
 	for _, word := range []string{"Welcome", "to", "the", "Confluent", "Kafka", "Golang", "client"} {
-		p.Produce(&kafka.Message{
+		err = p.Produce(&kafka.Message{
 			TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny},
 			Value:          []byte(word),
 		}, nil)
+                if err != nil {
+                    fmt.Printf("Message could not be enqueued: %v\n", err)
```
jeffwidman: Would `fmt.Println("Message could not be enqueued: ", err)` be more readable and concise?

silviolleite: All other impressions are using `fmt.Printf()`, I kept the same.

jeffwidman: Yes, because they need to interpolate values in the middle of their strings... here you are only appending a value, so you don't need `printf` and friends.

edenhill: Please use fmt.Printf() since it is consistent with other code, and more explicit what formatting is desired.

silviolleite: Yes, make sence. I fixed it.

README.md
```diff
@@ -55,15 +55,18 @@ func main() {
 		panic(err)
 	}
 
-	c.SubscribeTopics([]string{"myTopic", "^aRegex.*[Tt]opic"}, nil)
+	err = c.SubscribeTopics([]string{"myTopic", "^aRegex.*[Tt]opic"}, nil)
+        if err != nil {
```
edenhill: mix of tabs and space here, use tabs as that is what was used previously.

README.md
```diff
@@ -105,10 +108,13 @@ func main() {
 	// Produce messages to topic (asynchronously)
 	topic := "myTopic"
 	for _, word := range []string{"Welcome", "to", "the", "Confluent", "Kafka", "Golang", "client"} {
-		p.Produce(&kafka.Message{
+		err = p.Produce(&kafka.Message{
 			TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny},
 			Value:          []byte(word),
 		}, nil)
+                if err != nil {
+                    fmt.Println("Message could not be enqueued: ", err)
```
edenhill: Use fmt.Printf() to be consistent with existing code.

README.md
silviolleite: Done!

README.md
silviolleite: Done!

silviolleite: Done

README.md
edenhill: It still has whitespaces for the new code and tabs for the old.

README.md
edenhill: This is also using whitespaces rather than tabs.

silviolleite:
Waiting for review :(

PR #435, opened 2020-03-05
Apply default.topic.config before entering the config conversion loop

meweltman:
This fixes an issue where `default.topic.config` is inconsistently applied. An example might be
```
ConfigMap{
	"partitioner": "murmur2_random",
	"default.topic.config": ConfigMap{
		"acks": "-1",
	},
}
```
Depending on the order of the keys in the loop, the partitioner might be either `murmur2_random` or `consistent_random`.


edenhill:
Note: default.topic.config is deprecated and will be removed in a future release (v2.0 to not break semver).

meweltman:
> Note: default.topic.config is deprecated and will be removed in a future release (v2.0 to not break semver).

👍 We actually stopped using this feature on our end to avoid this bug.

PR #441, opened 2020-03-26
Refine offsetsForTimes doc string

mhowlett:
assuming the html docs will get auto-generated in due course

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -514,7 +514,8 @@ func (c *Consumer) GetWatermarkOffsets(topic string, partition int32) (low, high
 //
 // The returned offset for each partition is the earliest offset whose
 // timestamp is greater than or equal to the given timestamp in the
-// corresponding partition.
+// corresponding partition. If the provided timestamp exceeds that of the
+// last message in the partition, a value of -1 will be returned.
 //
 // The timestamps to query are represented as `.Offset` in the `times`
 // argument and the looked up offsets are represented as `.Offset` in the returned
--- a/kafka/offset.go
+++ b/kafka/offset.go
@@ -122,7 +122,8 @@ func OffsetTail(relativeOffset Offset) Offset {
 //
 // The returned offset for each partition is the earliest offset whose
 // timestamp is greater than or equal to the given timestamp in the
-// corresponding partition.
+// corresponding partition. If the provided timestamp exceeds that of the
+// last message in the partition, a value of -1 will be returned.
 //
 // The timestamps to query are represented as `.Offset` in the `times`
 // argument and the looked up offsets are represented as `.Offset` in the returned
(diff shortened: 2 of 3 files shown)


mhowlett:
when I try to run that command, i get the error below.

If I just run `godoc` and browse to `localhost:6060/pkg/github.com/confluentinc/confluent-kafka-go/kafka` i can see the changes reflect this PR. If i give godoc any argument other than `-http` (including `-url`), I get the `Missing args.` error. I'm on an ubuntu 19.04 derivative, go v1.10.4, godocs installed via `sudo apt-get install golang-go.tools`.

any idea what's up? a quick web search wasn't fruitful. i think you should just generate the docs, i've wasted enough time on it already.

```
matt@pop-os:~/go/src/github.com/confluentinc/confluent-kafka-go$ make -f mk/Makefile docs
(cd kafka/go_rdkafka_generr && go install)
(cd kafka && go generate ./...)
mk/doc-gen.py > kafka/api.html
Missing args.
usage: godoc -http=localhost:6060
  -analysis string
    	comma-separated list of analyses to perform (supported: type, pointer). See http://golang.org/lib/godoc/analysis/help.html
  -goroot string
    	Go root directory (default "/usr/lib/go-1.10")
  -http string
    	HTTP service address (default "localhost:6060")
  -index
    	enable search index
  -index_files string
    	glob pattern specifying index files; if not empty, the index is read from these files in sorted order
  -index_interval duration
    	interval of indexing; 0 for default (5m), negative to only index once at startup
  -index_throttle float
    	index throttle value; 0.0 = no time allocated, 1.0 = full throttle (default 0.75)
  -links
    	link identifiers to their declarations (default true)
  -maxresults int
    	maximum number of full text search results shown (default 10000)
  -notes string
    	regular expression matching note markers to show (default "BUG")
  -play
    	enable playground
  -templates string
    	load templates/JS/CSS from disk in this directory
  -timestamps
    	show timestamps with directory listings
  -url string
    	print HTML for named URL
  -v	verbose mode
  -write_index
    	write index to a file; the file name must be specified with -index_files
  -zip string
    	zip file providing the file system to serve; disabled if empty
Traceback (most recent call last):
  File "mk/doc-gen.py", line 16, in <module>
    'godoc -url=/pkg/github.com/confluentinc/confluent-kafka-go/kafka | grep -v "^using module mode"', shell=True)
  File "/usr/lib/python2.7/subprocess.py", line 223, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command 'godoc -url=/pkg/github.com/confluentinc/confluent-kafka-go/kafka | grep -v "^using module mode"' returned non-zero exit status 1
make: *** [mk/Makefile:16: docs] Error 1

```

edenhill:
You need to install godoc through 'go get' rather than the distro package manager.

but I'll get the docs updated later in the release cycle.

Thanks!

Issue #481, opened 2020-06-22
docs: At least once processing guarantee


wedneyyuri:
Description
===========

Hi guys, thanks for the work and the great tools you have produced for the Kafka community.

Today I am trying to understand what I need to do to guarantee **at least once** processing using Kafka. As far as I know the autocommit functionality is enabled by default and so if there is a problem in the message processing (for example a latency at the other end) the offset still will be updated (respecting `auto.commit.interval.ms`) and the message will be lost forever in case of process failure (for example a panic).

Generally the solutions I see to work around this problem involve manually managing the position of the offset, however I am having trouble finding examples of how to solve this using this golang package.

Proposal
===========

What I have in mind:

- Receive 10,000 messages
- Process the 10000 messages (if it is necessary to parallelize more I can use goroutines, channels, waitGroups, ...)
- Update the offsets
- Receive the next 10,000 messages and start over

Further details about this behavior that I'm trying to explain are available on the newrelic blog: https://blog.newrelic.com/engineering/kafka-consumer-config-auto-commit-data-loss/

wedneyyuri:
I'm not sure, but looks like making examples for manual commits will avoid common mistakes: https://github.com/confluentinc/confluent-kafka-go/issues/350, https://github.com/confluentinc/confluent-kafka-go/issues/286 and https://github.com/confluentinc/confluent-kafka-go/issues/195

mhowlett:
with librdkafka based clients, `auto.offset.commit` / `enable.auto.offset.store` are `true` / `true` by default and as you note, this does not give at least once semantics because the commit happens in a background thread independently of the application polling the consumer for new messages. By comparison, with the java client, `auto.offset.commit` does give at least once semantics, because the commit happens as a side effect of the next consumer.poll. 

The best way to achieve at least once semantics is to disable `enable.auto.offset.store`, which marks a message as eligible for commit as soon as it's delivered to the application and use StoreOffsets to manually indicate this instead. leave `auto.offset.commit` as true.

https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/consumer.go#L206

mhowlett:
yes, i mean `enable.auto.commit`.

marizaj:
[...]

Issue #350, opened 2019-07-05
It should be documented that offset means next offset instead of last one


leafduo:
Description
===========
APIs working with offsets, like `CommitOffsets `, `StoreOffsets`, does not specify what offset means. I recently found out offset means next offset to consume, not last offset just processed.

I think it's a common trap that many users fall into, some examples are:

- https://github.com/confluentinc/confluent-kafka-go/issues/286
- https://github.com/confluentinc/confluent-kafka-go/issues/195
- https://stackoverflow.com/questions/32902981/what-is-the-normal-kafka-offset-lag-why-minimum-is-1-not-0 (from a python client)

I've found no documentation in `confluent-kafka-go` that clarify this. `librdkafka` stated that in the [wiki](https://github.com/edenhill/librdkafka/wiki/Consumer-offset-management#terminology), which I think is not easy to find:

> Commit - Offset committed to permanent storage (broker, file). When consumer restarts this is where it will start consuming from. The committed offset should be last_message_offset+1.

Java client has a better [API documentation](https://kafka.apache.org/22/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#commitAsync-java.util.Map-org.apache.kafka.clients.consumer.OffsetCommitCallback-):

> Commit the specified offsets for the specified list of topics and partitions to Kafka.
> This commits offsets to Kafka. The offsets committed using this API will be used on the first fetch after every rebalance and also on startup. As such, if you need to store offsets in anything other than Kafka, this API should not be used. The committed offset should be the next message your application will consume, i.e. lastProcessedMessageOffset + 1.

It should be better that we include this clarification in godoc of related APIs.

jhaynie:
this caused me to spend several hours trying to debug why they weren't committing

edenhill:
Update docs

wedneyyuri:
Hi @leafduo, can you share any code describing how to solve this issue? As a user with little experience using kafka I miss solid examples of how to manual commit offset in the right way.

wedneyyuri:
What do you guys think about adding a new method `StoreMessageOffset` using the same logic as `CommitMessage` to avoid this "trap"? 

Of course it does not mean that the traditional methods `CommitOffsets` and` StoreOffsets` do not deserve better documentation but I believe it will help who only need to commit the current message offset and do not need to go into the details of increasing the offset position.

```go
// StoreMessageOffset stores offset based on the provided message.
// The provided message offset will be committed to the offset
[...]

Issue #286, opened 2019-02-13
consumer: Manual Commits give off-by-one error.


LiveOVErdrive:
Description
===========
When committing messages manually using *kafka.Consumer.CommitOffsets()*, the stored offset for the consumergroup will be 1 less than it should be.

For example, suppose a consumer begins at the beginning of the topic and consumes a single *Message* (offset 0).  The application passes the *TopicPartition* from this *Message* to *CommitOffsets()*.  This advances the consumergroup's *stored offset* from -1001 to 0.

Now if the consumer disconnects and reconnects, it will begin consuming from its stored offset.  This means it will consume message 0 again (despite the fact that this message's offset has already been committed).

How to reproduce
================
1. Create a kafka.Consumer
1. Create a kafka topic and publish at least one message into it.
1. Connect the consumer to the topic.
1. Consume one message with the consumer.
1. Call CommitOffsets() with the message's TopicPartition.
1. Disconnect the consumer
1. Connect another consumer with the same consumergroup.
1. It will re-consumer the message that was just committed.

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-go and librdkafka version (`LibraryVersion()`): v0.11.4
 - [x] Apache Kafka broker version:  kafka_2.12-0.10.2.1
 - [x] Client configuration:
```json
{
    "bootstrap.servers": "localhost:9092",
    "heartbeat.interval.ms": "2000",
    "enable.auto.commit": "false",
    "socket.receive.buffer.bytes": "3145728",
    "auto.offset.reset": "earliest",
    "session.timeout.ms": "30000",
    "request.timeout.ms": "90000",
    "max.partition.fetch.bytes": "524288",
    "fetch.wait.max.ms": "20000",
    "fetch.min.bytes": "262144",
    "statistics.interval.ms": "30000",
    "group.id": "test",
    "go.events.channel.size": 100
  }
```
 - [x] Operating system:  Alpine Linux, MacOS 10.13.6
 - [ ] Provide client logs (with `"debug": ".."` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

LiveOVErdrive:
One potential work-around is to increment the "Offset" field in the TopicPartition before committing.  This seems like it will work, but it feels a little unclean.

LiveOVErdrive:
I've just seen this and noted that I am supposed to commit offset+1.
https://github.com/confluentinc/confluent-kafka-go/issues/195

Closing this.

PR #482, opened 2020-06-23
add at least once processing example

wedneyyuri:
This is an example of how to achieve "at least once" processing guarantee using manual commits.

Although "exactly-once" processing is hard to achieve, I think it's common to tolerate any "at-least-once" processing guarantee. In fact, it's the guarantee offered by popular notifications solutions like [SQS](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/standard-queues.html#standard-queues-at-least-once-delivery), [Firebase](https://cloud.google.com/storage/docs/pubsub-notifications#delivery_guarantees) and [Azure Storage Queue](https://docs.microsoft.com/pt-br/azure/service-bus-messaging/service-bus-azure-and-service-bus-queues-compared-contrasted)

It's not a trivial task to build this minimal example without reference, so adding this example to the repository will avoid common mistakes and I hope will avoid issues described  in #350, #286 and #195

This issue closes https://github.com/confluentinc/confluent-kafka-go/issues/481

### DIFF
--- a/examples/consumer_at_least_once_example/consumer.go
+++ b/examples/consumer_at_least_once_example/consumer.go
@@ -1,0 +1,151 @@
+// Example function-based high-level Apache Kafka consumer
+package main
+
+/**
+ * Copyright 2016 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// consumer_at_least_once_example implements a consumer using the non-channel Poll() API
+// to retrieve messages and events disabling the auto commit configuration.
+// The default Kafka’s consumer auto commit configuration can lead to potential data loss.
+// It has no idea what you do with the message, and it’s much more free about committing offsets.
+// As far as the consumer is concerned, as soon as a message is pulled in, it’s “processed.”
+// There aren’t any easy fixes here. Fundamentally, this is a problem of weak consistency guarantees.
+// This example ilustrates how to “roll your own” at-least-once strategy that would manually
+// commit offsets after the end of the processing pipeline.
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
+)
+
+func main() {
+	var commitIntervalMessages int
+	flag.IntVar(&commitIntervalMessages, "commit.interval.messages", 10000, "Number of messages to process before each offset commit")
+
+	var commitIntervalTimeout time.Duration
+	flag.DurationVar(&commitIntervalTimeout, "commit.interval.timeout", 1*time.Minute, "The time to wait before the consumer offsets are committed (written) to offset storage.")
+
+	if len(os.Args) < 4 {
+		fmt.Fprintf(os.Stderr, "Usage: %s <broker> <group> <topics..>\n",
+			os.Args[0])
+		flag.PrintDefaults()
+		os.Exit(1)
+	}
+
+	broker := os.Args[1]
+	group := os.Args[2]
+	topics := os.Args[3:]
+
+	sigchan := make(chan os.Signal, 1)
+	signal.Notify(sigchan, syscall.SIGINT, syscall.SIGTERM)
+
+	c, err := kafka.NewConsumer(&kafka.ConfigMap{
+		"bootstrap.servers": broker,
+		// Avoid connecting to IPv6 brokers:
+		// This is needed for the ErrAllBrokersDown show-case below
+		// when using localhost brokers on OSX, since the OSX resolver
+		// will return the IPv6 addresses first.
+		// You typically don't need to specify this configuration property.
+		"broker.address.family": "v4",
+		"group.id":              group,
+		"session.timeout.ms":    6000,
+		"auto.offset.reset":     "earliest",
+
+		// We will commit the offset manually.
+		"enable.auto.commit": false,
+		// Number of messages to be buffered.
+		"queued.min.messages": 10000,
+	})
+	if err != nil {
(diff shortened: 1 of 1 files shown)


wedneyyuri:
[clabot:check]

mhowlett:
the preferred way to achieve at least once semantics is with StoreOffsets to avoid blocking the poll loop as described here: https://github.com/confluentinc/confluent-kafka-go/issues/481#issuecomment-647820887

wedneyyuri:
Thanks for the reply @mhowlett, I apologize for looking a little stupid but I’m rereading your comment since yesterday without being able to understand which way I should take. My goal is to describe in the pull request what I understand as the right way. I believe that I understand now, so if you allow me I will update the pull request with the suggestions you gave me.

Thank you for your patience and guidance.

wedneyyuri:
Thank you for your suggestions @mhowlett, I noticed that in addition to better performance (it's no more blocking the poll loop) the code is also easier to understand.

wedneyyuri:
Hi guys, I hope everyone is going well. I don't know if I was clear enough, but I did all the adjustments you asked and now I'm waiting for approval.

holive:
up

examples/consumer_at_least_once_example/consumer.go
```diff
@@ -0,0 +1,141 @@
...
+					fmt.Printf("%% Headers: %v\n", e.Headers)
+				}
+
+				// We can do any long running async operation here, because the next
+				// offset will not be available for commit until we call `c.StoreOffsets(...)`
+				_ = e
+				// In this example we store the next offset that will be committed
+				// to the offset store according to `auto.commit.interval.ms` or manual
+				// offset-less Commit().
+				// In this example we choose to terminate
```
edenhill, line 115: StoreOffsets() will only return an error if you pass partitions that are unknown to the consumer, but this is never the case since the partition is supplied with the Message returned from the consumer, so I think this comment and error handling here lures people into thinking this is a case that is both common and needs handling, which is not the case.

To keep things simple I recommend removing the error handling here, and instead documenting why the Offset need to be +1 (Kafka commits are for the next message to be consumed rather than the last message processed).

wedneyyuri: Hi @edenhill , thanks for the comment!

I believe it's not a good practice to ever ignore errors. This topic is discussed [here](https://golangbot.com/error-handling/#donotignoreerrors) and [here](https://stackoverflow.com/a/31028337/3738720). Usually when there are errors ignored throughout the code there is a need to reflect on why the error was ignored and whether it would be a bug or an interpretation of the package's behavior.

Instead of removing the error handling, what do you think about removing my comments and instead adding your explanation of why this error never happen? This will bring security to developers, especially those who are not so familiar with the package.

wedneyyuri: For example, I have a little doubt about what happens if I call the `StoreOffset` method with an old message after a repartitioning event (if this consumer is now working with another partition).

shubhang93: What happens if we manually call the `CommitMessage` method, for example:
```golang
for {
			msg, err := c.ReadMessage(1000*time.Millisecond)
			if err != nil && err.(kafka.Error).Code() != kafka.ErrTimedOut {
				log.Printf("err on instance %s -> %v", instanceID, err)
			}
			if msg != nil {
				log.Printf("Message received by [CONSUMER: %s TOPIC: %s PARTITION: %d]\n", instanceID, *msg.TopicPartition.Topic, msg.TopicPartition.Partition)
				processingFn(msg)
				_, cmtErr := consumer.CommitMessage(msg)
				if cmtErr != nil {
					break
				}
			}
		}
```
This way I can only commit messages if my `processingFn` has finished processing the message successfully

examples/consumer_at_least_once_example/consumer.go
```diff
@@ -121,15 +106,20 @@ func main() {
...
-				// be committed in the next loop.
-				lastMessage = e
-				msgsSinceLastCommit++
+				// In this example we store the next offset that will be committed
+				// to the offset store according to `auto.commit.interval.ms` or manual
+				// offset-less Commit().
+				// In this example we choose to terminate
+				// the application if the offset can't be stored.
+				offsets := []kafka.TopicPartition{e.TopicPartition}
+				offsets[0].Offset++
```
shubhang93, line 118: Also why are we incrementing the offset, suppose the offset delivered was `15`, shouldn't we be committing offset `15` instead of `16`. Correct me if I am wrong

wedneyyuri: Hi there, this is a little confusing, but offset means next offset instead of last one. https://github.com/confluentinc/confluent-kafka-go/issues/350

examples/consumer_at_least_once_example/consumer.go
wedneyyuri: Hi @shubhang93, if you manually call the `CommitMessage` method the consumer will block the poll loop. For systems that receive a lot of throughput this is not desirable as the consumer could be processing other messages in parallel.

> the preferred way to achieve at least once semantics is with StoreOffsets to avoid blocking the poll loop as described here: [#481 (comment)](https://github.com/confluentinc/confluent-kafka-go/issues/481#issuecomment-647820887)

shubhang93: Oh, makes sense. Thanks for clearing that out!

wedneyyuri: > I recommend documenting why the Offset need to be +1 (Kafka commits are for the next message to be consumed rather than the last message processed).

@edenhill Can we create a new method `StoreMessageOffset` as described here? https://github.com/confluentinc/confluent-kafka-go/issues/350#issuecomment-648290300

diegobernardes: I agree with @wedneyyuri position about the error handling. If the function can return an error it should be handled. If this change was inside the package, I would say that is still nice to handle the error, but the maintainer and contributors have enough knowledge to skip the error verification or not.

But on public a public contract? I think is a really bad decision to avoid error handling.

edenhill: > I believe it's not a good practice to ever ignore errors. 

I agree, and stand corrected.

examples/consumer_at_least_once_example/consumer.go
alexec: ah - so I wasn't mad - you do commit the next offset

wedneyyuri: Yes @alexec. I understand that this solution is a bit confusing but I couldn't find a better solution.

edenhill: There's a new StoreMessage() method that you can use instead, which abstracts these details.

PR #484, opened 2020-06-23
docs: offset means next offset instead of last one

wedneyyuri:
Closes https://github.com/confluentinc/confluent-kafka-go/issues/350

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -192,6 +192,12 @@ func (c *Consumer) CommitMessage(m *Message) ([]TopicPartition, error) {
 // CommitOffsets commits the provided list of offsets
 // This is a blocking call.
 // Returns the committed offsets on success.
+//
+// The offsets committed using this API will be used on the first fetch after
+// every rebalance and also on startup. As such, if you need to store offsets in
+// anything other than Kafka, this API should not be used.
+// The committed offset should be the next message your application will consume,
+// i.e. lastProcessedMessageOffset + 1.
 func (c *Consumer) CommitOffsets(offsets []TopicPartition) ([]TopicPartition, error) {
 	return c.commit(offsets)
 }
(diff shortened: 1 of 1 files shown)


wedneyyuri:
Hi guys, can you ask Travis to retry the build? I doubt my commit broke the build.

kafka/consumer.go
```diff
@@ -203,6 +209,12 @@ func (c *Consumer) CommitOffsets(offsets []TopicPartition) ([]TopicPartition, er
 // Returns the stored offsets on success. If at least one offset couldn't be stored,
 // an error and a list of offsets is returned. Each offset can be checked for
 // specific errors via its `.Error` member.
+//
+// The offsets committed using this API will be used on the first fetch after
```
edenhill, line 213: This implies that this commits the offsets, but it merely stores them for the upcoming auto commit, so this should probably be rephrased.

wedneyyuri: Thanks for the suggestion, when I created this pull request I didn't fully understand the difference between the two methods. Thanks for clarifying.

What do you think?

> StoreOffsets for the upcoming auto commit. 
> It's possible to enable or disable the autocommit using the option `enable.auto.commit` and the commit interval can be adjusted using the option `auto.commit.interval.ms`.
> The offsets stored using this API will be used on the first fetch after  every rebalance and also on startup. 
> As such, if you need to store offsets in anything other than Kafka, this API should not be used.
> 
> The stored offset should be the next message your application will consume,
> i.e. lastProcessedMessageOffset + 1.

wedneyyuri:
CLA ✅

kafka/consumer.go
edenhill: I think we'll want to avoid duplicating what is already on OffsetCommit() but instead point to that function doc for more info on how commits work.

Maybe just keep this part:
> The stored offset should be the next message your application will consume,
i.e. lastProcessedMessageOffset + 1.

PR #515, opened 2020-08-26
Return all partition-specific errors as Message

jdemeyer:
The API documentation for `ReadMessage` says

> partition-specific errors are returned as (msg, err) where msg.TopicPartition provides partition-specific information (such as topic, partition and offset). 

but this isn't always the case in practice (or I'm misunderstanding what this statement really means).

This PR makes that statement actually true by using the topicpartition information from the librdkafka error event.


jdemeyer:
The Travis CI build is having issues with Go 1.9, something about `terminfo` (which is very unrelated to my patch).

edenhill:
Thank you for this PR, it makes sense but I'll need to think through the consequences.

Did you run the integration tests with this change in place?

jdemeyer:
> Did you run the integration tests with this change in place?

No, I assumed that Travis CI would do that. I haven't figured out yet myself how to run the tests. (I did try it within my application to check that it did what I wanted it to do).

jdemeyer:
> Thank you for this PR, it makes sense but I'll need to think through the consequences.

Sure, this is a non-trivial change. The main reason that I made the change the way I did was because of the sentence in the documentation that I quoted above.

Another solution would be to add the `TopicPartition` information to the `Error` object. But then it's strange that an `Error` would contain a `TopicPartition` while a `TopicPartition` already contains an `error`.

jdemeyer:
[clabot:check]

PR #518, opened 2020-08-28
Replace librdkafka event polling with IO event triggering

KJTsanaktsidis:
This PR has two main aims:

* To add a new method to the consumer, PollContext(), which instead
  of accepting a max duration to poll for, accepts a context object
  and waits until either the context is canceled, or an event arrives.
  Importantly, it cancels the underlying call to rd_kafka_poll()
  cleanly once the context is canceled. This is achieved by using
  librdkafka's IO event triggering.
* To replace all places in the consumer and producer implementations
  that perform a similar kind of polling with usage of IO event
  triggering.

This patch uses the rd_kafka_queue_io_event_enable mechanism in
librdkafka. This causes librdkafka to write a byte to a file descriptor
whenever the number of events available transitions from zero to
nonzero. This enables us to "bridge" the world of librdkafka's internal
event loop, and the event loop maintained by the Golang scheduler, so
that we can effectively use a multiplexing `select{}` to wait for
events.

This change is reasonably invasive and unfortunately required the
shuffling around of a _small_ amount of code. The switch statement in
eventPoll has been factored into a new method processRkevToGoEvent,
which either internally handles an event and returns nil, or returns it.
Also, the explicit forwarding of delivery reports to the main events
channel in the RD_KAFKA_EVENT_DR case has been removed, and instead the
producer sets the main channel as the DR channel on the messages
_private field.

This is very much a first stab at this problem, so more than happy to take on
feedback and improve the patch! The main motivation for doing this is so that
confluent-kafka-go consumers can integrate nicely with application lifecycles
that cancel contexts to shut the app down, and the thought of papering over
the polling just kind of offended my aesthetic sensibilities :)


KJTsanaktsidis:
Oh, forgot about the CLA. Our legal team seems to turn these around in a couple of weeks... should give us plenty of time to iterate on the PR 😂

edenhill:
Very interesting! 💯 
 Will look at this closer in a bit.

KJTsanaktsidis:
[clabot:check]

KJTsanaktsidis:
@edenhill - finally got the CLA signed, so interested in your opinions on this change :)

mhowlett:
looks really good @KJTsanaktsidis - sorry for the delay in getting to it.

kafka/handle.go
```diff
@@ -323,3 +336,72 @@ func (h *handle) setOAuthBearerTokenFailure(errstr string) error {
 	}
 	return newError(cErr)
 }
+
+// handleIOTrigger wraps up librdkafka's rd_kafka_queue_io_event_enable functionality to let Go be notified of
+// new events available on librdkafka queues. The idea is that we set up a pair of pipe FD's in startIOTrigger,
+// and ask librdkafka to write a byte to the pipe when we go from 0 -> 1 events in a queue (i.e. it's edge
```
mhowlett, line 342: FWIW... 

My instinct is to prefer "papering over" the consumer polling than resorting to creating pipes. Can you use `rd_kafka_queue_yield` to exit the consumer poll early on done? The C# client has a consumer poll method that works like your producer.FlushContext. It doesn't bother calling queue_yield, but that would be possible using `CancellationToken.Register`. No one has complained about this yet - can you think of a practical scenario where that's not acceptable? I get that it's not appealing..

Is there any other way of waking up the go scheduler from C - even if that requires modifications to librdkafka?.

For blocking APIs that return a result (e.g. offsets for times) we really need some queue based alternatives...

KJTsanaktsidis: Hum. I reached for the IO triggering since it seemed like the reason it was in librdkafka at all was to support nodejs, and if you squint maybe the golang scheduler kind of looks like the nodejs event loop.

I must have missed rd_kafka_queue_yield, I didn't realise this was there at all. I have to spend a few hours looking more closely at the code, but I think that would perhaps lead to a less invasive PR - I wouldn't need to go and touch all the event handling stuff inside `eventPoll` for example.

Let me find some time to play around with rd_kafka_queue_yield and get back to you :)

mhowlett: btw: we had a discussion about this and other contributions internally, hopefully we'll get a meeting organized to discuss.

KJTsanaktsidis: Yup just chatting with your people now, keen to do that. We're now carrying a number of patches in our fork, so it's definitely worth having a conversastion about which bits of functionality make sense to merge into confluent-kafka-go, what shape they have to be in, and some plan of attack for what order to do this in. 

I'll look into rd_kafka_queue_yield in the meanwhile but yeah let's have that broader conversation first before we start working on the detail here for sure.

KJTsanaktsidis:
@mhowlett @edenhill Ok, took a bit longer than I was hoping, but I think I've got a version of this PR up that uses `rd_kafka_queue_yield` as you suggested. I made it as a separate PR so we could easily compare the two approaches: https://github.com/confluentinc/confluent-kafka-go/pull/626

PR #520, opened 2020-09-01
Make dynamic build use system librdkafka headers

### LINKED ISSUE #514: Build with dynamic tag finds bundled librdkafka header instead of system one
Description
===========
I'm trying to build confluent-kafka-go against a patched librdkafka, where I've added a function to the public interface. I've installed my patched librdkafka with the makefile into `/usr/local`, and pkg-config correctly figures out where it is and prints relevant include flags.

However, when I try and build my patched copy confluent-kafka-go that calls one of the new functions I added to librdkafka, I get the following error from cgo:

```
WORK=/tmp/go-build059416622
can't load package: package ./confluent/kafka: cannot find package "." in:
	/root/go/src/github.com/zendesk/zendesk_kafka_go/v2/confluent/kafka/confluent/kafka
root@eb416f2a9872:~/go/src/github.com/zendesk/zendesk_kafka_go/v2/confluent/kafka# cd ../..
root@eb416f2a9872:~/go/src/github.com/zendesk/zendesk_kafka_go/v2# go build -x -tags=dynamic ./confluent/kafka
WORK=/tmp/go-build552844409
mkdir -p $WORK/b001/
cd /root/go/src/github.com/zendesk/zendesk_kafka_go/v2/confluent/kafka
pkg-config --cflags -- rdkafka
pkg-config --libs -- rdkafka
CGO_LDFLAGS='"-g" "-O2" "-L/usr/local/lib" "-lrdkafka"' /usr/local/go/pkg/tool/linux_amd64/cgo -objdir $WORK/b001/ -importpath github.com/zendesk/zendesk_kafka_go/v2/confluent/kafka -- -I/usr/local/include -I $WORK/b001/ -g -O2 ./00version.go ./adminapi.go ./adminoptions.go ./build_dynamic.go ./config.go ./consumer.go ./error.go ./error_gen.go ./event.go ./generated_errors.go ./handle.go ./header.go ./kafka.go ./log.go ./message.go ./metadata.go ./misc.go ./offset.go ./producer.go ./testhelpers.go ./time.go
# github.com/zendesk/zendesk_kafka_go/v2/confluent/kafka
confluent/kafka/consumer.go:883:10: could not determine kind of name for C.rd_kafka_ssl_refresh
```

This is happening because cgo is using `librdkafka/rdkafka.h` from the confluent-kafka-go directory to figure out the types, _not_ the system copy in /usr/local/include/librdkafka/rdkafka.h (where this new function `rd_kafka_ssl_refresh` is defined). If I rename `rdkafka.h` in the source tree to `not_rdkafka.h`, everything works great.

I'm _pretty_ sure this is a cgo bug - it's not using different search paths for "" vs <> imports - although it could probably be worked around in confluent-kafka-go by nesting the `librdkafka` directory inside something else. I just wanted to check here first to make sure I'm not missing something obvious before I try and file a golang bug report.

How to reproduce
================
* Install a system librdkafka
* Write some garbage garbage that won't compile into the bundled librdkafka/rdkafka.h
* Run `go build -x -tags=dynamic ./kafka`
* Observe that the build fails because the bundled header doesn't parse. It shouldn't even be being used in the first place.

Checklist
=========

(Most of this isn't relevant):

 - [ ] confluent-kafka-go and librdkafka version (`LibraryVersion()`):
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `ConfigMap{...}`
 - [ ] Operating system: Ubuntu 18.04
 - [ ] Critical issue

KJTsanaktsidis:
cgo adds -I${SRCDIR} to the header search path by default; therefore,
using #include <librdkafka/rdkafka.h> can actually still find the
bundled librdkafka header in preference to the system one even when
`-tags=dynamic` is specified.

This change instead makes the static build set a preprocessor macro
USE_VENDORED_LIBRDKAFKA, which is used in a new header rdkafka_select.h
to either import the bundled one (with a special name, and quotes), or
the system one (with the <> import and the expected name).

Fixes #514


KJTsanaktsidis:
I couldn't actually _run_ the `./bundle-import.sh` script to check that my changes to it work, because I don't know where to find the `librdkafka-static-bundle-<VERSION>.tgz` artifact, and the nuget script in librdkafka can't build it without access to the `librdkafka-ci-packages` S3 bucket.

KJTsanaktsidis:
Think the go1.9 failures are the existing ones, but rewriting the imports in the examples seems to get the build to work elsewhere.

KJTsanaktsidis:
[clabot:check]

KJTsanaktsidis:
Well it only took nearly two months, but we (Zendesk) finally signed the corporate CLA 🎉 

The failing go 1.9 build looks to be what's fixed by https://github.com/confluentinc/confluent-kafka-go/pull/549

@edenhill - would ❤️ for you to take a look at this when you get a moment. Thanks!

examples/admin_create_topic/admin_create_topic.go
```diff
@@ -20,7 +20,7 @@ package main
 import (
 	"context"
 	"fmt"
-	"gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
```
edenhill, line 23: These import changes are not related to this fix, please revert.

kafka/00version.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 /*
-#include <librdkafka/rdkafka.h>
+#include "rdkafka_select.h"
```
edenhill, line 24: The "rdkafka_" prefix is used by librdkafka, so to be future-proof it is better if we use something else in the Go client, maybe "select_rdkafka.h".

kafka/generated_errors.go
```diff
@@ -1,9 +1,9 @@
 package kafka
 // Copyright 2016-2020 Confluent Inc.
-// AUTOMATICALLY GENERATED ON 2020-04-08 13:17:23.310983 +0200 CEST m=+0.001902312 USING librdkafka 1.4.0
+// AUTOMATICALLY GENERATED ON 2020-09-01 12:11:14.085648 +1000 AEST m=+0.000449739 USING librdkafka 1.4.2
```
edenhill, line 3: Rebase on latest master, we're now importiung v1.5.2

edenhill (changes requested):
Great stuff!
Some things to fix:
 - Rebase on latest master
 - Rename to select_rdkafka.h to avoid future collisions
 - Remove the unrelated import changes
 - Win!

KJTsanaktsidis:
@edenhill thanks for the review. Are you able to point me to where I can get a copy of librdkafka-static-bundle-v1.5.2.tgz from? As far as I can see, the travis builds are uploaded to a private S3 bucket and so I can't make this archive myself using the `./release.py --class StaticPackage` script in librdkafka.

It would be good to have a copy of that so I can confirm that my changes to the import script work as expected.

edenhill:
[librdkafka-static-bundle-v1.5.2.tgz.zip](https://github.com/confluentinc/confluent-kafka-go/files/5535476/librdkafka-static-bundle-v1.5.2.tgz.zip)
Unzip this file (github will not allow .tgz uploads..)

KJTsanaktsidis:
@edenhill Rebased and addressed most of your comments, PTAL! Two things outstanding:

Firstly - not sure how you want to handle git history around here - I was tempted to squash the re-import of the librdkafka bundle (which regenerated the build_*.go files and a couple other things) into "make dynamic build use system librdkafka headers". If you want me to squash it down, I can do so.

Secondly - unfortunately I think the rewrite of gopkg.in -> github.com paths is necessary. The problem is that I removed the package `github.com/confluentinc/confluent-kafka-go/kafka/librdkafka` by renaming the `librdkafka` directory to `librdkafka_vendor`. However, `gopkg.in/confluentinc/confluent-kafka-go.v1/kafka` still depends on `github.com/confluentinc/confluent-kafka-go/kafka/librdkafka` (at least until this change is merged). So, the examples won't build on this branch without having done this.

I think our options are:
* Keep the change to rewrite the paths away from gopkg.in
* Back out the rewrite and just merge it with the failing example build
* Merge https://github.com/confluentinc/confluent-kafka-go/pull/553 first which should make this a non-issue

edenhill:
Why is there an import of librdkafka v1.5.2 in this PR? We're already on v1.5.2.


We'll stick with gopkg.in for a bit longer and then eventually switch over to modules only, so for now it's okay with failed example builds.

KJTsanaktsidis:
OK - I've backed out the import path changes.

> Why is there an import of librdkafka v1.5.2 in this PR?

I ran the import script to regenerate all the `build_*.go` files (so we can be sure that a) the import script works and b) that there won't be a pointless diff next time the script is run). I rebased the import commits away though so you can't tell anymore.

kafka/librdkafka_vendor/import.sh
```diff
@@ -48,7 +48,7 @@ if [[ -z $version ]]; then
 fi
 
 # Verify branch state
-curr_branch=$(git symbolic-ref HEAD 2>/dev/null | cut -d"/" -f 3)
+curr_branch=$(git symbolic-ref HEAD 2>/dev/null | cut -d"/" -f 3-)
```
edenhill, line 51: Why this change? Can there be slashes in branches?

edenhill (approved):
Great stuff! Just one comment

kafka/kafka.go
```diff
@@ -254,14 +254,14 @@ package kafka
 import (
 	"fmt"
 	// Make sure librdkafka/ sub-directory is included in vendor pulls.
```
edenhill, line 256: librdkafka_vendor/

edenhill:
Are all comments addressed?

KJTsanaktsidis:
@edenhill Apologies - this totally fell of my radar! I've addressed the remaining feedback and I think this should be good to merge.

The PR is red because the examples won't build until it's merged as we discussed.

kafka/librdkafka_vendor/import.sh
KJTsanaktsidis: Yup, there definitely can - in fact this branch name has a slash in it too (`ktsanaktsidis/fix_dynamic_build`). At Zendesk it's conventional to structure branch names as `<your name>/<issue number>/<branch description>`, so I did this out of habit and found that the import script couldn't handle it.

edenhill:
Thanks alot for this, @KJTsanaktsidis !

edenhill:
Huhm, I wonder where this import is coming from:
https://travis-ci.org/github/confluentinc/confluent-kafka-go/jobs/753330232#L615

KJTsanaktsidis:
Hum.... not some caching or something from gopkg.in is it? There's no release process or anything to make it start returning the new package right?

KJTsanaktsidis:
and `PKG_CONFIG_PATH=/usr/local/Cellar/openssl@1.1/1.1.1i/lib/pkgconfig go get -tags=dynamic ./...` works on my machine too :/

edenhill:
Maybe it's time to get rid of the gopkg.in imports..

PR #546, opened 2020-10-24
Deprecate channel based consumer

### LINKED ISSUE #545: Channel based consumer not fully deprecated
Description
===========
Since the channel based consumer isnt fully deprecated, it leads to confusion while adopting this module. 
APIs, examples and documentation should be adapted

I will create a PR suggesting a fix soon

eran-levy:
Fixes #545 

There are still few tests such as the consumer_performance_test which are using the channel based consumer but I will create another issue to address that specifically.


eran-levy:
it looks like there are some issues with the tests that arent relevant to this PR specifically, created this issue to address that #548

eran-levy:
@edenhill would you like to schedule this full deprecation for 1.9 as well?

milindl:
Thanks for this PR! Closing this with #894 , we were able to deprecate channel based consumers.

PR #549, opened 2020-10-25
Fix failing Travis builds

### LINKED ISSUE #548: Fix failing Travis-CI builds
Description
===========
Travis-ci build failed for go 1.9 (both linux/osx) and go 1.14 OSX dynamic bundle

eran-levy:
Fixes #548


edenhill:
You seem to have checked in the vendor directory.

eran-levy:
@edenhill it seems like there is a version collision, the example that is using the gdamore/tcell module having version collision so I tried to use gopkg.in/gdamore/tcell.v1 but it still doesnt work and looking for gdamore/tcell/v2 which is already in a newer version.  Since in go 1.9 there is no way to get a specific version, the way to eliminate such errors is to use a vendor folder. 
We have few options here:
1.  use the vendor folder as suggested in this PR but since its only for the example to work, I'm not sure it worths it
2. init the examples as a go module but I've seen that go.mod has been removed so it might be for a reason
3. why do we still maintain go 1.9?

WDYT? any other suggestion here?

edenhill:
Thanks for the analysis, @eran-levy .

I'd say we remove the 1.9 builder, it is EOL anyway.
Care to update the PR to remove the 1.9 builders?

eran-levy:
sure, @edenhill updated the PR, tests pass see last commit, lmk if anything else is needed

.travis.yml
```diff
@@ -57,6 +51,7 @@ install:
   - for dir in kafka examples ; do (cd $dir && go install ${BUILD_TYPE} ./...) ; done
 
 script:
-  - if [[ -f .do_lint ]]; then golint -set_exit_status ./... ; fi
+  # should be replaced with golangci-lint
```
edenhill, line 54: why?

edenhill (approved):
LGTM

.travis.yml
eran-levy: Its much more configurable, sometimes golint not suitable and you want to run other linters to other folders such as:  examples.

PR #552, opened 2020-10-28
librdkafka static bundle v1.5.2

edenhill:


kafka/build_darwin.go
```diff
@@ -6,8 +6,8 @@
 package kafka
 
 // #cgo CFLAGS: -I${SRCDIR}
-// #cgo LDFLAGS: ${SRCDIR}/librdkafka/librdkafka_darwin.a  -lm -lsasl2 -lz -ldl -lpthread
+// #cgo LDFLAGS: ${SRCDIR}/librdkafka/librdkafka_darwin.a  -lm -lsasl2 -ldl -lpthread
```
mhowlett, line 9: i think this is statically linked now in the bundle, right?

kafka/librdkafka/rdkafka.h
```diff
@@ -55,13 +55,16 @@ extern "C" {
 #endif
 #endif
 
-#ifdef _MSC_VER
+#ifdef _WIN32
```
mhowlett, line 58: i assume this file is copied direct from librdkafka.

mhowlett (approved):

kafka/librdkafka/rdkafka.h
edenhill: Yep

kafka/build_darwin.go
edenhill: Huhm, no, this is due to changes with the .pc file.
Good catch!

edenhill:
I'll regenerate this with zlib fixed.

PR #554, opened 2020-10-29
librdkafka import improvements

edenhill:


edenhill:
Required to get v1.5.2 in

.travis.yml
```diff
@@ -55,3 +65,5 @@ script:
   - if [[ -f .do_lint ]]; then golint -set_exit_status ./examples/... ./kafka/... ./kafkatest/... ; fi
   - for dir in kafka ; do (cd $dir && go test -timeout 60s -v ${BUILD_TYPE} ./...) ; done
   - go-kafkacat --help
+  - library-version
+  - (library-version | grep "$EXPECT_LINK_INFO") || (echo "Incorrect linkage, expected $EXPECT_LINK_INFO" ; false)
```
KJTsanaktsidis, line 69: This test is (sort of) incorrect - because of the way cgo resolves imports, the app is _compiled_ against the bundled librdkafka headers, even if it is _linked_ against the dynamic system librdkafka. This of course works most of the time as long as the library stays ABI compatible, but it's definitely not ideal. We discussed this in this issue: https://github.com/confluentinc/confluent-kafka-go/issues/514
 
I proposed a fix for this in https://github.com/confluentinc/confluent-kafka-go/pull/520 that could perhaps be adapted or merged into this PR?

Then, we could add a constant or something to each branch of "rdkafka_select.h" and emit not just the linkage, but what header it was compiled with too, for this check.

edenhill: This just shows the linkage, either dynamic or static, which is irregardless of where the header file is found, so I believe this test to be correct.

mhowlett (approved):

PR #509, opened 2020-08-17
Import Windows Build of v1.5.0

neptoess:
Added scripts (`bundle-windows.cmd`, `bundle-windows.sh`) and artifacts for Windows static library. Resolves #501

PR #555, opened 2020-10-29
Import Windows Static Library for v1.5.2

neptoess:


neptoess:
@mhowlett @edenhill 
Let me know if this looks good or if you'd like anything from me. This is the same script from #509 , just run against v1.5.2 instead of v1.5.0

edenhill:
Can we add a Travis CI Windows worker to verify that the build works?

kafka/librdkafka/bundle-windows.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+
+# This should be run from the MSYS2 MinGW 64-bit shell to build librdkafka_windows.a
```
edenhill, line 3: Our Go client packaging workflow relies on all library assets already being built by librdkafka builders, so it would be best if we did the same thing here.
That means moving these build steps to a librdkafka Travis or AppVeyor worker, the artifacts will be automatically uploaded to the S3 bucket. Modify librdkafka/packaging/nuget/packaging.py to include the mingw build in its StaticPackage class, the resulting tar ball is what is used by the Go import.sh.

edenhill (changes requested):

kafka/librdkafka/bundle-windows.sh
neptoess: Working on this here https://github.com/edenhill/librdkafka/pull/3130

neptoess: @edenhill 
I believe that, if the changes in https://github.com/edenhill/librdkafka/pull/3130 are merged, only `bundle-import.sh` will need updated for the existing packaging workflow for confluent-kafka-go to support the Windows MinGW build. That said, I don't think _I'll_ be able to run `release.py`, as it requires AWS creds. How should I acquire `build_windows.go` and `librdkafka_windows.a` to add to this PR?

neptoess:
@edenhill 
> Can we add a Travis CI Windows worker to verify that the build works?
Definitely. Should I do that in this PR or open a separate one?

PR #564, opened 2020-11-05
Add Windows MinGW Static Library

### LINKED ISSUE #128: How to use Golang client on Windows
Description
===========
How can I configure windows in order for my Golang client to work ? e.g. where should I place `librdkafka` client ?

How to reproduce
================

1. download librdkafka from https://www.nuget.org/packages/librdkafka.redist/ & unzip its contents
2. `go get -u github.com/confluentinc/confluent-kafka-go/kafka`
3. setup `pkg-config` for windows
4. `go run main.go` results in

```
Package rdkafka was not found in the pkg-config search path.
Perhaps you should add the directory containing `rdkafka.pc'
to the PKG_CONFIG_PATH environment variable
No package 'rdkafka' found
pkg-config: exit status 1
```

neptoess:
Re-opening https://github.com/confluentinc/confluent-kafka-go/pull/555
Resolves #128


neptoess:
@edenhill ,
I will add the Travis CI worker in this PR

neptoess:
@edenhill 
Seeing this in the Travis log
```
=== RUN   TestProducerOAuthBearerConfig
    TestProducerOAuthBearerConfig: producer_test.go:283: NewProducer failed: No provider for SASL mechanism OAUTHBEARER: recompile librdkafka with openssl support. Current build options: PLAIN WindowsSSPI(GSSAPI)
--- FAIL: TestProducerOAuthBearerConfig (0.00s)
```
Since it seems intentional that PLAIN and GSSAPI are the only SASL options on Windows, should this test be skipped on Windows builds?

neptoess:
`build_windows.go` and `librdkafka_windows.a` came from a local librdkafka v1.5.2 build, but once https://github.com/edenhill/librdkafka/pull/3130 is merged, Travis will automatically build the static Windows bundle.

neptoess:
Resolves #128

neptoess:
@edenhill 
Now that Travis is building static librdkafka bundles for Windows, what else do we need to get this PR merged?

.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: go
 osx_image: xcode9.2
+go_import_path: github.com/confluentinc/confluent-kafka-go
```
edenhill, line 3: Why is this needed?

.travis.yml
```diff
@@ -55,6 +66,26 @@ before_install:
   - rm -rf tmp-build
   - if [[ -n $BUILD_TYPE ]]; then bash mk/bootstrap-librdkafka.sh ${LIBRDKAFKA_VERSION} tmp-build ; fi
   - go get -u golang.org/x/lint/golint && touch .do_lint
+  - mkdir -p "$GOPATH/src/gopkg.in/confluentinc/confluent-kafka-go.v1/"
+  - cp -r . "$GOPATH/src/gopkg.in/confluentinc/confluent-kafka-go.v1/"
```
edenhill, line 70: We're trying to move away from gopkg.in (in favour of go modules), do we really need this here?

.travis.yml
```diff
@@ -55,6 +66,26 @@ before_install:
...
+  - |-
+      case $TRAVIS_OS_NAME in
```
edenhill, line 72: Put the stuff below in mk/Makefile and call it from here, or a source script in mk/

kafka/librdkafka_vendor/bundle-import.sh
```diff
@@ -32,6 +32,11 @@ parse_dynlibs() {
             libs="${libs} $n"
         fi
     done
+    for n in $(grep ^Libs.private: $pc); do
```
edenhill, line 35: Why is Libs.private needed?
rdkafka-static.pc defines Libs:, not Libs-private (there are issues with that which I can't remember now)

edenhill (changes requested):
Don't do the actual import here, we'll do that for the upcoming 1.7.0 release, just update the tooling as needed. Thanks

.travis.yml
neptoess: The code in `examples/` uses the github.com import path. We're using https://docs.travis-ci.com/user/languages/go/#go-import-path to make sure that the Go compiler uses the code from this commit, instead of going to github and grabbing the latest master because the path doesn't exist.

.travis.yml
neptoess: It looks like all gopkg.in references for confluent-kafka-go have been removed from the code. I will remove this

kafka/librdkafka_vendor/bundle-import.sh
neptoess: This is where the `-lws2_32 -lsecur32 -lcrypt32` flags get defined for the MinGW build. Based on https://people.freedesktop.org/~dbn/pkg-config-guide.html, this appears to be correct behavior. Will this cause issues for one of the other static bundles?

neptoess:
> Don't do the actual import here, we'll do that for the upcoming 1.7.0 release, just update the tooling as needed. Thanks

This is done, but note that this will make the Travis Windows jobs fail (since build_windows.go and librdkafka_windows.a are missing)

kafka/librdkafka_vendor/bundle-import.sh
edenhill: I have a weak recollection that `pkg-config --libs --static ..` misbehaves when it comes to Libs.private, something about that it may pick up a dynamic library instead of a static, depending on path circumstances, or something with secondary dependencies. I can't really remember the exact details, but I do remember there being a strong reason for splitting out rdkafka .pc files into a static and non-static variant, and I'm not comfortable with ignoring that here, even though I don't have the full answer to why.

neptoess: I can appreciate that. I see two approaches:
    1. Modify the librdkafka CMake files to only use Libs for the static builds
    2. Modify bundle_import.sh to manually pop `-lws2_32 -lsecur32 -lcrypt32` for `build_windows.go`
Option #1 seems far more attractive to me. Agree?

neptoess: Also, I found a description of the misbehavior you were talking about in one of the links I sent
https://autotools.io/pkgconfig/dependencies.html
(Emphasis mine)
>2.1. Static linking
*When using the private definitions for dependencies, the behaviour of pkg-config will change depending on whether it's targeting dynamic or static linking*. In the former case, private dependencies will not be listed in the output, as they are not exposed to the consumer directly, and shared libraries will handle them.
>
>Contrarily, *when the link is static, it'll expand (recursively) all the private dependencies* so that all the libraries are brought in. This is once again needed because of the way UNIX link editors work, as archives do not list their own dependencies directly. While libtool actually provides a wrapper around these archives, they are getting less and less common, so *relying on them is not a good idea*.
>
>Unfortunately, at least as of February 2013, there is no easy way to tell at once to an Autotools-based build system that you intend to link statically; this means that the most common way to signal this to pkg-config becomes the command ./configure LDFLAGS="-static" PKG_CONFIG="pkg-config --static".

neptoess: PR opened for option 1
https://github.com/edenhill/librdkafka/pull/3354

neptoess:
@edenhill 
I think this should be ready to merge. Need anything else from me?

edenhill (approved):

edenhill:
Thanks alot for this!

edenhill:
@neptoess Hey, I just pushed an RC branch with your new Windows builds, can you try it out on Windows? (this RC branch will not be merged, to keep the repo size down).
https://github.com/confluentinc/confluent-kafka-go/tree/import_v1.7.0-RC6

edenhill:
Superb! Thank you for all your hard work on getting this done!

leonheess:
This PR sounds like Windows is now supported but the README still says _"confluent-kafka-go is not supported on Windows."_ - can you clarify if and how I can get it to run on Windows?

neptoess:
> This PR sounds like Windows is now supported but the README still says _"confluent-kafka-go is not supported on Windows."_ - can you clarify if and how I can get it to run on Windows?

It’s not officially supported, but, as of v1.7.0, a static bundle is being shipped. If you have a working cgo toolchain (the bundle is built with MSYS2 MinGW 64-bit, and is what I’d recommend), it _should_ “just work”

leonheess:
> It’s not officially supported, but, as of v1.7.0, a static bundle is being shipped. If you have a working cgo toolchain (the bundle is built with MSYS2 MinGW 64-bit, and is what I’d recommend), it _should_ “just work”

Thanks for the quick reply. I'm getting this error:
image
Can you point me in any direction?

neptoess:
> > It’s not officially supported, but, as of v1.7.0, a static bundle is being shipped. If you have a working cgo toolchain (the bundle is built with MSYS2 MinGW 64-bit, and is what I’d recommend), it _should_ “just work”
> 
> Thanks for the quick reply. I'm getting this error:
> image
> Can you point me in any direction?

Very bizarre. The Travis build and tests worked perfectly fine with this bundle https://travis-ci.org/github/confluentinc/confluent-kafka-go/jobs/770559600

__imp_wassert is defined in the C runtime library. What version of Windows are you running on? What gcc version?

leonheess:
> Very bizarre. The Travis build and tests worked perfectly fine with this bundle https://travis-ci.org/github/confluentinc/confluent-kafka-go/jobs/770559600

I saw that that is why I'm kind of confused

> __imp_wassert is defined in the C runtime library. What version of Windows are you running on? What gcc version?
image
image

neptoess:
> > Very bizarre. The Travis build and tests worked perfectly fine with this bundle https://travis-ci.org/github/confluentinc/confluent-kafka-go/jobs/770559600
> 
> I saw that that is why I'm kind of confused
> 
> > __imp_wassert is defined in the C runtime library. What version of Windows are you running on? What gcc version?
> 
> image
> image

Hmm. Gcc v8.1.0 is a little outdated. Per https://travis-ci.org/github/edenhill/librdkafka/jobs/770444016#L2714, the v1.7.0 bundle was built with v10.3. Maybe try upgrading?

leonheess:
> Hmm. Gcc v8.1.0 is a little outdated. Per https://travis-ci.org/github/edenhill/librdkafka/jobs/770444016#L2714, the v1.7.0 bundle was built with v10.3. Maybe try upgrading?

I upgraded to 11.10 and it's working! ty so much ❤

dassio:
for anyone that trying to build without too much background in C++, here is what i did :
install mingw-w64:  `pacman -S --needed  mingw-w64-x86_64-toolchain`
compile with command: `env GOOS=windows GOARCH=amd64 CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++ CGO_CXXFLAGS="-IC:\msys64\mingw64\include" go build`

neptoess:
> for anyone that trying to build without too much background in C++, here is what i did :
> install mingw-w64: `pacman -S --needed mingw-w64-x86_64-toolchain`
> compile with command: `env GOOS=windows GOARCH=amd64 CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++ CGO_CXXFLAGS="-IC:\msys64\mingw64\include" go build`

If you have C:\msys64\mingw64\bin in your PATH, you shouldn’t need to define all these environment variables. I just do plain old go build

neptoess:
Not sure about Scoop. The MinGW install I was using was via MSYS2

behoyh:
Thanks for the response, I believe the versions should be the same according to gcc --version:
```
gcc.exe (x86_64-posix-seh-rev2, Built by MinGW-W64 project) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

behoyh:
So as you say, what got this to work was installing MSYS2 version of gcc, via pacman:
##### update pacman
`pacman -Syu`
#### uninstall 32 target gcc if installed
`pacman -R gcc`
### install gcc on x64 targeting x64
`pacman -S mingw-w64-x86_64-gcc`

PR #566, opened 2020-11-10
Incremental rebalancing

edenhill:
This can't be merged until there's a v1.6.0 librdkafka release (to be bundled), but I think it is good to have the review out of the way as soon as possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Confluent's Golang client for Apache Kafka
 
+## v1.6.0
+
+v1.6.0 is a feature release adding support for KIP-429 Incremental rebalancing.
+
+ - Bundles librdkafka v1.6.0 - see release notes for all enhancements and fixes.
+
+
+confluent-kafka-go is based on librdkafka v1.6.0, see the
+[librdkafka release notes](https://github.com/edenhill/librdkafka/releases/tag/v1.6.0)
+for a complete list of changes, enhancements, fixes and upgrade considerations.
+
+
+### Fixes
+
+ * Consumer.Close() previously did not trigger the final RevokePartitions
+   callback, this is now fixed.
+
+
+
 ## v1.5.2
 
 v1.5.2 is a maintenance release with the following fixes and enhancements:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,11 @@
 
 ## v1.6.0
 
-v1.6.0 is a feature release adding support for KIP-429 Incremental rebalancing.
+v1.6.0 is a feature release adding support for KIP-429 Incremental rebalancing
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Confluent's Golang client for Apache Kafka
 
+## v1.6.0
+
+v1.6.0 is a feature release adding support for KIP-429 Incremental rebalancing.
```
mhowlett, line 5: also sticky partitioner.

mhowlett: very sticky release

kafka/consumer.go
```diff
@@ -122,6 +121,63 @@ func (c *Consumer) Unassign() (err error) {
 	return nil
 }
 
+// IncrementalAssign adds the specified partitions to the current set of
+// partitions to consume.
+// The new partitions must not be part of the current assignment.
+func (c *Consumer) IncrementalAssign(partitions []TopicPartition) (err error) {
```
mhowlett, line 127: I assume you can specify offsets here? there's no comment about this and it doesn't follow from the typing.

kafka/consumer_test.go
```diff
@@ -115,6 +115,38 @@ func TestConsumerAPIs(t *testing.T) {
 		t.Errorf("Unassign failed: %s", err)
 	}
 
+	// Incremental Assign & Unassign
+	err = c.IncrementalAssign([]TopicPartition{
+		{Topic: &topic1, Partition: 9},
```
mhowlett, line 120: offsets?

kafka/integration_test.go
```diff
@@ -337,12 +338,16 @@ func consumerTest(t *testing.T, testname string, msgcnt int, cc consumerCtrl, co
...
 	conf := ConfigMap{"bootstrap.servers": testconf.Brokers,
 		"go.events.channel.enable": cc.useChannel,
-		"group.id":                 testconf.GroupID,
-		"session.timeout.ms":       6000,
-		"api.version.request":      "true",
-		"enable.auto.commit":       cc.autoCommit,
-		"debug":                    ",",
-		"auto.offset.reset":        "earliest"}
+		"group.id": testconf.GroupID +
+			fmt.Sprintf("-%d", rand.Intn(1000000)),
```
mhowlett, line 342: nit: a 1/Uuid chance of dying would be better

mhowlett (approved):

kafka/integration_test.go
edenhill: There's no uuid generator in the standard library, don't want to pull in an external dependency just for this.

huanggx-sea:
Hi, any update for this? Seems like the librdkafka has already released v1.6.0.

PR #569, opened 2020-11-13
docs: update producer channel example with error handling

mavogel:
see https://github.com/edenhill/librdkafka/issues/3105 for details

### DIFF
--- a/examples/producer_channel_example/producer_channel_example.go
+++ b/examples/producer_channel_example/producer_channel_example.go
@@ -19,8 +19,9 @@ package main
 
 import (
 	"fmt"
-	"gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
 	"os"
+
+	"gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


mavogel:
[clabot:check]

examples/producer_channel_example/producer_channel_example.go
```diff
@@ -58,7 +59,15 @@ func main() {
 						*m.TopicPartition.Topic, m.TopicPartition.Partition, m.TopicPartition.Offset)
 				}
 				return
+			case *kafka.Error:
+				err := ev
+				if err.IsRetriable() {
```
edenhill, line 64: Delivery errors are passed as `*kafka.Message`, see the error checking on line 54.

`*kafka.Error` is for generic client-level errors, and there is no notion of retriable there (since it is not directly tied to an API call).

mavogel: Ok, thanks for the explanation. When and how should `err.IsRetriable()` be used? Can it be ignored or not be called at all because it's not tied to an API call? Then why not remove it? I am confused.

edenhill: IsRetriable() is currently only returned by the transactional producer APIs, but the plan is to have all APIs return this eventually, where it makes sense.

PR #582, opened 2021-01-07
Use github.com rather than gopkg.in for imports in examples

edenhill:
This is in line with go modules


edenhill:
Care to take a look at this, @KJTsanaktsidis ?

KJTsanaktsidis (approved):
Yeah i think this is a good idea - as you say, preparing the way for making confluent-kafka-go a go module.

rigelbm (approved):
LGTM

PR #611, opened 2021-03-18
Bump timeout on go test command

KJTsanaktsidis:
It seems the travis tests are failing because they're timing out. Bumping the timeout to see if that fixes it, or if the timeout is actually the result of some other brokenness.

EDIT: yeah that seems to be all it took. @mhowlett What do you think, should we merge this?


edenhill:
[clabot:check]

edenhill:
Thank you!

PR #617, opened 2021-03-25
Soak test for go client

jliunyu:
Running with 2 processes for soak test: 

Process-1 : Create Producer-1 to keep sending messages to the input-topic.
                    Create Consumer-2 to receive messages from the output-topic.

Process-2: Create Consumer-1 to receive messages from the input-topic.
                    Create transactional producer Producer-2 to sending the received messages to the output-topic.
                   The initTransaction API registers a transactional.id when creating the producer.
                   The producer sends the messages to the output-topic.
                   Commit transaction after sending 100 messages each time or 1 second.

There are 3 new files added:
1. soakclient_test.go: implementation of Process-1, Creating topics, and common functions used by both processes include: Producer delivery checking, Stats Events handling and received message handling
2. soakclient_transaction_test.go: implementation of Process-2
3. datadog.go: datadog functions and resource calculation

Command: 
1. Process-1: go test -run TestSoakGoClient
2. Process-2: go test -run TestTransaction

### DIFF
--- a/soaktest/README.md
+++ b/soaktest/README.md
@@ -1,0 +1,40 @@
+# Information for soak test for go client
+The soak test contains 2 applications:
+soakclient.go for one application : Create Producer-1 to keep sending messages to the input-topic.
+Create Consumer-2 to receive messages from the output-topic.
+
+soakclient_transaction.go for the other application: Create Consumer-1 to receive messages from the input-topic.
+Create transactional producer Producer-2 to sending the received messages to the output-topic.
+The initTransaction API registers a transactional.id when creating the producer.
+The producer sends the messages to the output-topic.
+Commit transaction after sending 100 messages each time or 1 second.
+
+# Instruction on configuration and running
+
+# Build soaktest API
+    $ cd soaktest
+    $ go mod init soaktest
+    $ go mod tidy
+    $ go install soaktest
+
+# Build and run soakclient.go
+    $ cd soaktest/soakclient
+    $ go build
+    $ ./soakclient -broker <broker> -inputTopic <..> -outTopic <..> -groupID <..> -inputTopicPartitionsNum <..> -outTopicPartitionsNum <..> -replicationFactor <..> -ccloudAPIKey <ccloudAPIKey> -ccloudAPISecret <ccloudAPISecret>
+
+# Build and run soakclient_transaction.go
+    $ cd soaktest/soakclient_transaction
+    $ go build
+    $ ./soakclient_transaction -broker <broker> -inputTopic <..> -outTopic <..> -outTopicPartitionsNum <..> -groupID <..> -transactionID <..> -ccloudAPIKey <ccloudAPIKey> -ccloudAPISecret <ccloudAPISecret>
+
+### The default values if doesn't input from the command lines
+    inputTopic: "inputTopic"
+    outTopic: "outTopic"
+    groupID : "groupID"
+    inputTopicPartitionsNum: 1
+    outTopicPartitionsNum: 1
+    replicationFactor: 1
+    transactionID: transactionID
+
+# Recommend instance type and size
+At lease 4 CPUs and 8 GB volume size
--- a/soaktest/soakclient/soakclient.go
+++ b/soaktest/soakclient/soakclient.go
@@ -1,0 +1,362 @@
+package main
+
+/**
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
+	"os"
+	"os/signal"
+	"soaktest"
+	"sync"
+	"syscall"
+	"time"
+)
+
+const endToEnd = "end.to.end."
+const rate = 2
+
+func main() {
+
+	broker := flag.String("broker", "", "bootstrap servers")
+	inputTopic := flag.String("inputTopic", "inputTopic",
+		"producer will deliver messages to this topic")
+	outTopic := flag.String("outTopic", "outTopic",
+		"consumer will consume messages from this topic")
+	groupID := flag.String("groupID", "groupID",
+		"the group consumer will join")
+	inputTopicPartitionsNum := flag.Int("inputTopicPartitionsNum", 1,
+		"inputTopic partition number")
+	outTopicPartitionsNum := flag.Int("outTopicPartitionsNum", 1,
+		"outTopic partition number")
+	replicationFactor := flag.Int("replicationFactor", 1, "topic replication")
+	ccloudAPIKey := flag.String("ccloudAPIKey", "", "sasl username")
+	ccloudAPISecret := flag.String("ccloudAPISecret", "", "sasl password")
+
+	flag.Parse()
+
+	soaktest.InitLogFiles(fmt.Sprintf("../log/soakclient_%s.log",
+		time.Now().Format("2006-01-02")))
+	soaktest.InfoLogger.Printf("Starting the application...\n")
+
+	num, version := kafka.LibraryVersion()
+	soaktest.InfoLogger.Printf("num = %d, librdkafka %q\n", num, version)
+
+	var wg sync.WaitGroup
+	doneChan := make(chan bool, 1)
+	errorChan := make(chan error, 1)
+
+	wg.Add(1)
+	go soaktest.GetRusageMetrics(endToEnd, &wg, doneChan, errorChan)
+
+	maxDuration, err := time.ParseDuration("30s")
+	if err != nil {
+		soaktest.ErrorLogger.Printf("Failed to create maxDuration with "+
+			"err %s\n", err)
+		os.Exit(1)
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), maxDuration)
+	defer cancel()
+
+	newTopics := []kafka.TopicSpecification{
+		{Topic: *inputTopic,
+			NumPartitions:     *inputTopicPartitionsNum,
+			ReplicationFactor: *replicationFactor},
+		{Topic: *outTopic,
+			NumPartitions:     *outTopicPartitionsNum,
+			ReplicationFactor: *replicationFactor}}
+
+	err = CreateTopic(ctx, newTopics, *broker, *ccloudAPIKey, *ccloudAPISecret)
+	if err != nil {
+		soaktest.ErrorLogger.Printf("Failed to create topics: %s\n", err)
+		close(doneChan)
+		os.Exit(1)
+	}
+
+	wg.Add(1)
+	go producer(*inputTopic, *broker, *ccloudAPIKey,
+		*ccloudAPISecret, &wg, doneChan, errorChan,
+		uint64(*inputTopicPartitionsNum))
+
+	wg.Add(1)
+	go consumer(*outTopic, *broker, *groupID, *ccloudAPIKey,
+		*ccloudAPISecret, &wg, doneChan, errorChan)
+
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	select {
+	case sig := <-sigChan:
+		soaktest.InfoLogger.Printf("Signal caught: %v\n", sig)
+	case err := <-errorChan:
+		soaktest.ErrorLogger.Printf("Error caught: %s\n", err)
+	}
+
+	close(doneChan)
+	close(errorChan)
+	for err = range errorChan {
+		soaktest.ErrorLogger.Printf("Error caught: %s\n", err)
+	}
+	wg.Wait()
+	soaktest.PrintConsumerStatus()
+	soaktest.PrintProducerStatus()
+}
+
+// CreateTopic creates the topics if it doesn't exist
+func CreateTopic(ctx context.Context,
+	topics []kafka.TopicSpecification,
+	broker, ccloudAPIKey, ccloudAPISecret string) error {
+	conf := kafka.ConfigMap{
+		"bootstrap.servers": broker,
+		"sasl.mechanisms":   "PLAIN",
+		"security.protocol": "SASL_SSL",
+		"sasl.username":     ccloudAPIKey,
+		"sasl.password":     ccloudAPISecret}
+	AdminClient, err := kafka.NewAdminClient(&conf)
+	if err != nil {
+		soaktest.ErrorLogger.Printf("Failed to create AdminClient: %s\n", err)
+		return err
+	}
+
+	result, err := AdminClient.CreateTopics(ctx, topics)
+	if err != nil {
+		soaktest.ErrorLogger.Printf("Failed to create new topics: %s\n", err)
+		return err
+	}
+
+	for _, res := range result {
+		switch res.Error.Code() {
+		case kafka.ErrTopicAlreadyExists:
+			soaktest.ErrorLogger.Printf("Failed to create topic %s: %s\n",
+				res.Topic, res.Error)
+
+		case kafka.ErrNoError:
+			soaktest.ErrorLogger.Printf("Succeed to create topic %s\n",
+				res.Topic)
+
+		default:
+			err = fmt.Errorf("failed to create topic %s: %s",
+				res.Topic, res.Error)
+			return err
+		}
+	}
+	return nil
+}
+
+// terminateProducer waits for all messages in the Producer queue to
+// be delivered, closes the producer and pass error to the error channel
+func terminateProducer(p *kafka.Producer, err error, errorChan chan error) {
+	remaining := p.Flush(30)
+	soaktest.InfoLogger.Printf("producer: %d message(s) remaining in queue "+
+		"after flush()\n", remaining)
+	p.Close()
+	errorChan <- err
+}
+
+// terminateConsumer closes consumer and pass error to the error channel
+func terminateConsumer(c *kafka.Consumer, err error, errorChan chan error) {
+	c.Close()
+	errorChan <- err
+}
+
+// producer produces messages to input topic
+func producer(inputTopic, broker, ccloudAPIKey, ccloudAPISecret string,
+	wg *sync.WaitGroup, termChan chan bool, errorChan chan error,
+	partitionNum uint64) {
+	defer wg.Done()
+	p, err := kafka.NewProducer(&kafka.ConfigMap{
+		"bootstrap.servers":      broker,
+		"statistics.interval.ms": 5000,
+		"sasl.mechanisms":        "PLAIN",
+		"security.protocol":      "SASL_SSL",
+		"sasl.username":          ccloudAPIKey,
+		"sasl.password":          ccloudAPISecret})
+	if err != nil {
+		soaktest.ErrorLogger.Printf("Failed to create producer %s\n", err)
+		errorChan <- err
+		return
+	}
+
+	run := true
+	doneChan := make(chan bool)
+	tags := []string{fmt.Sprintf("topic:%s", inputTopic)}
+	partitionToMsgIDMap := make(map[uint64]uint64)
+
+	go func() {
+		doTerm := false
+		for !doTerm {
+			select {
+			case e := <-p.Events():
+				switch ev := e.(type) {
+				case *kafka.Message:
+					// Message delivery report
+					soaktest.ProducerDeliveryCheck(ev)
+
+				case *kafka.Stats:
+					err = soaktest.HandleStatsEvent(ev, inputTopic)
+					if err != nil {
+						soaktest.ErrorLogger.Printf("Failed to "+
+							"HandleStatsEvent: %s\n", err)
+						errorChan <- err
+					}
+
+				case kafka.Error:
+					// kafka.Errors should generally be
+					// considered informational, the client
+					// will try to automatically recover.
+					soaktest.ProducerErrCnt++
+					soaktest.DatadogIncrement(soaktest.ProducerError, 1, tags)
+					soaktest.ErrorLogger.Printf("kafka.Error: %v: %v for "+
+						"producer %v\n", ev.Code(), ev, p)
+
+				default:
+					soaktest.WarningLogger.Printf("Ignored event: %v for"+
+						"producer %v\n", ev, p)
+				}
+			case <-doneChan:
+				doTerm = true
+			}
+		}
+		close(doneChan)
+	}()
+
+	value := "Hello Go!"
+
+	sleepIntvl := 1.0 / rate * 1000
+	ticker := time.NewTicker(time.Millisecond * time.Duration(sleepIntvl))
+
+	for run {
+		select {
+		case <-termChan:
+			doneChan <- true
+			run = false
+		case <-ticker.C:
+			encodingTime, err := time.Now().GobEncode()
+			if err != nil {
+				soaktest.ErrorLogger.Printf("Failed to get MarshalText "+
+					"%s\n", err)
+				doneChan <- true
+				terminateProducer(p, err, errorChan)
+				return
+			}
+			key := soaktest.ProduceMsgCnt % partitionNum
+			msgid := partitionToMsgIDMap[key] + 1
+			err = p.Produce(&kafka.Message{
+				TopicPartition: kafka.TopicPartition{
+					Topic:     &inputTopic,
+					Partition: kafka.PartitionAny},
+				Key:   soaktest.ConvertUint64ToByteArray(soaktest.ProduceMsgCnt % partitionNum),
+				Value: []byte(value),
+				Headers: []kafka.Header{{Key: "time",
+					Value: encodingTime},
+					{Key: "msgid",
+						Value: soaktest.ConvertUint64ToByteArray(msgid)}},
+			}, nil)
+			if err != nil {
+				soaktest.ErrorLogger.Printf("Failed to Produce message to "+
+					"inputTopic %s, %s\n", inputTopic, err)
+				soaktest.DatadogIncrement(soaktest.FailedToProduceMsg, 1, tags)
+			} else {
+				partitionToMsgIDMap[key] = msgid
+				soaktest.ProduceMsgCnt++
+			}
+		}
+	}
+	remaining := p.Flush(30)
+	p.Close()
+	soaktest.InfoLogger.Printf("producer: %d message(s) remaining in queue "+
+		"after flush()\n", remaining)
+}
+
+// consumer receives and verifies messages from output topic
+func consumer(topic, broker, groupID, ccloudAPIKey, ccloudAPISecret string,
+	wg *sync.WaitGroup, termChan chan bool, errorChan chan error) {
+	wg.Done()
+	hwmarks := make(map[string]uint64)
+	c, err := kafka.NewConsumer(&kafka.ConfigMap{
+		"bootstrap.servers":      broker,
+		"group.id":               "groupID",
+		"auto.offset.reset":      "earliest",
+		"enable.auto.commit":     false,
+		"statistics.interval.ms": 5000,
+		"sasl.mechanisms":        "PLAIN",
+		"security.protocol":      "SASL_SSL",
+		"sasl.username":          ccloudAPIKey,
+		"sasl.password":          ccloudAPISecret,
+	})
+	if err != nil {
+		soaktest.ErrorLogger.Printf("Failed to create consumer: %s\n", err)
+		errorChan <- err
+		return
+	}
+
+	err = c.Subscribe(topic, nil)
+
+	if err != nil {
+		soaktest.ErrorLogger.Printf("Failed to Subscribe to topic: %s with "+
+			"error %s\n", topic, err)
+		terminateConsumer(c, err, errorChan)
+		return
+	}
+
+	run := true
+	tags := []string{fmt.Sprintf("topic:%s", topic)}
+
+	for run {
+		select {
+		case <-termChan:
+			run = false
+		default:
+			ev := c.Poll(100)
+			if ev == nil {
+				continue
+			}
+
+			switch e := ev.(type) {
+			case *kafka.Message:
+				soaktest.HandleMessage(e, hwmarks)
+
+			case *kafka.Stats:
+				err = soaktest.HandleStatsEvent(e, topic)
+				if err != nil {
+					soaktest.ErrorLogger.Printf("Failed to "+
+						"HandleStatsEvent: %s\n", err)
+					terminateConsumer(c, err, errorChan)
+					return
+				}
+
+			case kafka.Error:
+				// kafka.Errors should generally be
+				// considered informational, the client
+				// will try to automatically recover.
+				soaktest.ConsumerErrCnt++
+				soaktest.DatadogIncrement(soaktest.ConsumerError, 1, tags)
+				soaktest.ErrorLogger.Printf("kafka.Error: %v: %v for "+
+					"consumer %v\n", e.Code(), e, c)
+
+			default:
+				soaktest.WarningLogger.Printf("Ignored %v for consumer %v\n",
+					e, c)
+			}
+
+		}
+	}
+	c.Close()
+	soaktest.ErrorLogger.Printf("Consumer closed\n")
+}
--- a/soaktest/soakclient/soakclient.go
+++ b/soaktest/soakclient/soakclient.go
@@ -20,17 +20,20 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"github.com/confluentinc/confluent-kafka-go/kafka"
 	"os"
 	"os/signal"
 	"soaktest"
 	"sync"
 	"syscall"
 	"time"
+
+	"github.com/confluentinc/confluent-kafka-go/kafka"
 )
 
 const endToEnd = "end.to.end."
 const rate = 2
+const producerType = "InputProducer"
+const consumerType = "OutputConsumer"
 
 func main() {
 
@@ -41,9 +44,9 @@ func main() {
 		"consumer will consume messages from this topic")
 	groupID := flag.String("groupID", "groupID",
 		"the group consumer will join")
-	inputTopicPartitionsNum := flag.Int("inputTopicPartitionsNum", 1,
+	inputTopicPartitionsNum := flag.Int("inputTopicPartitionsNum", 4,
 		"inputTopic partition number")
-	outTopicPartitionsNum := flag.Int("outTopicPartitionsNum", 1,
+	outTopicPartitionsNum := flag.Int("outTopicPartitionsNum", 4,
 		"outTopic partition number")
 	replicationFactor := flag.Int("replicationFactor", 1, "topic replication")
 	ccloudAPIKey := flag.String("ccloudAPIKey", "", "sasl username")
@@ -114,8 +117,8 @@ func main() {
 		soaktest.ErrorLogger.Printf("Error caught: %s\n", err)
 	}
 	wg.Wait()
-	soaktest.PrintConsumerStatus()
-	soaktest.PrintProducerStatus()
+	soaktest.PrintConsumerStatus(consumerType)
+	soaktest.PrintProducerStatus(producerType)
 }
 
 // CreateTopic creates the topics if it doesn't exist
@@ -261,7 +264,7 @@ func producer(inputTopic, broker, ccloudAPIKey, ccloudAPISecret string,
 				TopicPartition: kafka.TopicPartition{
 					Topic:     &inputTopic,
 					Partition: kafka.PartitionAny},
-				Key:   soaktest.ConvertUint64ToByteArray(soaktest.ProduceMsgCnt % partitionNum),
+				Key:   soaktest.ConvertUint64ToByteArray(key),
 				Value: []byte(value),
 				Headers: []kafka.Header{{Key: "time",
 					Value: encodingTime},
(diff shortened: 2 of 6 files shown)


tests/soak/soakgoclient/soakclient_transaction_test.go
```diff
@@ -0,0 +1,267 @@
...
+                transactionCommitNum += 1
+                err = p.CommitTransaction(nil)
+                if err != nil {
+                        DatadogIncrement(PRODUCER_TRANSACTION_COMMIT_FAILED, 1)
+                        t.Fatalf("CommitTransaction() failed: %s\n", err)
+                } else {
+                        DatadogIncrement(PRODUCER_TRANSACTION_COMMIT_SUCCEED, 1)
+                }
+       } else {
+               err = p.AbortTransaction(nil)
```
edenhill, line 257: The consumer needs to rewind its position to the pre-transaction offset, i.e., the committed offset.
I think you can do this by simply doing something like:
```go
   assignment := c.Assignment()
   for _, tp = range assignment {
       tp.Offset = kafka.OffsetStored  // the committed offset
       err := c.Seek(tp)
   }
```

tests/soak/soakgoclient/soakclient_transaction_test.go
```diff
@@ -0,0 +1,267 @@
...
+               if err != nil {
+                       DatadogIncrement(PRODUCER_TRANSACTION_ABORT_FAILED, 1)
+                       t.Fatalf("AbortTransaction() failed: %s\n", err)
+               } else {
+                       DatadogIncrement(PRODUCER_TRANSACTION_ABORT_SUCCEED, 1)
+               }
+               transactionCommitNum = 1
+       }
+       return transactionCommitNum
+}
```
edenhill, line 267: missing newline

edenhill:
Add `./soaktest/...`  to list of linted paths in .travis.yml

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,495 @@
...
+
+	var msgid uint64
+	run := true
+	commited := false
+	for run == true {
+		select {
+		case sig := <-sigchan:
+			fmt.Printf("Caught signal %v: terminating\n", sig)
+			run = false
+			if msgid != 0 && !commited {
```
edenhill, line 202: typo: committed

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,495 @@
...
+					*c,
+					transactionCommitNum)
+			}
+			soaktest.PrintConsumerStatus()
+			return err
+		case <-ticker.C:
+			if msgid != 0 && !commited {
+				transactionCommitNum, err = commitTransaction(
+					getAllPartitionPositions(partitionPositionMap),
+					*p,
```
edenhill, line 215: We use pointers for the client instances in all APIs so I think you can just pass them as pointers and not dereference them

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,495 @@
...
+				}
+
+			case *kafka.Stats:
+				soaktest.HandleStatsEvent(e, inputTopic)
+
+			case kafka.Error:
+				// kafka.Errors should generally be
+				// considered informational, the client
+				// will try to automatically recover.
+				soaktest.ConsumerMsgErrs += 1
```
edenhill, line 270: not a message error but a generic consumer error, use a separate counter

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,495 @@
...
+
+// Transaction producer, produce messages to output topic
+func produceMessage(e *kafka.Message,
+	p kafka.Producer,
+	msgid uint64,
+	outTopic string) {
+	encodingTime, err := time.Now().GobEncode()
+	if err != nil {
+		fmt.Errorf("Failed to encode current time %s\n", err)
+	}
```
edenhill, line 292: error not handled

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,495 @@
...
+		soaktest.ProducerDeliveryCheck(e)
+
+	case *kafka.Stats:
+		soaktest.HandleStatsEvent(e, outTopic)
+
+	case kafka.Error:
+		// kafka.Errors should generally be
+		// considered informational, the client
+		// will try to automatically recover.
+		soaktest.DrErrCnt += 1
```
edenhill, line 317: not a delivery error, these are generic client errors. use a separate counte

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,495 @@
...
+		TopicPartition: kafka.TopicPartition{
+			Topic:     &outTopic,
+			Partition: kafka.PartitionAny},
+		Value: []byte(string(e.Value)),
+		Headers: []kafka.Header{{Key: "time",
+			Value: encodingTime},
+			{Key: "msgid",
+				Value: soaktest.ConvertUint64ToByteArray(msgid)}},
+	}, nil)
+	// Wait for delivery report
```
edenhill, line 304: This makes producing synchronous, which is an anti-pattern and slow.
Use a separate go-routine or the same loop that calls produceMessage() to read the events channel asynchronously.

See https://github.com/edenhill/librdkafka/wiki/FAQ#why-is-there-no-sync-produce-interface

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,495 @@
...
+func commitTransaction(partitionPosition []kafka.TopicPartition,
+	p kafka.Producer,
+	c kafka.Consumer,
+	transactionCommitNum int) (commitNum int, err error) {
+	fmt.Printf("=== Committing transaction===\n")
+
+	cgmd, err := c.GetConsumerGroupMetadata()
+	if err != nil {
+		fmt.Printf("Failed to get Consumer Group Metadata %v\n", err)
+	}
```
edenhill, line 339: error is not handled

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,495 @@
...
+	// If SendOffsetsToTransaction failed with err.(kafka.Error).TxnRequiresAbort(),
+	// AbortTransaction and return (transactionCommitNum, err)
+	// If failed with other errors, return transactionCommitNum and the
+	// fatal error
+	for i := 0; i < RETRY_NUM; i++ {
+		err = p.SendOffsetsToTransaction(nil, partitionPosition, cgmd)
+		if err != nil {
+			fmt.Printf("SendOffsetsToTransaction() failed: %s\n", err)
+			if err.(kafka.Error).IsRetriable() {
+				if i == RETRY_NUM {
```
edenhill, line 354: This is never reached (for condition is i < RETRY_NUM)

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,495 @@
...
+			err = p.AbortTransaction(nil)
+			if err != nil {
+				if i == 0 {
+					soaktest.DatadogIncrement(
+						PRODUCER_TRANSACTION_ABORT_FAILED,
+						1)
+				}
+				fmt.Printf("AbortTransaction() failed: %s\n", err)
+				if err.(kafka.Error).IsRetriable() {
+					time.Sleep(SLEEP_TIME)
```
edenhill, line 442: Sleep should be after the next if

soaktest/soaktest.go
```diff
@@ -0,0 +1,228 @@
...
+	return true
+}
+
+// Convert Stats events to map
+func HandleStatsEvent(e *kafka.Stats, topic string) error {
+	var raw map[string]interface{}
+	err := json.Unmarshal([]byte(e.String()), &raw)
+	if err != nil {
+		fmt.Printf("json unmarshall error: %s", err)
+	}
```
edenhill, line 115: error not handled

soaktest/soaktest.go
```diff
@@ -0,0 +1,228 @@
...
+				timestamp.GobDecode(hdr.Value)
+				DatadogGauge(
+					LATENCY+*e.TopicPartition.Topic,
+					time.Now().Sub(timestamp).Seconds(),
+					[]string{fmt.Sprintf("%s%d",
+						"partition:",
+						e.TopicPartition.Partition)})
+			}
+
+			if hdr.Key == "msgid" {
```
edenhill, line 178: else if

soaktest/soaktest.go
```diff
@@ -0,0 +1,228 @@
...
+				msgid := binary.LittleEndian.Uint64(hdr.Value)
+				verifyMessage(e, hwmarks, msgid)
+			}
+		}
+	}
+	return true
+}
+
+// Verify if there is any duplicate or lost messages from consumer side
```
edenhill, line 187: there isn't

edenhill (changes requested):
Left some comments, but this is getting close now! Good work!

soaktest/README.md
```diff
@@ -0,0 +1,28 @@
+# Information for soak test for go client
+The soak test contains 2 clients:
```
edenhill, line 2: Maybe "two applications" is more appropriate, since it is really 4 clients underneath

soaktest/README.md
```diff
@@ -0,0 +1,28 @@
...
+# Build soaktest API
+    $ cd confluent-kafka-go/soaktest
+    $ go mod init soaktest
+    $ go mod tidy
+    $ go install soaktest
+   
+# Build and run soakclient.go
+    $ cd confluent-kafka-go/soaktest/soakclient
+    $ go build
+    $ ./soakclient [-broker] [-inputTopic] [-outTopic] [-groupId] [numPartitions] [-replicationFactor] [-ccloudAPIKey] [-ccloudAPISecret]
```
edenhill, line 23: Syntax is typically:
`$ ./soakclient -broker <broker> -inputTopic <..> ` for required arguments.
`[..]` denotes optional arguments, e.g.:
`$ ./soakclient -broker <..> -inputTopic <..> [-timeout <..>]`

soaktest/soakclient/soakclient.go
```diff
@@ -0,0 +1,343 @@
...
+func main() {
+
+	broker := flag.String("broker", "127.0.0.1:9092", "bootstrap servers")
+	inputTopic := flag.String("inputTopic", "inputTopic",
+		"producer will deliver messages to this topic")
+	outTopic := flag.String("outTopic", "outTopic",
+		"consumer will consume messages from this topic")
+	groupId := flag.String("groupId", "groupId",
+		"the group consumer will join")
+	numPartitions := flag.Int("numPartitions", 1, "topic partition")
```
edenhill, line 43: "number of partitions in ?which? topic"

soaktest/soakclient/soakclient.go
```diff
@@ -0,0 +1,343 @@
...
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	select {
+	case sig := <-sigChan:
+		fmt.Printf("Signal caught: %v\n", sig)
+	case err := <-errorChan:
+		fmt.Printf("Error caught: %s\n", err)
+	}
+
```
edenhill, line 97: I think there's a deadlock here.
errorChan and sigChan are unbuffered, thus allowing only one event, subsequent events will block on the write side.
If a signal is received the above select will read that and then come down here.
If shortly after both producer and consumer writes an error event to the errorChan, one of them will block on the write since errorChan can only hold one event and nothing is reading from it.
Now we call wg.Wait() here, but since one of the consumer/producer is blocking on a channel write they will never complete their waitgroup, and the process hangs in a deadlock here.

soaktest/soaktest.go
```diff
@@ -0,0 +1,240 @@
...
+	hwmarks map[string]uint64) bool {
+	if e.TopicPartition.Error != nil {
+		ConsumerMsgErrs += 1
+		DatadogIncrement(CONSUMER_RECEIVE_ERROR_FROM+
+			*e.TopicPartition.Topic, 1)
+		fmt.Printf("Message receive failed from: %s\n",
+			e.TopicPartition.Error)
+		return false
+	}
+	DatadogIncrement(CONSUMER_RECEIVE_FROM+*e.TopicPartition.Topic, 1)
```
edenhill, line 172: I think the topic should be a label/tag, not part of the metric name itself.

soaktest/soakclient/soakclient.go
```diff
@@ -0,0 +1,343 @@
...
+			run = false
+		default:
+			ev := c.Poll(100)
+			if ev == nil {
+				continue
+			}
+
+			switch e := ev.(type) {
+			case *kafka.Message:
+				soaktest.ConsumerMsgCnt += 1
```
edenhill, line 314: Think this should be counted in HandleMessage

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,552 @@
...
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	select {
+	case sig := <-sigChan:
+		fmt.Printf("Signal caught: %v", sig)
+	case err := <-errorChan:
+		fmt.Printf("Error caught: %v", err)
+	}
+
```
edenhill, line 81: Same deadlock here, unbuffered channels will block on write until there is a reader.

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,552 @@
...
+}
+
+// getConsumerPosition gets consumer position according to partition id
+func getConsumerPosition(consumer *kafka.Consumer, partition int32,
+	topic string) (kafka.TopicPartition, error) {
+	position, err := consumer.Position([]kafka.TopicPartition{{
+		Topic: &topic, Partition: partition}})
+	if err != nil {
+		fmt.Printf("Failed to get position %s", err)
+		return kafka.TopicPartition{}, err
```
edenhill, line 95: I think it is more standard to return `nil, err` on error.

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,552 @@
...
+	} else if len(position) == 0 {
+		err = fmt.Errorf("The position doesn't contain any element")
+		return kafka.TopicPartition{}, err
```
edenhill, line 98: `nil, err`

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,552 @@
...
+		if err != nil {
+			fmt.Errorf("InitTransactions() failed: %s\n", err)
+			if err.(kafka.Error).IsRetriable() {
+				if i == RETRY_NUM-1 {
+					fmt.Printf("InitTransactions() failed after %d "+
+						"times retries: %s\n", RETRY_NUM, err)
+					terminate(c, p, err, errorChan)
+					return
+				}
+				time.Sleep(SLEEP_TIME)
```
edenhill, line 200: I find this naming harder to comprehend than actual values, "SLEEP_TIME" is ambigious/generic and when seen at the definition it doesn't say what the sleep time is used for, and when seen here it doesn't give us an idea how long this sleeps.
Either have a INIT_RETRY_SLEEP_TIME, or just specify the actual number here (I would go with the latter.. less boilerplate)

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,552 @@
...
+					e.Code(), e, c)
+
+			default:
+				fmt.Printf("Ignored %v for consumer %v\n", e, c)
+			}
+		}
+	}
+	p.Close()
+	c.Close()
+}
```
edenhill, line 350: no wg signalling done here?

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,552 @@
...
+					fmt.Printf("SendOffsetsToTransaction() failed with max "+
+						"retry %d times: %s\n", RETRY_NUM, err)
+					return transactionCommitNum, err
+				}
+				time.Sleep(SLEEP_TIME)
+				continue
+			} else if err.(kafka.Error).TxnRequiresAbort() {
+				err = p.AbortTransaction(nil)
+				if err != nil {
+					fmt.Printf("SendOffsetsToTransaction() failed: %s\n", err)
```
edenhill, line 420: AbortTran..

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,552 @@
...
+					return transactionCommitNum, err
+				}
+				rewindConsumerPosition(c)
+				return transactionCommitNum, nil
+			} else {
+				fmt.Errorf("SendOffsetsToTransaction() failed: %s\n", err)
```
edenhill, line 426: This is already printed above,.

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,552 @@
...
+						fmt.Printf("AbortTransaction() failed: %s\n", err)
+						return transactionCommitNum, err
+					}
+					err = rewindConsumerPosition(c)
+					if err != nil {
+						fmt.Printf("rewindConsumerPosition() failed: %s\n", err)
+					}
+					return transactionCommitNum, nil
+				} else {
+					fmt.Errorf("CommitTransaction() failed: %s\n", err)
```
edenhill, line 471: already printed above

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,552 @@
...
+					continue
+				} else {
+					fmt.Printf("AbortTransaction() failed: %s\n", err)
+					return transactionCommitNum, err
+				}
+			} else {
+				soaktest.DatadogIncrement(
+					PRODUCER_TRANSACTION_ABORT_SUCCEED,
+					1)
+				transactionCommitNum = 1
```
edenhill, line 513: Is this supposed to be an increment or reset?

soaktest/soakclient_transaction/soakclient_transaction.go
```diff
@@ -0,0 +1,552 @@
...
+// rewindConsumerPosition Rewinds consumer's position to the
+// pre-transaction offset
+func rewindConsumerPosition(c *kafka.Consumer) error {
+	assignment, err := c.Assignment()
+	if err != nil {
+		fmt.Printf("Assignment() failed: %s\n", err)
+		return err
+	}
+
+	committed, err := c.Committed(assignment, 10*1000 /* 10s */)
```
edenhill, line 535: I think it might be a good idea to allow a higher timeout here, 10s isnt much if the cluster is temporarily unreachable. Maybe 30s?

soaktest/soaktest.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		for _, val := range topparsValue.(map[string]interface{}) {
+			if val.(map[string]interface{})["topic"] != topic {
+				continue
+			}
+			partition := val.(map[string]interface{})["partition"]
+			f, ok := partition.(float64)
+			if !ok {
+				fmt.Printf("Failed to convert partition"+
```
edenhill, line 76: missing space

soaktest/soaktest.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		nid, ok := broker.(map[string]interface{})["nodeid"].(float64)
+		if !ok {
+			fmt.Printf("Failed to convert nodeid to float64 %v\n", ok)
+			return ok
+		}
+		nodeid = int(nid)
+
+		tags := []string{fmt.Sprintf("broker:%d, type: %s",
```
edenhill, line 106: I think these should be two different tags perhaps?
Tags are usually static names, e.g. "broker", "type", "partition", while the tag value is the variable part.

soaktest/soaktest.go
```diff
@@ -0,0 +1,240 @@
...
+			nodeid, raw["type"])}
+		DatadogGauge(BROKER_RTT_P99+topic, p99/1000000.0, tags)
```
edenhill, line 108: The topic should be a tag.

soaktest/soaktest.go
```diff
@@ -0,0 +1,240 @@
...
+	if e.TopicPartition.Error != nil {
+		ConsumerMsgErrs += 1
+		DatadogIncrement(CONSUMER_RECEIVE_ERROR_FROM+
+			*e.TopicPartition.Topic, 1)
+		fmt.Printf("Message receive failed from: %s\n",
+			e.TopicPartition.Error)
+		return false
+	}
+	DatadogIncrement(CONSUMER_RECEIVE_FROM+*e.TopicPartition.Topic, 1)
+	fmt.Printf("consumer out Message on %s:\n%s\n",
```
edenhill, line 173: We probably don't want to print each message or the logs will fill up real quick

edenhill (changes requested):
Looking good! Almost there

soaktest/soakclient_transaction/soakclient_transaction.go
jliunyu: Hi Magnus, This is reset

soaktest/soakclient_transaction/soakclient_transaction.go
jliunyu: Hi Magnus, thanks for pointing out this. Because we already added the wg signalling done from the caller side verifyProducerTransaction(), do we still need to add it here?

soaktest/soakclient_transaction/soakclient_transaction.go
jliunyu: Hi Magnus, looks kafka.TopicPartition{}, err is required for compiling, if changed to nil, err, the following compile error will be reported: 
./soakclient_transaction.go:94:3: cannot use nil as type kafka.TopicPartition in return argument
./soakclient_transaction.go:97:3: cannot use nil as type kafka.TopicPartition in return argument

soaktest/README.md
```diff
@@ -0,0 +1,39 @@
...
+soakclient_transaction.go for the other application: Create Consumer-1 to receive messages from the input-topic.
+Create transactional producer Producer-2 to sending the received messages to the output-topic.
+The initTransaction API registers a transactional.id when creating the producer.
+The producer sends the messages to the output-topic.
+Commit transaction after sending 100 messages each time or 1 second.
+
+# Instruction on configuration and running
+
+# Build soaktest API
+    $ cd confluent-kafka-go/soaktest
```
edenhill, line 15: nit: people are probably reading this from the confluent-kafka-go topdir, so remove the confluent-kafka-go/ prefix here and below.

soaktest/README.md
```diff
@@ -0,0 +1,39 @@
...
+    $ go build
+    $ ./soakclient -broker <..> -inputTopic <..> -outTopic <..> -groupId <..> -inputTopicPartitionsNum <..> -outTopicPartitionsNum <..> -replicationFactor <..> -ccloudAPIKey <ccloudAPIKey> -ccloudAPISecret <ccloudAPISecret>
+   
+# Build and run soakclient_transaction.go
+    $ cd confluent-kafka-go/soaktest/soakclient_transaction
+    $ go build
+    $ ./soakclient_transaction -broker <..> -inputTopic <..> -outTopic <..> -groupId <..> -transactionId <..> -ccloudAPIKey <ccloudAPIKey> -ccloudAPISecret <ccloudAPISecret>
+
+### The default values if doesn't input from the command lines
+    broker: "127.0.0.1:9092"
```
edenhill, line 31: I'm not very fond of default server addresses, even though the default broker is local it might be running for completely different purposes (some other test) and running these clients on that cluster without the user knowing might mess things up, so don't provide a default broker address.

soaktest/soakclient/soakclient.go
```diff
@@ -0,0 +1,357 @@
...
+	outTopicPartitionsNum := flag.Int("outTopicPartitionsNum", 1,
+		"outTopic partition number")
+	replicationFactor := flag.Int("replicationFactor", 1, "topic replication")
+	ccloudAPIKey := flag.String("ccloudAPIKey", "", "sasl username")
+	ccloudAPISecret := flag.String("ccloudAPISecret", "", "sasl password")
+
+	flag.Parse()
+
+	soaktest.InitLogFiles("../log", fmt.Sprintf("../log/soakclient_%s.log",
+		time.Now().Format("01-02-2006")))
```
edenhill, line 54: Not sure what that format is (the great american DD/MM mixup), can we go with YYYY-MM-DD?

soaktest/soakclient/soakclient.go
```diff
@@ -0,0 +1,357 @@
...
+		"sasl.password":          ccloudAPISecret})
+	if err != nil {
+		soaktest.ErrorLogger.Printf("Failed to create producer %s\n", err)
+		errorChan <- err
+		return
+	}
+
+	run := true
+	doneChan := make(chan bool)
+	tags := []string{fmt.Sprintf("topic: %s", inputTopic)}
```
edenhill, line 195: In datadog-go/statsd examples it looks like the tag should not contain a space, so: topic:%s

soaktest/soakclient/soakclient.go
```diff
@@ -0,0 +1,357 @@
...
+		"sasl.mechanisms":        "PLAIN",
+		"security.protocol":      "SASL_SSL",
+		"sasl.username":          ccloudAPIKey,
+		"sasl.password":          ccloudAPISecret,
+	})
+	if err != nil {
+		soaktest.ErrorLogger.Printf("Failed to create consumer: %s\n", err)
+		errorChan <- err
+		return
+	} else {
```
edenhill, line 300: else not needed, the if-clause returns. (keep nesting down)

soaktest/soaktest.go
```diff
@@ -0,0 +1,258 @@
...
+const CONSUMER_RECEIVE_ERROR = "consumer.receive.err"
+const PRODUCER_DROK = "producer.drok"
+const LATENCY = "latency"
+const COMSUMER_CONSUME_DUP_MSG = "consumer.consume.dup.msg"
+const COMSUMER_CONSUME_MISS_MSG = "consumer.consume.miss.msg"
+const BROKER_RTT_P99 = "broker.rtt.p99"
+const BROKER_RTT_AVG = "broker.rtt.avg"
+
+func InitLogFiles(path, filename string) {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
edenhill, line 60: the directory can be derrived from the filename with Dirname(), so the path argument isn't really needed.

soaktest/soaktest.go
```diff
@@ -0,0 +1,258 @@
...
+		os.Mkdir(path, os.ModePerm)
+	}
+
+	file, err := os.OpenFile(filename, os.O_APPEND|os.O_CREATE|os.O_WRONLY,
+		0666)
```
edenhill, line 65: 0644

soaktest/soaktest.go
```diff
@@ -0,0 +1,258 @@
...
+	if err != nil {
+		log.Fatal(err)
```
edenhill, line 67: Is this panicking? Else missing a return.

soaktest/soaktest.go
```diff
@@ -0,0 +1,258 @@
...
+	var raw map[string]interface{}
+	err := json.Unmarshal([]byte(e.String()), &raw)
+	if err != nil {
+		ErrorLogger.Printf("Json unmarshall error: %s\n", err)
+		return err
+	}
+
+	ok := rttStats(raw, topic)
+	if !ok {
+		return fmt.Errorf("Failed to calculate broker rtt statistics")
```
edenhill, line 148: I think it is better to return this error from rttStats instead of a bool so that the actual error is not lost and converted to this generic one.

soaktest/soaktest.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	return nil
+}
+
+// ProducerDeliveryCheck handles delivery report for producer
+func ProducerDeliveryCheck(e *kafka.Message) {
+	tags := []string{fmt.Sprintf("topic: %s", *e.TopicPartition.Topic)}
```
edenhill, line 155: the partition should also be a tag.

soaktest/soaktest.go
```diff
@@ -0,0 +1,258 @@
...
+	InfoLogger.Printf("Producer: %d messages produced, %d delivered, %d "+
+		"failed to deliver, %d producer client-level errors\n",
+		ProducerMsgId, DrCnt, DrErrCnt, ProducerErr)
+}
+
+// HandleMessage handles received messages, monitors latency and
+// verifies messages
+func HandleMessage(e *kafka.Message,
+	hwmarks map[string]uint64) bool {
+	tags := []string{fmt.Sprintf("topic: %s", *e.TopicPartition.Topic)}
```
edenhill, line 188: add partition tag
remove space

soaktest/soaktest.go
```diff
@@ -0,0 +1,258 @@
...
+	if e.TopicPartition.Error != nil {
+		ConsumerMsgErrs += 1
+		DatadogIncrement(CONSUMER_RECEIVE_ERROR, 1, tags)
+		ErrorLogger.Printf("Message receive failed from: %s\n",
```
edenhill, line 192: Should also include the topic+partition so we know which top+par failed.

soaktest/soaktest.go
```diff
@@ -0,0 +1,258 @@
...
+			e.TopicPartition.Error)
+		return false
+	}
+	ConsumerMsgCnt += 1
+	DatadogIncrement(CONSUMER_CONSUME_MSG, 1, tags)
+	if ConsumerMsgCnt%1000 == 0 {
+		InfoLogger.Printf("Consumer out Message on %s:\n%s\n%s\n",
```
edenhill, line 199: Multiline logs should be avoided if possible since they make grepping and sorting cumbersome.

edenhill (changes requested):
Very close now!

soaktest/soaktest.go
jliunyu: Yes, the application will error out if met log.Fatal(err)

soaktest/soaktest.go
```diff
@@ -251,7 +251,8 @@ func HandleMessage(e *kafka.Message,
 // lost messages from consumer side
 func verifyMessage(e *kafka.Message,
 	hwmarks map[string]uint64, msgid uint64) {
-	tags := []string{fmt.Sprintf("topic:%s", *e.TopicPartition.Topic)}
+	tags := []string{fmt.Sprintf("topic:%s partition:%d",
```
edenhill, line 254: Is this how multiple tags are passed? Seeing how tags is an array of strings shouldnt it be one key:val per array item?
(I don't know, it just looks a bit uncommon)

soaktest/soaktest.go
```diff
@@ -0,0 +1,285 @@
...
+		*e.TopicPartition.Topic, e.TopicPartition.Partition)}
+	if e.TopicPartition.Error != nil {
+		consumerMsgErrs++
+		DatadogIncrement(consumerReceiveError, 1, tags)
+		ErrorLogger.Printf("Message receive failed from: %s\n",
+			e.TopicPartition.Error)
+		return false
+	}
+	consumerMsgCnt++
+	DatadogIncrement(consumerConsumeMSG, 1, tags)
```
edenhill, line 223: The capital MSG looks like an acronym and I think golint will complain, should probably be consumerConsumeMsg

soaktest/soaktest.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		nid, ok := broker.(map[string]interface{})["nodeid"].(float64)
+		if !ok {
+			err := fmt.Errorf("failed to convert nodeid to float64 %v", ok)
+			return err
+		}
+		nodeid = int(nid)
+
+		tags := []string{fmt.Sprintf("broker:%d, type:%s, topic:%s, parts:%s",
```
edenhill, line 148: I don't think topic and partitions should  be part of the per-broker rtts since any change in leadership will create a new data series.
What I do in the Python soaks is have one series for the broker rtt, and one series for partition leaders. This allows us to manually map current-partition-leader-broker-rtts in datadog by comparing the two charts.

soaktest/soaktest.go
```diff
@@ -0,0 +1,285 @@
...
+			nodeid, raw["type"], topic, parts)}
+
+		DatadogGauge(brokerRttP99, p99/1000000.0, tags)
+		DatadogGauge(brokerRttAvg, avg/1000000.0, tags)
+	}
+	return nil
+}
+
+// HandleStatsEvent converts Stats events to hashmap
```
edenhill, line 157: nit, they're just called maps in Go

soaktest/soaktest.go
```diff
@@ -0,0 +1,285 @@
...
+		*e.TopicPartition.Topic, e.TopicPartition.Partition)}
+	if e.TopicPartition.Error != nil {
+		drErrCnt++
+		DatadogIncrement(producerDrErr, 1, tags)
+		ErrorLogger.Printf("Delivery failed: %v\n", e.TopicPartition)
+	} else {
+		drCnt++
+		DatadogIncrement(producerDr, 1, tags)
+		if drCnt%1000 == 0 {
+			InfoLogger.Printf("Delivered message to topic at events channel "+
```
edenhill, line 187: remove "at events channel"

soaktest/soaktest.go
```diff
@@ -0,0 +1,285 @@
...
+				"%s [%d] at offset %d\n", *e.TopicPartition.Topic,
+				e.TopicPartition.Partition, e.TopicPartition.Offset)
+		}
+	}
+}
+
+// PrintConsumerStatus prints the information for the consumer
+func PrintConsumerStatus() {
+	InfoLogger.Printf("Consumer: %d messages consumed, %d duplicates, "+
```
edenhill, line 196: Since there are two consumers, can we prefix this with either TxnConsumer or OutputConsumer?

soaktest/soaktest.go
```diff
@@ -0,0 +1,285 @@
...
+		"%d missed, %d message errors, %d consumer client-level errors\n",
+		consumerMsgCnt, msgDupCntForOutput, msgMissCntForOutput,
+		consumerMsgErrs, ConsumerErrCnt)
+}
+
+// PrintProducerStatus prints the information for the producer
+func PrintProducerStatus() {
+	InfoLogger.Printf("Producer: %d messages produced, %d delivered, %d "+
```
edenhill, line 204: Same,  InputProducer or TxnProducer

soaktest/soaktest.go
```diff
@@ -0,0 +1,285 @@
...
+		"failed to deliver, %d producer client-level errors\n",
+		ProduceMsgCnt, drCnt, drErrCnt, ProducerErrCnt)
+}
+
+// HandleMessage handles received messages, monitors latency and
+// verifies messages
+func HandleMessage(e *kafka.Message,
+	hwmarks map[string]uint64) bool {
+	tags := []string{fmt.Sprintf("topic:%s, partition:%d",
```
edenhill, line 213: Should topic: and partition: perhaps be separate entries in the tags list vector? It seems odd that it is multidimensional like this

soaktest/soaktest.go
```diff
@@ -0,0 +1,285 @@
...
+		consumerMsgErrs++
+		DatadogIncrement(consumerReceiveError, 1, tags)
+		ErrorLogger.Printf("Message receive failed from: %s\n",
+			e.TopicPartition.Error)
+		return false
+	}
+	consumerMsgCnt++
+	DatadogIncrement(consumerConsumeMSG, 1, tags)
+	if consumerMsgCnt%1000 == 0 {
+		InfoLogger.Printf("Consumer out Message on TopicPartition: %s, "+
```
edenhill, line 225: out?

edenhill (changes requested):
LGTM, but left some minor comments that I think should be addressed.

soaktest/soaktest.go
```diff
@@ -145,16 +146,16 @@ func rttStats(raw map[string]interface{}, topic string) error {
 		}
 		nodeid = int(nid)
 
-		tags := []string{fmt.Sprintf("broker:%d, type:%s, topic:%s, parts:%s",
-			nodeid, raw["type"], topic, parts)}
+		tags := []string{fmt.Sprintf("broker:%d, type:%s, parts:%s",
```
jliunyu, line 149: If topparsValue.(map[string]interface{}) doesn't have any values, the parts will be empty, partitions won't be part of the per-broker rtts. But if there is value at topparsValue.(map[string]interface{}), this will contribute to the series for partition leaders.

edenhill: rtt is per broker though, not per partition, and if the partitions are part of the tags it means that any partition leader change will make a new time series for the broker, even though it is the same one.
timeseries 1: "broker:0, parts:1,2"
timeseries 2: "broker:0, parts:1"

But it is still the same broker and should be the same time series, so remove the partitions from the tags.

Only, I'm still not convinced that the tags should be concatenated in a single string like that, I think each key:value should be its own tag in that array.
E.g.: `[]string{"broker:0", "type:foo"}`

soaktest/soaktest.go
edenhill: This is not fixed, the TopicPartition itself is not printed, just the error message.

edenhill (changes requested):

soaktest/soaktest.go
```diff
@@ -0,0 +1,267 @@
...
+const consumerConsumeMissMSG = "consumer.consume.miss.msg"
+const brokerRttP99 = "broker.rtt.p99"
+const brokerRttAvg = "broker.rtt.avg"
+
+// InitLogFiles initials the log folder and file, if the folder or file
+// doesn't exist, create one, otherwise open it directly
+func InitLogFiles(filename string) {
+	path := filepath.Dir(filename)
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		os.Mkdir(path, os.ModePerm)
```
edenhill, line 83: This should be error-checked, and I think a linter will complain.

soaktest/soaktest.go
edenhill: Not fixed?

edenhill (changes requested):
Two minor comments!

soaktest/soaktest.go
jliunyu: I missed this comment before, sorry

PR #626, opened 2021-04-10
Make event polling interruptable with rd_kafka_queue_yield

KJTsanaktsidis:
(this PR follows on from feedback in https://github.com/confluentinc/confluent-kafka-go/pull/518 )

## Interface

This PR makes the process of polling for librdkafka events interruptible from the Go side. This is achieved by creating a new method `PollContext`, which works like `Poll` except instead of providing a timeout duration, a `context.Context` is provided. if this context is canceled, either by explicit cancellation or by timeout/deadline, then `PollContext` will return with an error (`context.Canceled`).

## Implementation

The guts of this implementation is adding a goroutine which runs in parallel to any call to `rd_kafka_queue_poll` in `eventPoll`/`pollLogEvents` which monitors for context cancellation, and if so, runs `rd_kafka_queue_yield` to break out of the call to `rd_kafka_queue_poll`.

The main structural change to `eventPoll` is that it no longer accepts a channel as an argument, and instead always returns only zero or one events for application processing. This is mostly a case of changing instances of `retval = ...` to `return ...` in the main processing select statement. However, there is one slightly more substantial change.

Previously, the producer implemented the `go.delivery.reports` setting by passing along the producer events channel to `eventPoll`, and having `eventPoll` check the value of `h.fwdDr` when deciding what to do with delivery reports. However, we can't do this anymore, because we're not passing a channel to `eventPoll`.

Instead, we now handle `h.fwdDr` at produce time by setting the `_private` field on the message to contain a reference to the `p.events` channel. This is how dispatching DR's to an alternate per-message channel already worked, so it doesn't seem a big change to do this, and it allows the `eventPoll` code to unconditionally just dispatch DR's to whatever `_private` contains.


KJTsanaktsidis:
[clabot:check]

edenhill:
This is great, @KJTsanaktsidis ! Will review tomorrow,.

kafka/event.go
```diff
@@ -145,170 +147,179 @@ func (o OAuthBearerTokenRefresh) String() string {
...
+			if ch != nil {
 				select {
-				case channel <- retval:
-				case <-termChan:
-					retval = nil
-					term = true
-					break out
+				case *ch <- msg:
+				case <-ctx.Done():
+					// This is a _very_ dangerous codepath. If the context passed in is canceled, and causes us
```
edenhill, line 279: I'm not comfortable with this, all the delivery reports in this event should be forwarded before checking the context.
As for the deadlock/unserved channel on termination; that is an existing problem and not something that needs to be fixed now, and there are proper workarounds (call Flush() and read the dr chan in separate go-routines).

kafka/integration_test.go
```diff
@@ -233,7 +233,7 @@ func producerTest(t *testing.T, testname string, testmsgs []*testmsgType, pc pro
 
 	if testmsgs == nil {
 		createTestMessages()
-		testmsgs = pAllTestMsgs
+		testmsgs = p0TestMsgs
```
edenhill, line 236: Why this change?

kafka/integration_test.go
```diff
@@ -395,12 +397,14 @@ func consumerTest(t *testing.T, testname string, assignmentStrategy string, msgc
...
+		// Sleep for a while to actually let the auto commit happen
+		// We set auto.commit.interval.ms to 500, so give it a bit longer.
+		time.Sleep(1 * time.Second)
+	}
+
 	// Trigger RevokePartitions
 	c.Unsubscribe()
-
-	// Handle RevokePartitions
-	c.Poll(500)
```
edenhill, line 402: Why is this removed?

We should be very careful with changing existing tests since any change to tests may hide regression, it is better to add another test-case instead.

kafka/log.go
```diff
@@ -47,33 +48,55 @@ func (h *handle) newLogEvent(cEvent *C.rd_kafka_event_t) LogEvent {
...
-			return
+		// See if there's anything ready-to-go without polling
+		cEvent := C.rd_kafka_queue_poll(h.logq, C.int(0))
 
-		default:
-			cEvent := C.rd_kafka_queue_poll(h.logq, C.int(timeoutMs))
-			if cEvent == nil {
-				continue
-			}
+		if cEvent == nil {
```
edenhill, line 56: Since this logic is occurring in a couple of places, maybe move it to its own function, eventPollSingle() or something, that performs the immediate and then blocking polls.

kafka/producer.go
```diff
@@ -296,6 +304,13 @@ func (p *Producer) produceBatch(topic string, msgs []*Message, msgFlags int) err
 	cmsgs := make([]C.rd_kafka_message_t, len(msgs))
 	for i, m := range msgs {
 		p.handle.messageToC(m, &cmsgs[i])
+		// Batch production doesn't support a per-message delivery report channel, but it does support
+		// dispatching delivery reports to the main channel. Writing the channel cgoid to _private is
+		// done by set_message_private_cgoid and ensures that eventPoll can find it.
+		if p.handle.fwdDr {
+			cgoid := p.handle.cgoPut(cgoDr{deliveryChan: p.events})
+			C.set_message_private_cgoid(&cmsgs[i], C.uintptr_t(cgoid))
```
edenhill, line 312: Each C call is costly. Better to pass the cgoid to messageToC() above.

edenhill (changes requested):
This is great stuff!

edenhill:
Since this is a pretty big change it will not make the 1.7.0 release (<1mo away), but the release after that for sure.

kafka/integration_test.go
KJTsanaktsidis: I'll rebase this out - I had thought that having the invalid partition message in pAllTestMsgs would cause problems but it seems like it all works.

kafka/integration_test.go
KJTsanaktsidis: This does need to go - in the test cases where `go.events.channel.enable` is set, calling `Poll()` here will be called concurrently with `Poll()` in the `consumerReader` loop. The two concurrent calls to Poll() is a bit daft, because it's totally undefined which one will actually receive any messages.

The reason for this being here in the first place, to handle RevokePartitions, is I believe no longer necessary anyway, because `Close()` does this itself now (I think this is a new addition), and does it _after_ shutting down the `consumerReader` loop so that there's only one call to `rd_kafka_queue_poll` in flight at a time.

KJTsanaktsidis: I removed it initially because it caused deadlocks with my first implementation of `rd_kafka_queue_yield`, because that woke up all `rd_kafka_queue_poll` calls I think. I subsequently sure that my implementation could handle concurrent calls to Poll(), but left the test case change here because I don't think concurrent calls to Poll() is something that really _should_ be supported.

kafka/log.go
KJTsanaktsidis: Yeah I'll factor something out here.

kafka/producer.go
KJTsanaktsidis: good call. I'll shuffle it around.

kafka/event.go
```diff
@@ -145,170 +147,179 @@ func (o OAuthBearerTokenRefresh) String() string {
...
+		// Need to actually wait for a while.
+		messageWaitCh := make(chan struct{})
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			select {
+			case <-messageWaitCh:
+			case <-ctx.Done():
+				C.rd_kafka_queue_yield(h.rkq)
```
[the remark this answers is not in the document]
KJTsanaktsidis: Switch statements in Go don't fallthrough, so it's actually working as you describe. I'll add a comment in the first case to make that explicit.

kafka/event.go
KJTsanaktsidis: I need to have a long think about this in front of the whiteboard with @seancaffery I think - I suspect my understanding of how the shutdown process is supposed to work might be wrong.

Once I've done that, yeah, I'll probably remove this check.

kafka/log.go
KJTsanaktsidis: Looking at this more carefully - doing this will have a performance impact. The log poller just does a straight-up `rd_kafka_queue_poll`, but the event poller does a call to `_rk_queue_poll` instead which makes some librdkafka calls to fill out message structure info if the returned event is a `RD_KAFKA_EVENT_FETCH`.

Options here would be:
 
1. Leave it how it is, with the duplicated code
2. Change the log poller to also call `_rk_queue_poll`
3. Change the event poller to call `rd_kafka_queue_poll` directly (and take the performance hit of then needing to subsequently make another cgo call in the message case).
4. Make your proposed `eventPollsingle()` generic by allowing one to pass in a callback function.

I think i'm going to go with option 2 here, but let me know if you feel differently.

kafka/event.go
KJTsanaktsidis: OK - I think we've nutted out how the shutdown is supposed to work here. Agree that we should block on `*ch <- msg` irrespective of the context being canceled or not. If this behaviour is _not_ what a user of the library wants, they can pass in a buffered channel to `Produce` - then this line will never block regardless of what the application is doing (and the DR's will be thrown in the bin just like the above code is doing now).

kafka/handle.go
```diff
@@ -247,26 +309,44 @@ type cgoDr struct {
...
+//
+// This method therefore returns an unsafe.Pointer that is guaranteed to never
+// point to a valid Go object, by invoking C.malloc to produce it. This can
+// safely be passed through cgo calls and back into cgoGet() to return the
+// passed-in object.
+//
 // Thread-safe.
-// FIXME: the uniquity of the id is questionable over time.
-func (h *handle) cgoPut(cg cgoif) (cgoid int) {
+func (h *handle) cgoPut(cg cgoif) unsafe.Pointer {
```
KJTsanaktsidis, line 325: This signature change (of `int` to `unsafe.Pointer`) is needed to avoid repeated cgo calls in the batch produce case as you asked for. I explained this in the commit message but I'll paste it here for easy review too.

    To round-trip values through the _private void* field in the
    rd_kafka_message_t type, we currently generate an int in cgoPut(), save
    the value in a map, and cast the int to a uintptr_t when passing into
    cgo, which is then cast (inside C code) to a void* before sending to
    librdkafka.
    
    To add support for passing _private in the batch-produce case, we need
    to be able to assign the _private field (of type void*/unsafe.Pointer)
    in Go, rather than in C. In Go, it is illegal to cast any int to an
    unsafe.Pointer (even if it is never dereferenced).
    
    Therefore, we must change cgoPut/cgoGet to deal with unsafe.Pointer
    tokens, rather than int tokens. The most straightforward way to generate
    such pointers which are guaranteed to never point to Go objects is to
    allocate them on the C heap with C.malloc.
    
    To avoid a bunch of pointless malloc/free calls, we also implement a
    token cache so that these pointers can be reused.

KJTsanaktsidis:
@edenhill I think I've addressed your feedback - PTAL when you get a chance? 🙏 thanks!

KJTsanaktsidis:
Also - since I've been mucking around with the cgo machinery - I ran the test suite a few times on my machine with `GODEBUG=cgocheck=2`. This actually shook out some problems with other implementations of `cgoPut`/`cgoGet` that I tried so I feel pretty comfortable that these ones are correct.

KJTsanaktsidis:
I'm going to go ahead and assume that the windows test failures on my branch are unrelated...

KJTsanaktsidis:
@mhowlett @edenhill just a gentle bump to see if you could find some time to 👀  this 🙏  thanks!

KJTsanaktsidis:
@edenhill Any update? :)

hariso:
I'm really interested in this, and it would be great if we could see this merged.: )

KJTsanaktsidis:
To be clear I'm not committed to actively maintaining a fork (e.g. I haven't been regularly merging master into these trees, and I'm not going to merge PR's from outside contributors) but - we've been using these two trees for months now in production at Zendesk and they've been working great

https://github.com/zendesk/librdkafka/tree/zendesk_release
https://github.com/zendesk/confluent-kafka-go/tree/zendesk_release

@hariso you're welcome to use them or base your own tree off them if that's helpful - it's rolled in a few different PR's to confluent-kafka-go plus some bugfixes to the PR's we found while getting them into production :)

It does remain my aim to get this stuff upstreamed though :)

hariso:
@KJTsanaktsidis It's great to hear you've been using them successfully for months now. Even though it's not urgent for me at the moment, thank you very much for letting me know they can be freely used; I'll definitely consider that.

Issue #616, opened 2021-03-25
Support KIP-464: Defaults for AdminClient#createTopic


tobgu:
Description
===========
Support for this already exists in librdkafka by I think the input validation in `AdminClient.CreateTopics(...)` in the Go client makes it impossible to use. librdkafka assumes `-1` for both partition count and replication factor to mean "use default server side values". It's not, as far as I can tell, possible to pass a value of `-1` or `0` for `ReplicationFactor` to CreateTopics without specifying an explicit `ReplicaAssignment` however.

How to reproduce
================

```go
adminClient.CreateTopics(ctx, []kafka.TopicSpecification{{Topic: "test", NumPartitions: -1, ReplicationFactor: -1}})
```

Checklist
=========
Please provide the following information:

 - [X] confluent-kafka-go and librdkafka version (`LibraryVersion(1.6.1)`):
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `ConfigMap{...}`
 - [ ] Operating system:
 - [ ] Provide client logs (with `"debug": ".."` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

PR #635, opened 2021-04-23
Default parameters for AdminClient#createTopic

jliunyu:
This is the fix for https://github.com/confluentinc/confluent-kafka-go/issues/616

librdkafka supports AdminClient.CreateTopics(...) with default value(-1) of NumPartitions and ReplicationFactor, but currently it's not possible to pass a value of -1 or 0 for ReplicationFactor to CreateTopics without specifying an explicit ReplicaAssignment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Confluent's Golang client for Apache Kafka
 
+## v1.7.0
+
+### Fixes
+
+* AdminClient.CreateTopics() previously did not accept default value(-1) of
+  ReplicationFactor without specifying an explicit ReplicaAssignment, this is
+  now fixed.
+
+
+
 ## v1.6.1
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ for a complete list of changes, enhancements, fixes and upgrade considerations.
 
  * Consumer.Close() previously did not trigger the final RevokePartitions
    callback, this is now fixed.
+ * AdminClient.CreateTopics() previously did not accept default value(-1) of
```
edenhill, line 29: This is in the v1.6.1 section of the CHANGELOG, but v1.6.1 has already been released so this change will go in the next release.
Create a new release section (call it v1.7.0) add this line there instead.

edenhill (changes requested):
Good stuff!

PR #636, opened 2021-04-23
Add headers to go.delivery.report.fields (disabled by default)

edenhill:
Disabled by default due to the extra CGo calls required (costly) and the deserialization of C headers to Go Headers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## v1.7.0
 
+confluent-kafka-go is based on librdkafka v1.7.0, see the
+[librdkafka release notes](https://github.com/edenhill/librdkafka/releases/tag/v1.7.0)
+for a complete list of changes, enhancements, fixes and upgrade considerations.
+
+### Enhancements
+
+ * The produced message headers are now available in the delivery report
+   `Message.Headers` if the Producer's `go.delivery.report.fields`
+   configuration property is set to include `headers`, e.g.:
+   `"go.delivery.report.fields": "key,value,headers"`
+   This comes at a performance cost and are thus disabled by default.
+
+
 ### Fixes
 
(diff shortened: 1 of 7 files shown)


edenhill:
A future enhancement is to add `go.message.fields` to the consumer to be able to disable header deserialization for high throughput consumery thingies.

jliunyu (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+for a complete list of changes, enhancements, fixes and upgrade considerations.
+
+### Enhancements
+
+ * Experimental Windows support (by @neptoess).
+ * The produced message headers are now available in the delivery report
+   `Message.Headers` if the Producer's `go.delivery.report.fields`
+   configuration property is set to include `headers`, e.g.:
+   `"go.delivery.report.fields": "key,value,headers"`
+   This comes at a performance cost and are thus disabled by default.
```
mhowlett, line 16: how much of a performance cost - are you preemptively over optimizing perf over ease of use/things just working out of the box? in .net, it defaults to including all fields. Is still think that is the best tradeoff. There are other things that affect performance, and I think to address users who care about this, we should have an easily discoverable "## How Do You Make It Go Fast" section on the README.md, which would counter the argument for having high perf here as a default (my main concern is people may not realize this). in .net, there are also "none" and "all" values. defaults is "all". no need to add this now, but i wanted to point out also that it is kind of nice to have "all" as a default rather than a subset of the list.

all that said, not a huge point and not going to argue it further if you disagree.

mhowlett (approved):

CHANGELOG.md
edenhill: I theoretically agree, but the CGo overhead is too big in practice, and we don't want to employ that cost on the 99.9% users who don't need this functionality.

Issue #610, opened 2021-03-17
v1.6.1 - kafka.AssignedPartitions or kafka.RevokedPartitions events not returned with Poll()


richiejenkins:
Description
===========

The behaviour of `"go.application.rebalance.enable": true` appears to have changed with `v1.6.1`, with neither `kafka.AssignedPartitions` or `kafka.RevokedPartitions` being returned by `Poll()`.

The only way to capture these events is separately with a `rebalanceCb` method when calling `Subscribe()`.

Is this an intended change?

How to reproduce
================

- Create a consumer with `"go.application.rebalance.enable": true`
- Poll for events
- Observe no `kafka.AssignedPartitions` or `kafka.RevokedPartitions`

Checklist
=========
Please provide the following information:

 - [X] confluent-kafka-go and librdkafka version (1.6.1):
 - [ ] Apache Kafka broker version: 
 - [X] Client configuration: `ConfigMap{...}`
 - [X] Operating system: macOS 10.15.7 
 - [ ] Provide client logs (with `"debug": ".."` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

larry-cdn77:
Same problem here, those notification no longer come out of the polling function. They may have broken or even been removed for the incremental rebalancing feature introduced in version 1.6.1. As shown in the [incremental balancing example](https://github.com/confluentinc/confluent-kafka-go/blob/v1.6.1/examples/cooperative_consumer_example/cooperative_consumer_example.go), rebalance callback function is another way of getting assign/revoke notifications, and works for me.

nsagnett:
Same issue here. some statements has been cleared between two versions for the rebalance process
Working code : https://github.com/confluentinc/confluent-kafka-go/blob/8bd1e41d4448e6f6afb8539ad3ce7ada168c2724/kafka/event.go#L154
New code : 
https://github.com/confluentinc/confluent-kafka-go/blob/f0d7516a9eb10d64e32bd8ea25607ce3930c7bf7/kafka/consumer.go#L838

@edenhill, will this be fixed ?

Many thanks in advance

Regards

jliunyu:
@nsagnett Thanks for your comment, yes, we are working on the fix for this issue

jfrancisco0:
I believe I'm facing this with `v1.5.2`.

jliunyu:
[...]

PR #645, opened 2021-05-14
Fix Rebalance events behavior when config go.application.rebalance.enable to true

jliunyu:
The behaviour of "go.application.rebalance.enable": true appears to have changed with v1.6.1, with neither kafka.AssignedPartitions or kafka.RevokedPartitions being returned by Poll().

```
Local Test:
jliu@C02DM32YMD6T kafka % go test -run TestConsumerRebalanceEvents
Consumer closed
Consumer closed
COOPERATIVE rebalance: 4 new partition(s) assigned: [testConsumerRebalanceEvents[0]@unset testConsumerRebalanceEvents[1]@unset testConsumerRebalanceEvents[2]@unset testConsumerRebalanceEvents[3]@unset]
COOPERATIVE rebalance: 4 partition(s) revoked: [testConsumerRebalanceEvents[0]@unset testConsumerRebalanceEvents[1]@unset testConsumerRebalanceEvents[2]@unset testConsumerRebalanceEvents[3]@unset]
Consumer closed
EAGER rebalance: 4 new partition(s) assigned: [testConsumerRebalanceEvents[0]@unset testConsumerRebalanceEvents[1]@unset testConsumerRebalanceEvents[2]@unset testConsumerRebalanceEvents[3]@unset]
EAGER rebalance: 4 partition(s) revoked: [testConsumerRebalanceEvents[0]@unset testConsumerRebalanceEvents[1]@unset testConsumerRebalanceEvents[2]@unset testConsumerRebalanceEvents[3]@unset]
Consumer closed
PASS
ok  	_/Users/jliu/Documents/Code/jliunyu/confluent-kafka-go/kafka	0.198s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Confluent's Golang client for Apache Kafka
 
+## v1.8.0
+
+v1.8.0 is a maintenance release with the following fixes and enhancements:
+
+### Fixes
+
+* Fix Rebalance events behavior when config go.application.rebalance.enable to true.
+
+confluent-kafka-go is based on librdkafka v1.8.0, see the
+[librdkafka release notes](https://github.com/edenhill/librdkafka/releases/tag/v1.8.0)
+for a complete list of changes, enhancements, fixes and upgrade considerations.
+
+
 ## v1.7.0
 
 confluent-kafka-go is based on librdkafka v1.7.0, see the
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -905,5 +905,5 @@ func (c *Consumer) handleRebalanceEvent(channel chan Event, rkev *C.rd_kafka_eve
 		c.events <- newError(cErr)
 	}
 
-	return nil
+	return ev
 }
--- a/kafka/consumer_test.go
+++ b/kafka/consumer_test.go
@@ -419,3 +419,166 @@ func TestConsumerLog(t *testing.T) {
 		}
 	}
 }
+
+func TestConsumerRebalanceEvents(t *testing.T) {
+	if !testconfRead() {
+		t.Skipf("Missing testconf.json")
+	}
+	broker := testconf.Brokers
+	topic := testconf.Topic
+
+	testConsumerRebalanceEvents("range", broker, topic, false)
+	testConsumerRebalanceEvents("cooperative-sticky", broker, topic, false)
+
+	testConsumerRebalanceEvents("cooperative-sticky", broker, topic, true)
+	testConsumerRebalanceEvents("range", broker, topic, true)
+}
+
+func testConsumerRebalanceEvents(partitionAssignmentStrategy, broker, topic string,
+	callback bool) {
+	c, err := NewConsumer(&ConfigMap{
+		"bootstrap.servers":               testconf.Brokers,
+		"group.id":                        "groupID",
+		"go.application.rebalance.enable": true,
+		"partition.assignment.strategy":   partitionAssignmentStrategy,
+	})
+	if err != nil {
+		fmt.Printf("Failed to create consumer: %s\n", err)
+		return
+	}
+
+	if callback && partitionAssignmentStrategy == "range" {
+		err = c.Subscribe(topic, rebalanceCallbackEager)
+	} else if callback && partitionAssignmentStrategy == "cooperative-sticky" {
+		err = c.Subscribe(topic, rebalanceCallbackCooperative)
+	} else {
+		err = c.Subscribe(topic, nil)
+	}
+
+	if err != nil {
+		fmt.Printf("Failed to subscribe to topic: %s with "+
+			"error %s\n", topic, err)
+		return
+	}
+
+	run := true
+	for run {
+		ev := c.Poll(100)
+		if ev == nil {
+			continue
+		}
+
+		switch e := ev.(type) {
+		case *Message:
+			fmt.Printf("Message on %s:\n%s\n",
+				e.TopicPartition, string(e.Value))
+			if e.Headers != nil {
+				fmt.Printf("Headers: %v\n", e.Headers)
+			}
+			run = false
+
+		case RevokedPartitions:
+			fmt.Fprintf(os.Stderr,
+				"%s rebalance: %d partition(s) revoked: %v\n",
+				c.GetRebalanceProtocol(), len(e.Partitions),
+				e.Partitions)
+			if c.AssignmentLost() {
+				// Our consumer has been kicked out of the group and the
+				// entire assignment is thus lost.
+				fmt.Fprintf(os.Stderr, "Current assignment lost!\n")
+			}
+			run = false
+
+		case AssignedPartitions:
+			fmt.Fprintf(os.Stderr,
+				"%s rebalance: %d new partition(s) assigned: %v\n",
+				c.GetRebalanceProtocol(), len(e.Partitions),
+				e.Partitions)
+
+		case Error:
+			// Errors should generally be
+			// considered informational, the client
+			// will try to automatically recover.
+			fmt.Printf("Error: %v: %v for "+
+				"consumer %v\n", e.Code(), e, c)
+
+		default:
+			fmt.Printf("Ignored %v for consumer %v\n",
+				e, c)
+		}
+	}
+	c.Close()
+	fmt.Printf("Consumer closed\n")
+}
+
+func rebalanceCallbackCooperative(c *Consumer, event Event) error {
+
+	switch ev := event.(type) {
+	case AssignedPartitions:
+		fmt.Fprintf(os.Stderr,
+			"%s rebalance: %d new partition(s) assigned: %v\n",
+			c.GetRebalanceProtocol(), len(ev.Partitions),
+			ev.Partitions)
+
+		// The application may update the start .Offset of each
+		// assigned partition and then call IncrementalAssign().
+		// Even though this example does not alter the offsets we
+		// provide the call to IncrementalAssign() as an example.
+		err := c.IncrementalAssign(ev.Partitions)
+		if err != nil {
+			panic(err)
+		}
+
+	case RevokedPartitions:
+		fmt.Fprintf(os.Stderr,
+			"%s rebalance: %d partition(s) revoked: %v\n",
+			c.GetRebalanceProtocol(), len(ev.Partitions),
+			ev.Partitions)
+		if c.AssignmentLost() {
+			// Our consumer has been kicked out of the group and the
+			// entire assignment is thus lost.
+			fmt.Fprintf(os.Stderr, "%% Current assignment lost!\n")
+		}
+
+		// The client automatically calls IncrementalUnassign() unless
+		// the callback has already called that method.
+	}
+
+	return nil
+}
+
+func rebalanceCallbackEager(c *Consumer, event Event) error {
+
+	switch ev := event.(type) {
+	case AssignedPartitions:
+		fmt.Fprintf(os.Stderr,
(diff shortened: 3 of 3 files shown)


kafka/consumer_test.go
```diff
@@ -419,3 +419,166 @@ func TestConsumerLog(t *testing.T) {
...
+	}
+	c.Close()
+	fmt.Printf("Consumer closed\n")
+}
+
+func rebalanceCallbackCooperative(c *Consumer, event Event) error {
+
+	switch ev := event.(type) {
+	case AssignedPartitions:
+		fmt.Fprintf(os.Stderr,
```
edenhill, line 518: and here and below (assert)

edenhill (changes requested):

kafka/consumer_test.go
```diff
@@ -419,3 +419,166 @@ func TestConsumerLog(t *testing.T) {
...
+		case *Message:
+			fmt.Printf("Message on %s:\n%s\n",
+				e.TopicPartition, string(e.Value))
+			if e.Headers != nil {
+				fmt.Printf("Headers: %v\n", e.Headers)
+			}
+			run = false
+
+		case RevokedPartitions:
+			fmt.Fprintf(os.Stderr,
```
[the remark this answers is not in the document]
jliunyu: Hi Magnus, thanks for the reviewing!

From the description in https://github.com/confluentinc/confluent-kafka-go/issues/610 (below), my understanding is kafka.AssignedPartitions and kafka.RevokedPartitions should be returned from both Poll() and callback functions. So I added tests for both poll() and callback regression tests. If the expected behavior is that kafka.AssignedPartitions and kafka.RevokedPartitions are only be seen if callback is not set, I will change more on the logic.

```
The behaviour of "go.application.rebalance.enable": true appears to have changed with v1.6.1, with neither kafka.AssignedPartitions or kafka.RevokedPartitions being returned by Poll().

The only way to capture these events is separately with a rebalanceCb method when calling Subscribe().

```

kevinpark1217:
Is there an approximate date/timeline on where this fix will be landed and rolled out? possibly to `1.6.2`?

kafka/consumer_test.go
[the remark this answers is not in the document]
edenhill: If a rebalance callback has been set then Poll() should not return these events.

Here's the pre-1.6 logic for reference:
https://github.com/confluentinc/confluent-kafka-go/blob/v1.5.2/kafka/event.go#L180-L227

edenhill:
This will be released in v1.8.0, scheduled for august.

kafka/consumer_test.go
jliunyu: This and below are under callback functions, so I didn't add assert here

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Confluent's Golang client for Apache Kafka
 
+## v1.8.0
+
+v1.8.0 is a maintenance release with the following fixes and enhancements:
+
+### Fixes
+
+* Fix Rebalance events behavior when config go.application.rebalance.enable to true.
```
edenhill, line 9: I think it needs to be a bit more specific what the fix is.

kafka/consumer.go
```diff
@@ -905,5 +905,5 @@ func (c *Consumer) handleRebalanceEvent(channel chan Event, rkev *C.rd_kafka_eve
 		c.events <- newError(cErr)
 	}
 
-	return nil
+	return ev
```
edenhill, line 908: I'm not sure it should return the event here, there is a code path that leads here even though a rebalanceCb is set, namely when the rebalanceCb did not make the assignment and the c.appReassigned was never set.
Since a rebalance callback is used poll should not return an event in this case either.
This would've been caught by the new tests if the rebalance callback did not call Assign(), but it does so for the AssignedPartitions case, but not RevokedPartitions. Unfortunately the test only triggers the assignment, not the revokal (I believe that's hidden in Close()).

So, I recommend to update the test to force a revoke after an assignment has been seen, easiest way is to call Unsubscribe(). That should fail the test with this current code.

kafka/consumer_test.go
```diff
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
```
edenhill, line 28: Let's try to avoid extra dependencies unless strictly needed.

edenhill (changes requested):

kevinpark1217:
@edenhill Is it possible to backport this fix to the `1.6.x` versions? That would be very helpful from the Robinhood's perspective

kafka/consumer.go
jliunyu: Hi @edenhill, thanks for reviewing. 

Close() is added at the test case, https://github.com/confluentinc/confluent-kafka-go/pull/645/files#diff-a9cef34b4369fffc694b2b116cb24394eb75fb0e5d67568a2114140b06c1652eR511, so revoke is triggered when the Close is called.
```
Local Test:
jliu@C02DM32YMD6T kafka % go test -run TestConsumerRebalanceEvents
Consumer closed
Consumer closed
COOPERATIVE rebalance: 4 new partition(s) assigned: [testConsumerRebalanceEvents[0]@unset testConsumerRebalanceEvents[1]@unset testConsumerRebalanceEvents[2]@unset testConsumerRebalanceEvents[3]@unset]
COOPERATIVE rebalance: 4 partition(s) revoked: [testConsumerRebalanceEvents[0]@unset testConsumerRebalanceEvents[1]@unset testConsumerRebalanceEvents[2]@unset testConsumerRebalanceEvents[3]@unset]
Consumer closed
EAGER rebalance: 4 new partition(s) assigned: [testConsumerRebalanceEvents[0]@unset testConsumerRebalanceEvents[1]@unset testConsumerRebalanceEvents[2]@unset testConsumerRebalanceEvents[3]@unset]
EAGER rebalance: 4 partition(s) revoked: [testConsumerRebalanceEvents[0]@unset testConsumerRebalanceEvents[1]@unset testConsumerRebalanceEvents[2]@unset testConsumerRebalanceEvents[3]@unset]
Consumer closed
PASS
ok  	_/Users/jliu/Documents/Code/jliunyu/confluent-kafka-go/kafka	0.198s
```

But if Call Unsubscribe() then call Close(), will see this error ```%3|1626911382.640|REBALANCE|rdkafka#consumer-3| [thrd:main]: Group "gotest": unexpected instruction to revoke current assignment and rebalance (terminating=0, assignment_lost=0, LEAVE_ON_UNASSIGN_DONE=0)```

Can you please help to conform the expected behavior? since both Unsubscribe() and Close() will trigger revoke, so Unsubscribe() will trigger the revoke if it's been called, otherwise close() will trigger it?

jliunyu:
> > This will be released in v1.8.0, scheduled for august.
> 
> @edenhill Is it possible to backport this fix to the `1.6.x` versions? And hopefully soon? That would be very helpful from the Robinhood's side

Hi @kevinpark1217, thanks for asking, since we didn't release bug fix in this way before, we are discussing about this and update soon.

jliunyu:
@kevinpark1217 , sorry for the late response, we will update to you early next week

jliunyu:
@kevinpark1217, after discussing with the team, we never backport any fix before. I'm really sorry for the inconvenience.  Please feel free to let us know if anything else we can help.

edenhill:
Can you force push to this branch to re-trigger CI?

johnseekins:
So if I understand this correctly, `go.application.rebalance.enable` is currently broken from >= 1.6? How do we handle rebalance requests then?

scanterog:
Is there an approximate timeline on where this fix will be landed and rolled out?

seruman:
Hi there! Thanks you all for the fix. We applied *only this change* to our own fork of 1.7.0 and 1.8.2 and observed an increase in memory usage;
* 1.7.0: https://github.com/peak/confluent-kafka-go/commit/3576112dc12415f5d0824143df5745f329df9dd0
* 1.8.2: https://github.com/peak/confluent-kafka-go/commit/4905de20f067439b58c184e1be5d91f3685fccb8

Used application is a consumer that batches messages up to 10s and ingest them in to destination;
* Runs on Kubernetes with 6 replicas.
* Subscribed to 4 topics which all are 12 partitions.
* Total message rate of topics is between 18.2K to 34K

Rebalance events are only used to log subscibed/revoked topic-partitions and clean up internal topic-partition based metrics.

**`confluentinc/confluent-kafka-go@v1.8.2`**;
* Total average message/sec in topics: Between 18K-34K

Pods' memory usage starts around ~500MiB/700MiB and increases up to ~1GiB with the increase in message rate.
ckafka-1 8 2

Color changes in the graph where some of the pods are rescheduled by Kubernetes to be moved to another worker node which triggered rebalances.


**`peak/confluent-kafka-go@v1.7.0-fix-rebalance`**;
* Total average message/sec in topics: Between 20K-25K

This time memory usage starts around ~800MiB/2GiB and it looked like the first starting consumer was using the highest memory.
peak-1 7 0

**`peak/confluent-kafka-go@v1.8.2-fix-rebalance`**;
* Total average message/sec in topics: Between 23K-25K

Memory usage gradually increases up to ~1GiB/~2GiB between pods. After manually tringgering a rebalance -marked by dotted lines- by killing one of the pods, memory usage of the remaning pods increased ~1GiB.
peak-1 8 2


There was not any changes to the application code between these runs besides Kafka client versions. Even with the higher message rate, the version without the fix was using the least memory.

jliunyu:
> Hi there! Thanks you all for the fix. We applied _only this change_ to our own fork of 1.7.0 and 1.8.2 and observed an increase in memory usage;
> 
> * 1.7.0: [peak@3576112](https://github.com/peak/confluent-kafka-go/commit/3576112dc12415f5d0824143df5745f329df9dd0)
> * 1.8.2: [peak@4905de2](https://github.com/peak/confluent-kafka-go/commit/4905de20f067439b58c184e1be5d91f3685fccb8)
> 
> Used application is a consumer that batches messages up to 10s and ingest them in to destination;
> 
> * Runs on Kubernetes with 6 replicas.
> * Subscribed to 4 topics which all are 12 partitions.
> * Total message rate of topics is between 18.2K to 34K
> 
> Rebalance events are only used to log subscibed/revoked topic-partitions and clean up internal topic-partition based metrics.
> 
> **`confluentinc/confluent-kafka-go@v1.8.2`**;
> 
> * Total average message/sec in topics: Between 18K-34K
> 
> Pods' memory usage starts around ~500MiB/700MiB and increases up to ~1GiB with the increase in message rate. ckafka-1 8 2
> 
> Color changes in the graph where some of the pods are rescheduled by Kubernetes to be moved to another worker node which triggered rebalances.
> 
> **`peak/confluent-kafka-go@v1.7.0-fix-rebalance`**;
> 
> * Total average message/sec in topics: Between 20K-25K
> 
> This time memory usage starts around ~800MiB/2GiB and it looked like the first starting consumer was using the highest memory. peak-1 7 0
> 
> **`peak/confluent-kafka-go@v1.8.2-fix-rebalance`**;
> 
> * Total average message/sec in topics: Between 23K-25K
> 
> Memory usage gradually increases up to ~1GiB/~2GiB between pods. After manually tringgering a rebalance -marked by dotted lines- by killing one of the pods, memory usage of the remaning pods increased ~1GiB. peak-1 8 2
> 
> There was not any changes to the application code between these runs besides Kafka client versions. Even with the higher message rate, the version without the fix was using the least memory.

This PR is discarded, If you don't use channel based consumer, please use the rebalance callback to handle the rebalance, it will be called from the Poll() function: https://github.com/confluentinc/confluent-kafka-go/issues/610#issuecomment-1076732662

PR #663, opened 2021-07-15
Correct InitTransactions() method comment

pmagix:
Remove the following comment lines, since these only apply once BeginTransaction() has been called:
// Upon successful return from this function the application has to perform at
// least one of the following operations within `transaction.timeout.ms` to
// avoid timing out the transaction on the broker:
//  * `Produce()` (et.al)
//  * `SendOffsetsToTransaction()`
//  * `CommitTransaction()`
//  * `AbortTransaction()`

### DIFF
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -722,14 +722,6 @@ func (p *Producer) SetOAuthBearerTokenFailure(errstr string) error {
 // will acquire the internal producer id and epoch, used in all future
 // transactional messages issued by this producer instance.
 //
-// Upon successful return from this function the application has to perform at
-// least one of the following operations within `transaction.timeout.ms` to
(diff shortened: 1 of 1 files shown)


pmagix:
[clabot:check]

kafka/producer.go
```diff
@@ -722,14 +722,6 @@ func (p *Producer) SetOAuthBearerTokenFailure(errstr string) error {
 // will acquire the internal producer id and epoch, used in all future
 // transactional messages issued by this producer instance.
 //
-// Upon successful return from this function the application has to perform at
```
edenhill, line 725: I think this should be moved down to BeginTransaction

edenhill (changes requested):

PR #676, opened 2021-08-17
Added convenience method StoreMessage that calls StoreOffsets

finncolman:
If you have "enable.auto.commit" set to true, but "enable.auto.offset.store" set to false then you need to call  StoreOffsets after the consumer processing has completed, but at this stage you only have the message that was fetched, not the offsets.
This means that you end up writing code like this all over the place:
`if m.TopicPartition.Error != nil {
		return nil, newErrorFromString(ErrInvalidArg, "Can't store errored message")
	}
	offsets := []TopicPartition{m.TopicPartition}
	offsets[0].Offset++
	return c.StoreOffsets(offsets)`

This PR adds a convenience method StoreMessage that does this for you. This is similar to the CommitMessage method in that it simply gets the offset to store from the message.

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -292,6 +292,17 @@ func (c *Consumer) StoreOffsets(offsets []TopicPartition) (storedOffsets []Topic
 	return storedOffsets, nil
 }
 
+// StoreMessage stores offset based on the provided message.
+// This is a convenience method that uses StoreOffsets to do the actual work
+func (c *Consumer) StoreMessage(m *Message) (storedOffsets []TopicPartition, err error) {
+	if m.TopicPartition.Error != nil {
+		return nil, newErrorFromString(ErrInvalidArg, "Can't store errored message")
+	}
+	offsets := []TopicPartition{m.TopicPartition}
+	offsets[0].Offset++
+	return c.StoreOffsets(offsets)
+}
+
 // Seek seeks the given topic partitions using the offset from the TopicPartition.
 //
 // If timeoutMs is not 0 the call will wait this long for the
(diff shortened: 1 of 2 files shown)


kafka/consumer.go
```diff
@@ -292,6 +292,17 @@ func (c *Consumer) StoreOffsets(offsets []TopicPartition) (storedOffsets []Topic
 	return storedOffsets, nil
 }
 
+// StoreMessage stores offset based on the provided message.
+// This is a convenience method that uses StoreOffsets to do the actual work
```
edenhill, line 296: Missing "." at end of line.

edenhill (changes requested):
Great stuff!
One small nit, maybe the smallest of them all.

A future performance improvement is to make a cgo implementation of StoreOffsets to cut it down to a single C call (instead of something like 4), since each C call has a large overhead in Go.

kafka/consumer.go
```diff
@@ -292,6 +292,17 @@ func (c *Consumer) StoreOffsets(offsets []TopicPartition) (storedOffsets []Topic
 	return storedOffsets, nil
 }
 
+// StoreMessage stores offset based on the provided message.
+// This is a convenience method that uses StoreOffsets to do the actual work.
+func (c *Consumer) StoreMessage(m *Message) (storedOffsets []TopicPartition, err error) {
+	if m.TopicPartition.Error != nil {
```
edenhill, line 298: Maybe we should also make sure Offset is >= 0. You never know..

edenhill (changes requested):

PR #697, opened 2021-10-18
Producer: fixed events to be sent even when no error

eko:
Hi,

It looks like actually, message events are sent only when an error occurred on the broker.

This PR simply allows to send them in call cases (which I think should be the case).

Could you please confirm?

Thank you


eko:
Sorry, I misunderstood something, it seems to be working properly :)

PR #703, opened 2021-10-27
Add example for consumer manual commit batch process

Bohatman:
Add some example for batch processing usage with required control commit
Example case:
 Kafka to database program insert batch to database 500 records then commit database transaction  If insert database working fine then commit offset

I think this example might be useful

### DIFF
--- a/examples/consumer_manual_batch_commit_example/consumer_manual_batch_commit_example.go
+++ b/examples/consumer_manual_batch_commit_example/consumer_manual_batch_commit_example.go
@@ -1,0 +1,108 @@
+package main
+
+import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


examples/consumer_manual_batch_commit_example/consumer_manual_batch_commit_example.go
```diff
@@ -0,0 +1,108 @@
...
+					doCommit(c)
+					controller = 0
+				}
+				continue
+			}
+
+			switch e := ev.(type) {
+			case *kafka.Message:
+				controller++
+				if controller == batchSize {
```
fengkai, line 68: What if controller always < batchSize? I think this should have a timeout

PR #732, opened 2022-01-22
Using a buffered channel on Producer example code

robsonpeixoto:
Using a buffered channel on producer code sample just to make clear that is safe use it.

### DIFF
--- a/examples/producer_example/producer_example.go
+++ b/examples/producer_example/producer_example.go
@@ -53,6 +53,9 @@ func main() {
 		Value:          []byte(value),
 		Headers:        []kafka.Header{{Key: "myTestHeader", Value: []byte("header values are binary")}},
 	}, deliveryChan)
+	if err != nil {
+		fmt.Printf("Failed to produce the message: %s\n", err.Error())
+	}
 
 	e := <-deliveryChan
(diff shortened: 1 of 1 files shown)


edenhill:
[clabot:check]

examples/producer_example/producer_example.go
```diff
@@ -45,14 +45,17 @@ func main() {
...
+	deliveryChan := make(chan kafka.Event, 1)
 
 	value := "Hello Go!"
 	err = p.Produce(&kafka.Message{
 		TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny},
 		Value:          []byte(value),
 		Headers:        []kafka.Header{{Key: "myTestHeader", Value: []byte("header values are binary")}},
 	}, deliveryChan)
+	if err != nil {
+		fmt.Printf("Failed to produce the message: %s\n", err.Error())
```
edenhill, line 57: This should probably exit/return as well since the program will hang waiting on the deliveryChan if no message was produced.

robsonpeixoto: Done in 116c9559ba54e9381b7e6f5a5a64bf6632396208

PR #757, opened 2022-04-06
Fix Rebalance events behavior for static membership and fix for consu…

jliunyu:
…mer close takes longer than 10s
```
EAGER rebalance: 6 partition(s) revoked: [testrebalance[0]@unset testrebalance[1]@unset testrebalance[2]@unset testrebalance[3]@unset testrebalance[4]@unset testrebalance[5]@unset]
Ignored OffsetsCommitted (<nil>, [testrebalance[0]@953 testrebalance[1]@900 testrebalance[2]@919 testrebalance[3]@unset testrebalance[4]@857 testrebalance[5]@925]) for consumer rdkafka#consumer-1
rdkafka#consumer-1, EAGER rebalance: 3 new partition(s) assigned: [testrebalance[0]@unset testrebalance[1]@unset testrebalance[2]@unset]
rdkafka#consumer-2, EAGER rebalance: 3 new partition(s) assigned: [testrebalance[3]@unset testrebalance[4]@unset testrebalance[5]@unset]
Message on testrebalance[1]@900:
test
EAGER rebalance: 3 partition(s) revoked: [testrebalance[0]@unset testrebalance[1]@unset testrebalance[2]@unset]
Ignored OffsetsCommitted (<nil>, [testrebalance[0]@unset testrebalance[1]@901 testrebalance[2]@unset]) for consumer rdkafka#consumer-1
EAGER rebalance: 3 partition(s) revoked: [testrebalance[3]@unset testrebalance[4]@unset testrebalance[5]@unset]
rdkafka#consumer-2, EAGER rebalance: 2 new partition(s) assigned: [testrebalance[2]@unset testrebalance[3]@unset]
rdkafka#consumer-1, EAGER rebalance: 2 new partition(s) assigned: [testrebalance[0]@unset testrebalance[1]@unset]
rdkafka#consumer-3, EAGER rebalance: 2 new partition(s) assigned: [testrebalance[4]@unset testrebalance[5]@unset]
rdkafka#consumer-4, EAGER rebalance: 2 new partition(s) assigned: [testrebalance[2]@unset testrebalance[3]@unset]
PASS
ok  	_/Users/jliu/Documents/Code/jliunyu/confluent-kafka-go/kafka	38.019s
```

Manually tested for the `Close`, before this fix, the consumer takes more than 10s, with this fix, closing time: 730.602µs

```
consumer closed. closing time: 730.602µs
PASS
ok  	_/Users/jliu/Documents/Code/jliunyu/confluent-kafka-go/kafka	0.573s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ This is a feature release:
    See [examples/mock_cluster](examples/mock_cluster).
 
 
+### Fixes
+
+ * Fix Rebalance events behavior for static membership (@jliunyu, #757).
+ * Fix consumer close taking 10 seconds when there's no rebalance
+   needed (@jliunyu, #757).
+
 confluent-kafka-go is based on librdkafka FUTUREFIXME, see the
 [librdkafka release notes](https://github.com/edenhill/librdkafka/releases/tag/v1.9.0)
 for a complete list of changes, enhancements, fixes and upgrade considerations.
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -424,29 +424,32 @@ func (c *Consumer) Close() (err error) {
 		close(c.events)
 	}
 
-	// librdkafka's rd_kafka_consumer_close() will block
-	// and trigger the rebalance_cb() if one is set, if not, which is the
-	// case with the Go client since it registers EVENTs rather than callbacks,
-	// librdkafka will shortcut the rebalance_cb and do a forced unassign.
-	// But we can't have that since the application might need the final RevokePartitions
-	// before shutting down. So we trigger an Unsubscribe() first, wait for that to
-	// propagate (in the Poll loop below), and then close the consumer.
-	c.Unsubscribe()
+	doneChan := make(chan bool)
 
-	// Poll for rebalance events
-	for {
-		c.Poll(10 * 1000)
-		if int(C.rd_kafka_queue_length(c.handle.rkq)) == 0 {
-			break
+	go func() {
+		C.rd_kafka_consumer_close(c.handle.rk)
+		// wake up Poll()
+		C.rd_kafka_queue_yield(c.handle.rkq)
+		doneChan <- true
+	}()
+
+	// wait for consumer_close() to finish while serving c.Poll() for rebalance callbacks/events
+	run := true
+	for run {
+		select {
+		case <-doneChan:
+			run = false
+
+		default:
+			c.Poll(100)
 		}
 	}
+
+	close(doneChan)
 
 	// Destroy our queue
 	C.rd_kafka_queue_destroy(c.handle.rkq)
 	c.handle.rkq = nil
-
-	// Close the consumer
-	C.rd_kafka_consumer_close(c.handle.rk)
 
 	c.handle.cleanup()
 
--- a/kafka/consumer_test.go
+++ b/kafka/consumer_test.go
@@ -22,6 +22,8 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 )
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Confluent's Golang client for Apache Kafka
 
+## v1.9.0
+
+This is a maintenance release:
+
+ * Fix Rebalance events behavior for static membership and fix for consumer close takes longer than 10s (@jliunyu, #757).
```
edenhill, line 7: Split this up into two entries, since they're two issues.
Rephrase the second one to:
" * Fix consumer close taking 10 seconds when there's no rebalance needed"

kafka/consumer.go
```diff
@@ -424,30 +424,30 @@ func (c *Consumer) Close() (err error) {
...
+		C.rd_kafka_queue_yield(c.handle.rkq)
+		doneChan <- true
+	}()
+
+	// wait for consumer_close() to finish while serving c.Poll() for rebalance callbacks/events
+	run := true
+	for run {
+		c.Poll(100)
+		select {
+		case <-doneChan:
```
edenhill, line 440: I think this will block until doneChan is written to, which means if Poll() is triggered once for some other reason, e.g., a stats callback, then it will hang here.

kafka/consumer_test.go
```diff
@@ -429,3 +430,223 @@ func TestConsumerLog(t *testing.T) {
...
+
+func rebalanceCallback(c *Consumer, event Event) error {
+	switch ev := event.(type) {
+	case AssignedPartitions:
+		atomic.AddInt32(&assignedEvents, 1)
+		fmt.Fprintf(os.Stderr,
+			"%v, %s rebalance: %d new partition(s) assigned: %v\n",
+			c, c.GetRebalanceProtocol(), len(ev.Partitions),
+			ev.Partitions)
+		// The application may update the start .Offset of each
```
edenhill, line 445: nit: no need for this comment in the test code.

kafka/consumer_test.go
```diff
@@ -429,3 +430,223 @@ func TestConsumerLog(t *testing.T) {
...
 }
+
+var assignedEvents int32
+var revokedEvents int32
+
+func rebalanceCallback(c *Consumer, event Event) error {
+	switch ev := event.(type) {
+	case AssignedPartitions:
+		atomic.AddInt32(&assignedEvents, 1)
+		fmt.Fprintf(os.Stderr,
```
edenhill, line 441: nit: Tests should use t.Logf if possible, but that might be tricky here in the callback.

kafka/consumer_test.go
```diff
@@ -429,3 +430,223 @@ func TestConsumerLog(t *testing.T) {
...
+// According to KIP-345, the consumer group will not trigger rebalance unless
+// 1) A new member joins
+// 2) A leader rejoins (possibly due to topic assignment change)
+// 3) An existing member offline time is over session timeout
+// 4) Broker receives a leave group request containing alistof
+//    `group.instance.id`s (details later)
+//
+// This test uses 3 consumers while each consumer joins after the assignment
+// finished for the previous consuemrs.
+// The expected behavior for these consuemrs are:
```
edenhill, line 519: typo

kafka/consumer_test.go
```diff
@@ -429,3 +430,223 @@ func TestConsumerLog(t *testing.T) {
...
+
+			default:
+				fmt.Printf("Ignored %v for consumer %v\n",
+					e, c)
+			}
+		}
+	}
+}
+
+// TestConsumerCloseForStaticMember verifies the rebalance
```
edenhill, line 508: Very good test doc!

kafka/consumer_test.go
```diff
@@ -429,3 +430,223 @@ func TestConsumerLog(t *testing.T) {
...
+// 2) Second consumer joins, RevokedPartitions happens from the first consumer,
+//    then AssignedPartitions happens to both consumers.
+// 3) Third consumer joins, RevokedPartitions happens from the previous two
+//    consumers, then AssignedPartitions happens to all the three consumers.
+// 4) Close the second consumer, RevokedPartitions should not happen.
+// 5) Rejoin the second consumer, rebalance should not happen to all the other
+//    consumers since it's not the leader, AssignedPartitions only happened
+//    to this consumer to assign the partitions.
+//
+// The total number of AssignedPartitions is 7 and the total number of
```
edenhill, line 531: This should preferably be tracked per consumer so that we know that it was Consumer 2 that had the rebalance and not  the others.

kafka/consumer_test.go
```diff
@@ -429,3 +430,223 @@ func TestConsumerLog(t *testing.T) {
...
+	topic := testconf.Topic
+
+	conf := ConfigMap{
+		"bootstrap.servers":    broker,
+		"group.id":             "rebalance",
+		"session.timeout.ms":   "6000",
+		"max.poll.interval.ms": "10000",
+		"group.instance.id":    "staticmember",
+	}
+	c, err := NewConsumer(&conf)
```
edenhill, line 547: nit: I think it would be clearer if the first consumer was called c1 and staticmember1, the second c2 and staticmember2, etc.

kafka/consumer_test.go
```diff
@@ -429,3 +430,223 @@ func TestConsumerLog(t *testing.T) {
...
+func testConsumerWaitAssignment(c *Consumer, t *testing.T) {
+	run := true
+	for run {
+		assignment, err := c.Assignment()
+		if err != nil {
+			t.Fatalf("Assignment (after incremental) failed: %s", err)
+		}
+
+		if len(assignment) != 0 {
+			t.Logf("%v Assigned\n", c)
```
edenhill, line 648: Perhaps print the assignment here for troubleshooting help.

kafka/consumer_test.go
```diff
@@ -429,3 +430,223 @@ func TestConsumerLog(t *testing.T) {
...
+			atomic.LoadInt32(&revokedEvents))
+	}
+}
+
+func testConsumerWaitAssignment(c *Consumer, t *testing.T) {
+	run := true
+	for run {
+		assignment, err := c.Assignment()
+		if err != nil {
+			t.Fatalf("Assignment (after incremental) failed: %s", err)
```
edenhill, line 644: remove "(after incremental)"?

kafka/consumer_test.go
```diff
@@ -429,3 +430,223 @@ func TestConsumerLog(t *testing.T) {
...
+// 3) Third consumer joins, RevokedPartitions happens from the previous two
+//    consumers, then AssignedPartitions happens to all the three consumers.
+// 4) Close the second consumer, RevokedPartitions should not happen.
+// 5) Rejoin the second consumer, rebalance should not happen to all the other
+//    consumers since it's not the leader, AssignedPartitions only happened
+//    to this consumer to assign the partitions.
+//
+// The total number of AssignedPartitions is 7 and the total number of
+// RevokedPartitions is 3.
+func TestConsumerCloseForStaticMember(t *testing.T) {
```
edenhill, line 533: Good test!

edenhill (changes requested):
Good stuff!

kafka/consumer.go
jliunyu: Can we use `int(C.rd_kafka_queue_length(c.handle.rkq)) == 0` to exist the loop which is same with the previous way?
```
for run {
		c.Poll(100)
		select {
		case <-doneChan:
			run = false

		default:
			if int(C.rd_kafka_queue_length(c.handle.rkq)) == 0 {
				run = false
			}
		}
	}
```

jliunyu: And I'm also confused why it will hang, I tested with this code with the existing change in this PR, it doesn't hang

```
func TestConsumerClose1(t *testing.T) {

	c, err := NewConsumer(&ConfigMap{
		"group.id":               "gotest"})
	if err != nil {
		t.Fatalf("%s", err)
	}

	startClose := time.Now()
	err = c.Close()
	if err != nil {
		panic(err)
	}
	fmt.Println(fmt.Sprintf("consumer closed. closing time: %v", time.Since(startClose)))
}
```

edenhill: I think with the following sequence of events it will hang:
 1. application calls Close()
 2. Close() is blocking on c.Poll()
 3. A stats event (or something else which is not a rebalance event) triggers the Poll and Poll() wakes up and returns
 4. Close() is now blocking on doneChan
 5. librdkafka emits a rebalance event and waits for the app/go to ack it
 6. .. but since Close() is no longer calling Poll() the rebalance event is never handled.
 7. indefinite hang

edenhill: I think this is easily solved by moving the Poll to the default case in the select, like in other places in the code.

kafka/consumer_test.go
jliunyu: Yeah, is there any good way to use `t *testing.T`  in the callback?

edenhill: I don't know, really. Go 🤷

kafka/consumer.go
jliunyu:
```
for run {
		select {
		case <-doneChan:
**-------------->// Is it possible that after the doneChan, there are still other events we need to handle? We always have to wait to read the doneChan value. If we exit from default, the doneChan will report some error**
			run = false

		default:
                         c.Poll(100)
			if int(C.rd_kafka_queue_length(c.handle.rkq)) == 0 {
				run = false
			}
		}
	}
```

edenhill: I don't think so, when doneChan is done we don't need to call Poll no more.

edenhill: Why is the queue_length() check needed?

jliunyu: We don't need it, I already removed it from my code.

kafka/consumer_test.go
```diff
@@ -429,3 +431,277 @@ func TestConsumerLog(t *testing.T) {
...
+		t.Fatalf("%s", err)
+	}
+	wrapRebalancecb1 := wrapRebalanceCb(&assignedEvents1, &revokedEvents1, t)
+	err = c1.Subscribe(topic, wrapRebalancecb1)
+	if err != nil {
+		t.Fatalf("Failed to subscribe to topic %s: %s\n", topic, err)
+		return
+	}
+
+	wg := &sync.WaitGroup{}
```
edenhill, line 597: I think it is better to keep this as a non-pointer here, and then pass it as pointer when needed (to testPoll()).

edenhill (changes requested):
Do all existing tests pass now?

jliunyu:
> Do all existing tests pass now?

Yes:)

```
jliu@C02DM32YMD6T kafka % go test -run "TestConsumer*" 
%4|1649999288.094|CONFWARN|rdkafka#consumer-1| [thrd:app]: Configuration property `fetch.wait.max.ms` (500) should be set lower than `socket.timeout.ms` (1000) by at least 1000ms to avoid blocking and timing out sub-sequent requests
rdkafka#consumer-6 Assigned partitions are: [testrebalance[0]@stored testrebalance[1]@stored testrebalance[2]@stored testrebalance[3]@stored testrebalance[4]@stored testrebalance[5]@stored]
rdkafka#consumer-7 Assigned partitions are: [testrebalance[3]@stored testrebalance[4]@stored testrebalance[5]@stored]
rdkafka#consumer-8 Assigned partitions are: [testrebalance[4]@stored testrebalance[5]@stored]
rdkafka#consumer-9 Assigned partitions are: [testrebalance[2]@stored testrebalance[3]@stored]
Static test passed
```

kafka/consumer_test.go
jliunyu: If we keep this as a non-pointer here, we need to use it as pointer each time, and there are a lot of use cases in the same function. 

For example:
wg.Add(1)
wg.Wait()

jliunyu:
This is the CI test: https://app.travis-ci.com/github/confluentinc/confluent-kafka-go/builds/249581975

kafka/consumer_test.go
sermojohn: The pointer receiver methods apply also to non pointer variables, based on [this](https://go.dev/tour/methods/6).

So pointer should be needed to pass the WaitGroup struct by reference, but not to call its methods with pointer receiver.

kafka/consumer_test.go
```diff
@@ -429,3 +431,277 @@ func TestConsumerLog(t *testing.T) {
...
+func testConsumerWaitAssignment(c *Consumer, t *testing.T) {
+	run := true
+	for run {
+		assignment, err := c.Assignment()
+		if err != nil {
+			t.Fatalf("Assignment failed: %s", err)
+		}
+
+		if len(assignment) != 0 {
+			fmt.Printf("%v Assigned partitions are: %v\n", c, assignment)
```
sermojohn, line 703: Did you consider using `t.Logf` in tests instead of `fmt.Printf `? This makes debugging logs better for debugging and easy to disable, by not using the `-v` flag.

kafka/consumer_test.go
```diff
@@ -429,3 +431,277 @@ func TestConsumerLog(t *testing.T) {
...
+			ev := c.Poll(100)
+			if ev == nil {
+				continue
+			}
+			switch e := ev.(type) {
+			case *Message:
+				t.Logf("Message on %s:\n%s\n",
+					e.TopicPartition, string(e.Value))
+				if e.Headers != nil {
+					fmt.Printf("Headers: %v\n", e.Headers)
```
sermojohn, line 487: [Comment](https://github.com/confluentinc/confluent-kafka-go/pull/757/files#r860762989) applies here too.

kafka/consumer_test.go
edenhill: > If we keep this as a non-pointer here, we need to use it as pointer each time, and there are a lot of use cases in the same function.

I still believe it is more correct to keep it as a non-pointer here and pass it as a pointer where needed.
But it is a nit, so do as you wish.

kafka/consumer_test.go
```diff
@@ -429,3 +431,277 @@ func TestConsumerLog(t *testing.T) {
...
+	c2, err = NewConsumer(&conf2)
+	if err != nil {
+		t.Fatalf("%s", err)
+	}
+
+	wrapRebalancecb4 := wrapRebalanceCb(&assignedEvents4, &revokedEvents4, t)
+	err = c2.Subscribe(topic, wrapRebalancecb4)
+	if err != nil {
+		t.Fatalf("Failed to subscribe to topic %s: %s\n", topic, err)
+		return
```
edenhill, line 638: not reached, thus not needed

kafka/consumer_test.go
```diff
@@ -429,3 +431,277 @@ func TestConsumerLog(t *testing.T) {
...
+
+	c2, err = NewConsumer(&conf2)
+	if err != nil {
+		t.Fatalf("%s", err)
+	}
+
+	wrapRebalancecb4 := wrapRebalanceCb(&assignedEvents4, &revokedEvents4, t)
+	err = c2.Subscribe(topic, wrapRebalancecb4)
+	if err != nil {
+		t.Fatalf("Failed to subscribe to topic %s: %s\n", topic, err)
```
edenhill, line 637: Inconsistent use of newline in Fatalf(). Not sure it needs a newline.

kafka/consumer_test.go
```diff
@@ -429,3 +431,277 @@ func TestConsumerLog(t *testing.T) {
...
+	doneChan <- true
+	close(doneChan)
+
+	c3.Close()
+	c2.Close()
+	c1.Close()
+
+	wg.Wait()
+
+	// Wait 30s to make sure no revokedEvents happens
```
edenhill, line 653: 30s is a long time for a test, can't we rely on some multiple of the session.timeout.ms to know how long we need to wait? E.g., 2*session.timeout.ms (2*6)?

edenhill (changes requested):
Still some unaddressed comments in the test file.

kafka/consumer_test.go
jliunyu: newline is not needed, but other locations in this file contains newline. Do you think we should remove them?

edenhill (approved):
LGTM!

sermojohn:
Is there a plan to make a release with this fix?

jliunyu:
> Is there a plan to make a release with this fix?

Yes, this one will be released soon.

PR #231, opened 2018-09-24
WIP: Avro/Schema-Registry 

rnpridgeon:
This is a request for a review/comment of a WIP Schema Registry implementation. 

I have attempted to stick to the Java API as much as possible with some minor changes but I've run into a bit of a mental block on deserialization. I'm not sure how to best handle deserialization into specific types.

I have toyed with a few implementations but I think they too are less than ideal 

1. take a new()interface{} function pointer upon instantiating the deserializer. This is similar to the way sync.Pool works

2. Move Avro serialization/deserialization out of the Producer/Consumer opting instead to have a stateful marshaler/unmarshaler. 

3. The current implementation which takes an instance on instantiation and makes a zero copy then  places the results in the Message.Object field which was added in the serde_support development branch to facilitate handling objects for serialization/deserialization. 

In the end I suspect the best way to move forward is option 2. This puts serialization/deserialization outside of the producer/consumer however which may not be what we want. 

All relevant code is located in the encoding/avro folder. Specifically serializer.go

PR #776, opened 2022-05-04
Add Schema Registry support

rayokota:
This PR adds support for Schema Registry to the golang client.  Much of the low-level SR client code has been adapted from https://github.com/confluentinc/confluent-kafka-go/pull/231.   There is a new Mock SR client that has been added.  All the serdes are new.  The following serdes are supported:

- generic Avro serde
- specific Avro serde that uses https://github.com/actgardner/gogen-avro
- JSON Schema serde
- Protobuf serde

Both unit tests and runnable examples have been added.  All the runnable examples have been tested locally.


examples/avro_generic_consumer_example/avro_generic_consumer_example.go
```diff
@@ -0,0 +1,128 @@
...
+		"auto.offset.reset":     "earliest"})
+
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to create consumer: %s\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Printf("Created Consumer %v\n", c)
+
+	deser := schemaregistry.GenericAvroDeserializer{}
```
mhowlett, line 66: is this way of creating and initializing the deserializers idiomatic?

it differs from the way consumers and producers are created, seems like we probably should have consistency.

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
```diff
@@ -0,0 +1,128 @@
...
+	err = deser.Configure(&kafka.ConfigMap{
```
mhowlett, line 67: First up, I don't have a clear view on how configuration should be done in a broader sense for this client. But here are some things that come to mind on that atm:

- `ConfigMap` has some very librdkafka specific stuff in it - it doesn't seem right to me that it is reused to configure the serdes.
- I don't reckon I even think this is a good design choice for the clients - i think a plain old map (or something else idiomatic that isn't librdkafka specific) would probably have been a better - encapsulate the librdkafka stuff further down. And a manifestation of this being not the best way to do it is that it complicates working out what's best now we have something else to configure. 
- I also don't like how a reference to a reference is passed, superfluous. But do we go for consistency or improvement?

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
```diff
@@ -0,0 +1,128 @@
...
+		"schema.registry.url": url,
```
mhowlett, line 68: in other clients we've taken the approach of not fully encapsulating the schema registry client - first create and configure that, then pass it to the serdes.

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
mhowlett: Also, i can't quite recall, but won't librdkafka/the go client complain if you give it a config it doesn't know about? which means you'd never create a single config and pass it to both.

Anyway, I think it would pay to think through how best to do config some more. Get a go fanatic to weigh in.

examples/protobuf_producer_example/protobuf_producer_example.go
```diff
@@ -0,0 +1,91 @@
...
+	fmt.Printf("Created serializer\n")
+
+	// Optional delivery channel, if not specified the Producer object's
+	// .Events channel is used.
+	deliveryChan := make(chan kafka.Event)
+
+	value := MyRecord{
+		ProductName: "Hello!",
+	}
+	payload, err := ser.Serialize(topic, &value)
```
mhowlett, line 68: Oh interesting - you've decoupled the serdes from the produce call. Not how it's done elsewhere, but I feel this approach is ok. The primary downside I can think of is that you don't get the original record in the context of the delivery report (just the serialized payload). The upside is it's much easier to implement.

examples/avro_generic_producer_example/avro_generic_producer_example.go
```diff
@@ -0,0 +1,96 @@
...
+	fmt.Printf("Created serializer\n")
+
+	// Optional delivery channel, if not specified the Producer object's
+	// .Events channel is used.
+	deliveryChan := make(chan kafka.Event)
+
+	value := MyRecord{
+		ProductName: "Hello!",
+	}
+	payload, err := ser.Serialize(topic, value)
```
mhowlett, line 68: Oh interesting - you've decoupled the serdes from the produce call. Not how it's done elsewhere, but I feel this approach is ok. The primary downside I can think of is that you don't get the original record in the context of the delivery report (just the serialized payload). The upside is it's much easier to implement.

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
```diff
@@ -0,0 +1,128 @@
...
+	}, false)
```
mhowlett, line 69: `false` doesn't read well. does go have enums?

examples/avro_generic_producer_example/avro_generic_producer_example.go
mhowlett: i guess not having generics forces you to do this.

some more disadvantages:
- Explicit serialization/deserialization is more key-strokes for the user.
- When you instantiate the serializer/deserializer the user needs to specify key or value. They then implicitly specify this again when passing the payload into the produce call, that's superfluous, seems not ideal.
- The serialize call is potentially blocking due to the HTTP call, whereas the produce call isn't (or shouldn't be IMO, i'll need to check how that's setup in the golang client case). I think you ideally want the entire produce operation to be async, and to achieve that the user would need to write some additional code. This may not matter in the goroutine paradigm though (as compared to async/await where you typically never want to block a thread).

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
```diff
@@ -0,0 +1,128 @@
...
+		default:
+			ev := c.Poll(100)
+			if ev == nil {
+				continue
+			}
+
+			switch e := ev.(type) {
+			case *kafka.Message:
+				value := MyRecord{}
+				err := deser.DeserializeInto(*e.TopicPartition.Topic, e.Value, &value)
```
mhowlett, line 96: is there a reason you can't return `deserialized_value, err` rather than use an out parameter? seems more idiomatic

schemaregistry/avro_generic.go
```diff
@@ -0,0 +1,112 @@
+package schemaregistry
```
mhowlett, line 1: is there any overhead associated with having avro/proto/json all in the one package from a dependency perspective?

mhowlett: if there is, i guess probably just compile time, and probably not significant?

schemaregistry/avro_generic.go
```diff
@@ -0,0 +1,112 @@
...
+	writer, name, err := s.toType(info)
+	if err != nil {
+		return nil, err
+	}
+	subject := s.subjectNameStrategy(topic, s.isKey, info)
+	msg, err := s.messageFactory(subject, name)
+	if err != nil {
+		return nil, err
+	}
+	_, err = avro.Unmarshal(payload[5:], msg, writer)
```
mhowlett, line 69: should check the magic byte and return error if not as expected.

schemaregistry/avro_generic.go
rayokota: Yes I don't believe there is

schemaregistry/avro_generic.go
mhowlett: ahh, that happens in getSchema, well done ...

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
```diff
@@ -0,0 +1,128 @@
...
+			run = false
+		default:
+			ev := c.Poll(100)
+			if ev == nil {
+				continue
+			}
+
+			switch e := ev.(type) {
+			case *kafka.Message:
+				value := MyRecord{}
```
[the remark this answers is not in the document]
rayokota: Yes the Avro library doesn't support that

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
rayokota: We need to know the type of the object to return

schemaregistry/json_schema.go
```diff
@@ -0,0 +1,155 @@
...
+
+import (
+	"encoding/json"
+	"github.com/invopop/jsonschema"
+	jsonschema2 "github.com/santhosh-tekuri/jsonschema/v5"
+	"io"
+	"strings"
+)
+
+// JSONSchemaSerializer represents either a JSON Schema serializer
```
mhowlett, line 11: ... or ?

schemaregistry/rest_service.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+type restService struct {
+	url     *url.URL
+	headers http.Header
+	*http.Client
+}
+
+// newRestService returns a new REST client for the Confluent Schema Registry
+func newRestService(conf *kafka.ConfigMap) (*restService, error) {
```
mhowlett, line 90: kafka config map making it all the way down here seems extra unfortunate..

schemaregistry/rest_service.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	transport, err := configureTransport(conf)
+	if err != nil {
+		return nil, err
+	}
+
+	timeout, err := conf.Get("request.timeout.ms", 10000)
```
mhowlett, line 118: is the intent to use the producer config of the same name? I think having a separate config value is more appropriate (with different name if they're intended to be in the same namespace).

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
rayokota: One advantage of abstracting SR client creation is that the type of client should be driven by the URL.  In both Java and here, if the URL starts with "mock://", we create a mock client.  In Java we also have clients that are of the form "local://" which are embedded in SR.  These are clients which talk to SR locally since they are instantiated on the server side.  Explicitly creating an HTTP client assumes that the URL always starts with http(s), and doesn't take advantage of the URL abstraction.

schemaregistry/serde.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	s.client = client
+	s.conf = conf
+	s.isKey = isKey
+	s.subjectNameStrategy = TopicNameStrategy
+	return nil
+}
+
+// SubjectNameStrategy determines the subject for the given parameters
+type SubjectNameStrategy func(topic string, isKey bool, schema SchemaInfo) string
```
mhowlett, line 73: The name of this type is the same as the name of the function that gets it. Maybe call this `SubjectNameStrategyFunc` or the method that gets it `GetSubjectNameStrategy`? (or something else more idiomatic?)

schemaregistry/serde.go
```diff
@@ -0,0 +1,241 @@
...
+	return s.subjectNameStrategy
+}
+
+// SetSubjectNameStrategy sets the subject naming strategy.
+func (s *serde) SetSubjectNameStrategy(strategy SubjectNameStrategy) {
+	s.subjectNameStrategy = strategy
+}
+
+// TopicNameStrategy creates a subject name by appending -[key|value] to the topic name.
+func TopicNameStrategy(topic string, isKey bool, schema SchemaInfo) string {
```
mhowlett, line 88: are you going to implement the other ones OOTB?

schemaregistry/serde.go
```diff
@@ -0,0 +1,241 @@
...
+		info := SchemaInfo{
+			Schema:     metadata.Schema,
+			SchemaType: metadata.SchemaType,
+			References: metadata.References,
+		}
+		_, err = resolveAvroReferences(c, info, ns)
+		if err != nil {
+			return nil, err
+		}
+
```
mhowlett, line 225: funny whitespace?

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
mhowlett: i don't mean explicitly create an HTTP client, i mean explicitly create a schema registry client defined in schemaregistry_client.go.

in .NET: https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/examples/Protobuf/Program.cs#L99

schemaregistry/serde.go
```diff
@@ -0,0 +1,241 @@
...
+// Deserializer represents a deserializer
+type Deserializer interface {
+	Configure(conf *kafka.ConfigMap, isKey bool) error
+	// Deserialize will call the MessageFactory to create an object
+	// into which we will unmarshal data.
+	Deserialize(topic string, payload []byte) (interface{}, error)
+	// DeserializeInto will unmarshal data into the given object.
+	DeserializeInto(topic string, payload []byte, msg interface{}) error
+	MessageFactory() MessageFactory
```
mhowlett, line 37: i feel like message factory would be better encapsulated, with type(s!?) specified on construction.

rayokota:
> How does this handle .proto definitions with imports that are `import "foo/bar/baz.proto"`? As far as I can tell, the resolution that happens here cannot handle that as the subjects are not path escaped.

@tquachbot , good point, I'll take a look.

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
mhowlett: This is for clarity.
In the other non-java clients we use an enum called `MessageComponent`.
I think it's particularly important in the go case, because you're not implementing a serialization framework yet, so this is user-facing.
It's also good, because it's more extensible.

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
rayokota: I've changed the code to use factory methods similar to NewConsumer/NewProducer

schemaregistry/serde.go
rayokota: Not at the moment, as we may want to deprecate the other strategies in the future

schemaregistry/serde.go
rayokota: Done

schemaregistry/serde.go
rayokota: The `MessageFactory` can dynamically construct the target type based on the subject and the name of the type

examples/protobuf_consumer_example/protobuf_consumer_example.go
```diff
@@ -0,0 +1,127 @@
...
+		fmt.Printf("Failed to create deserializer: %s\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Printf("Created deserializer\n")
+
+	// Register the Protobuf type so that Deserialize can be called.
+	// An alternative is to pass a pointer to an instance of the Protobuf type
+	// to the DeserializeInto method.
+	deser.ProtoRegistry.RegisterMessage((&MyRecord{}).ProtoReflect().Type())
```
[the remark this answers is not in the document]
rayokota: The `ProtoRegistry` is an object that Protobuf developers are familiar with, similar to Jackson's `ObjectMapper` in JSON.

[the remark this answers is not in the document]
rayokota: Both Deserialize and DeserializeInto are needed, as the former uses the `MessageFactory`, but the latter does not.

schemaregistry/serde/serde.go
```diff
@@ -0,0 +1,218 @@
...
+
+const (
+	// EnableValidation enables validation
+	EnableValidation = true
+	// DisableValidation disables validation
+	DisableValidation = false
+)
+
+// MagicByte is prepended to the serialized payload
+const MagicByte byte = 0x0
```
mhowlett, line 29: probably doesn't need to be public, i think probably make it private. if public, name could be something like `ConfluentFramingMagicByte`, or something better, don't know what.

schemaregistry/schemaregistry_client.go
```diff
@@ -0,0 +1,528 @@
...
+
+	return result, err
+}
+
+// DeleteSubjectVersion removes the version identified by delete from the subject's registration
+// Returns integer id for the deleted version
+func (c *client) DeleteSubjectVersion(subject string, delete int) (deleted int, err error) {
+	var result int
+	err = c.restService.handleRequest(newRequest("DELETE", versions, nil, url.PathEscape(subject), delete), &result)
+
```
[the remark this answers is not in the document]
rayokota: Fixed

schemaregistry/config.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"reflect"
+)
+
+// ConfigValue supports the following types:
+//  bool, int, string, any type with the standard String() interface
+type ConfigValue interface{}
+
+// ConfigMap is a map containing ConfigValues by key
+type ConfigMap map[string]ConfigValue
```
mhowlett, line 29: A couple of things I don't like about the way we do config with the go client:

1. it's not strongly typed. this adds a lot of value IMO (intellisense documenting + preventing errors at IDE time), can be autogenerated in the difficult (librdkafka config) case, and is something everyone using the client touches (and it's first impressions).
2. we pass in a reference to a map, but the values are allocated on the heap, so this is pointless.

With the disclaimer that I don't know golang, I'm thinking we might want something like:

```
type MyConfig struct {
        // In the client case, autogenerated from https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
	BootstrapServers string
	MessageMaxBytes  int
	MaxInFlight      int
}

func NewMyConfig(config map[string]string) MyConfig {
	var r MyConfig
	for key, val := range config {
		r.Set(key, val)
	}
	return r
}

func (c *MyConfig) Set(key string, val string) error {
	// In the client case, autogenerated from https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
	switch key {
	case "bootstrap.servers":
		c.BootstrapServers = val
	case "message.max.bytes":
		c.MessageMaxBytes = strconv.Atoi(val)
	case "max.in.flight":
		c.MaxInFlight = strconv.Atoi(val)
	default:
		return errors.New("unknown config")
	}
	return nil
}

func (c *MyConfig) Get(key string) {
	// blah
}
```

where there is a different config type for the each client type, schema registry, and each serde type.

finally, i think we don't need to pass a pointer to this (even though it's not clear to me in go if it's passed on the stack or heap - it's not critical path so doesn't matter, i think a cleaner api is better).

now is the time to figure out the story for config going forward (and game plan for updating the client types, or not)

schemaregistry/config.go
```diff
@@ -0,0 +1,66 @@
...
+	SslCaLocation                  string
+	SslDisableEndpointVerification bool
+
+	// Timeouts
+	ConnectionTimeoutMs int
+	RequestTimeoutMs    int
+}
+
+// NewConfig returns a new configuration instance with sane defaults.
+func NewConfig(url string) *Config {
```
mhowlett, line 45: I expect you considered having a parameterless constructor as well, and went with having url (the only thing required) as a parameter to minimize verbosity.

```
client, err := schemaregistry.NewClient(url)
```

vs

```
srConf := schemaregistry.NewConfig()
srConf.SchemaRegistryURL = url
client, err := schemaregistry.NewClient(srConf)
```

the argument against this is extensibility. right now there is only one required parameter, but what if in the future there is more? i realize it's still extensible in that you just set the other required parameter after you construct it, but then it's inconsistent. i also realize it's unlikely we'll ever have another required parameter. my argument is more theoretical / philosophical. i'm that kind of person. which is why i prefer going with the parameterless version. but i understand that other people see the world differently.

if go had method overloads or optional parameters, my sense here on what is best would probably be different.

schemaregistry/config.go
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Config is used to pass multiple configuration options to the Schema Registry client.
+type Config struct {
+	// SchemaRegistryURL determines the URL of Schema Registry
+	SchemaRegistryURL string
```
mhowlett, line 22: caps for URL does appear to be idiomatic: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

schemaregistry/serde/jsonschema/json_schema.go
```diff
@@ -0,0 +1,184 @@
...
+type Deserializer struct {
+	serde.BaseDeserializer
+	validate bool
+}
+
+var _ serde.Serializer = new(Serializer)
+var _ serde.Deserializer = new(Deserializer)
+
+// NewSerializer creates a JSON BaseSerializer for generic objects
+func NewSerializer(client schemaregistry.Client, conf *serde.Config, serdeType serde.Type, validate bool) (*Serializer, error) {
```
mhowlett, line 29: I think there should be a different config type for each of proto / avro / json. And there should be a different type for serializers and deserializers. No superfluous properties. Can derive from serde.Config for commonality.

mhowlett: even if not needed now, for extensibility.

schemaregistry/serde/config.go
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Config is used to pass multiple configuration options to the serdes.
+type Config struct {
+	// AutoRegisterSchemas determines whether to automatically register schemas during serialization
+	AutoRegisterSchemas bool
+	// UseSchemaID specifies a schema ID to use during serialization
+	UseSchemaID int
```
mhowlett, line 24: This reads like a boolean property name. I guess you are naming consistently with Java, so on balance it makes sense to leave. If not, maybe think of a different name.

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	deser, err := avro.NewGenericDeserializer(client, serde.NewConfig(), serde.ValueSerde)
+
+	if err != nil {
+		fmt.Printf("Failed to create deserializer: %s\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Printf("Created deserializer\n")
```
mhowlett, line 82: Maybe "Subscribing to ..." would be better.

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
```diff
@@ -0,0 +1,134 @@
...
+	sigchan := make(chan os.Signal, 1)
+	signal.Notify(sigchan, syscall.SIGINT, syscall.SIGTERM)
+
+	c, err := kafka.NewConsumer(&kafka.ConfigMap{
+		"bootstrap.servers": broker,
+		// Avoid connecting to IPv6 brokers:
+		// This is needed for the ErrAllBrokersDown show-case below
+		// when using localhost brokers on OSX, since the OSX resolver
+		// will return the IPv6 addresses first.
+		// You typically don't need to specify this configuration property.
```
mhowlett, line 55: i don't think we need this.
we should have a "how to really do error handling properly" example which includes this, and point people to that. i think keep this example focussed on serdes.

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
```diff
@@ -0,0 +1,134 @@
...
+					fmt.Printf("%% Headers: %v\n", e.Headers)
+				}
+			case kafka.Error:
+				// Errors should generally be considered
+				// informational, the client will try to
+				// automatically recover.
+				// But in this example we choose to terminate
+				// the application if all brokers are down.
+				fmt.Fprintf(os.Stderr, "%% Error: %v: %v\n", e.Code(), e)
+				if e.Code() == kafka.ErrAllBrokersDown {
```
mhowlett, line 118: i think remove this in this example. and the two lines of comment above related to it.

examples/avro_specific_consumer_example/avro_specific_consumer_example.go
```diff
@@ -0,0 +1,129 @@
...
+	sigchan := make(chan os.Signal, 1)
+	signal.Notify(sigchan, syscall.SIGINT, syscall.SIGTERM)
+
+	c, err := kafka.NewConsumer(&kafka.ConfigMap{
+		"bootstrap.servers": broker,
+		// Avoid connecting to IPv6 brokers:
+		// This is needed for the ErrAllBrokersDown show-case below
+		// when using localhost brokers on OSX, since the OSX resolver
+		// will return the IPv6 addresses first.
+		// You typically don't need to specify this configuration property.
```
mhowlett, line 55: ditto. remove i think. same for all examples.

examples/avro_specific_consumer_example/avro_specific_consumer_example.go
```diff
@@ -0,0 +1,129 @@
...
+					fmt.Printf("%% Headers: %v\n", e.Headers)
+				}
+			case kafka.Error:
+				// Errors should generally be considered
+				// informational, the client will try to
+				// automatically recover.
+				// But in this example we choose to terminate
+				// the application if all brokers are down.
+				fmt.Fprintf(os.Stderr, "%% Error: %v: %v\n", e.Code(), e)
+				if e.Code() == kafka.ErrAllBrokersDown {
```
mhowlett, line 118: ditto.

examples/avro_specific_producer_example/avro_specific_producer_example.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	ser, err := avro.NewSpecificSerializer(client, serde.NewConfig(), serde.ValueSerde)
+
+	if err != nil {
+		fmt.Printf("Failed to create serializer: %s\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Printf("Created serializer\n")
```
mhowlett, line 64: ditto 

(this is a nit... )

examples/json_consumer_example/json_consumer_example.go
```diff
@@ -0,0 +1,134 @@
...
+	fmt.Printf("Created Consumer %v\n", c)
+
+	client, err := schemaregistry.NewClient(schemaregistry.NewConfig(url))
+
+	if err != nil {
+		fmt.Printf("Failed to create schema registry client: %s\n", err)
+		os.Exit(1)
+	}
+
+	deser, err := jsonschema.NewDeserializer(client, serde.NewConfig(), serde.ValueSerde, serde.DisableValidation)
```
mhowlett, line 75: wouldn't `DisableValidation` be better as a config property?

examples/json_producer_example/json_producer_example.go
```diff
@@ -0,0 +1,101 @@
...
+	fmt.Printf("Created Producer %v\n", p)
+
+	client, err := schemaregistry.NewClient(schemaregistry.NewConfig(url))
+
+	if err != nil {
+		fmt.Printf("Failed to create schema registry client: %s\n", err)
+		os.Exit(1)
+	}
+
+	ser, err := jsonschema.NewSerializer(client, serde.NewConfig(), serde.ValueSerde, serde.DisableValidation)
```
mhowlett, line 57: same as deserialize - `DisableValidation` as config property.

schemaregistry/config.go
rayokota: golint complains if you specify `Url`

schemaregistry/serde/jsonschema/json_schema.go
rayokota: Ok, fixed

schemaregistry/config.go
```diff
@@ -0,0 +1,66 @@
...
+	SaslPassword  string
+
+	// Ssl Keys
+	SslCertificationLocation       string
+	SslKeyLocation                 string
+	SslCaLocation                  string
+	SslDisableEndpointVerification bool
+
+	// Timeouts
+	ConnectionTimeoutMs int
```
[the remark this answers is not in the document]
rayokota: Done

examples/json_consumer_example/json_consumer_example.go
rayokota: Added config prop `EnableValidation`

schemaregistry/confluent/meta.proto
```diff
@@ -0,0 +1,27 @@
...
+import "google/protobuf/descriptor.proto";
+
+option go_package="../confluent";
+
+message Meta {
+  string doc = 1;
+  map<string, string> params = 2;
+}
+
+extend google.protobuf.FileOptions {
```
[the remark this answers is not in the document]
rayokota: Yes :)

examples/avro_generic_consumer_example/avro_generic_consumer_example.go
```diff
@@ -0,0 +1,124 @@
...
+	fmt.Printf("Created Consumer %v\n", c)
+
+	client, err := schemaregistry.NewClient(schemaregistry.NewConfig(url))
+
+	if err != nil {
+		fmt.Printf("Failed to create schema registry client: %s\n", err)
+		os.Exit(1)
+	}
+
+	deser, err := avro.NewGenericDeserializer(client, avro.NewDeserializerConfig(), serde.ValueSerde)
```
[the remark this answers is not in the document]
rayokota: Ok, I'll fix

examples/avro_specific_producer_example/avro_specific_producer_example.go
```diff
@@ -0,0 +1,95 @@
...
+	fmt.Printf("Created Producer %v\n", p)
+
+	client, err := schemaregistry.NewClient(schemaregistry.NewConfig(url))
+
+	if err != nil {
+		fmt.Printf("Failed to create schema registry client: %s\n", err)
+		os.Exit(1)
+	}
+
+	ser, err := avro.NewSpecificSerializer(client, serde.ValueSerde, avro.NewSerializerConfig())
```
mhowlett, line 58: will the specific and generic serdes ever need different config?
FWIW, in .NET we use the same config class for both, like you are doing, but thought i'd ask the question.

examples/avro_generic_producer_example/avro_generic_producer_example.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	// Optional delivery channel, if not specified the Producer object's
+	// .Events channel is used.
+	deliveryChan := make(chan kafka.Event)
+
+	value := MyRecord{
+		ProductName: "Hello!",
+	}
+	payload, err := ser.Serialize(topic, value)
```
mhowlett, line 72: the other examples pass &value - is there a reason for the difference?

examples/json_consumer_example/json_consumer_example.go
```diff
@@ -0,0 +1,124 @@
...
+	url := os.Args[1]
+	bootstrapServers := os.Args[2]
+	group := os.Args[3]
+	topics := os.Args[4:]
+	sigchan := make(chan os.Signal, 1)
+	signal.Notify(sigchan, syscall.SIGINT, syscall.SIGTERM)
+
+	c, err := kafka.NewConsumer(&kafka.ConfigMap{
+		"bootstrap.servers": bootstrapServers,
+		// Avoid connecting to IPv6 brokers:
```
mhowlett, line 52: you don't need the `broker.address.family` config in any of the examples

mhowlett (approved):
One last round of minor comments, 
with that LGTM.

examples/json_consumer_example/json_consumer_example.go
rayokota: Done

examples/avro_specific_producer_example/avro_specific_producer_example.go
rayokota: I don't anticipate needed different configs

examples/avro_generic_producer_example/avro_generic_producer_example.go
rayokota: Ok, I modified the code so that the generic Avro serializer can also take a pointer.  This way the usage is always consistent.

emasab (approved):
LGTM. Thanks Robert!

schemaregistry/rest_service.go
```diff
@@ -0,0 +1,293 @@
...
+		err = configureUSERINFOAuth(conf, header)
+	default:
+		err = fmt.Errorf("unrecognized value for basic.auth.credentials.source %s", source)
+	}
+	return header, err
+}
+
+// handleRequest sends a HTTP(S) request to the Schema Registry, placing results into the response object
+func (rs *restService) handleRequest(request *api, response interface{}) error {
+	endpoint, err := rs.url.Parse(fmt.Sprintf(base+request.endpoint, request.arguments...))
```
jordaniversen, line 252: Hi @emasab, @edenhill,

Curious why is this done?  This seems to break for schema registry servers that are located at a specific path, not just a hostname.  For example if my schema registry endpoint is at "https://foo.bar/baz", and I want to call the GET "subjects" API, then this will send the request to host https://foo.bar with path "/subjects" instead of "/baz/subjects".

Was this an attempt to trim the potentially-trailing slash?  I.e. to change "https://foo.bar/baz/" to "https://foo.bar/baz"?

schemaregistry/serde/protobuf/protobuf.go
```diff
@@ -0,0 +1,470 @@
...
+		return msgIndexes[0:1]
+	default:
+		// parent is another MessageDescriptor.  We were nested so get that
+		// descriptor's indexes and append the index of this one
+		msgIndexes := toMessageIndexes(v, count+1)
+		return append(msgIndexes, index)
+	}
+}
+
+func ignoreFile(name string) bool {
```
jordaniversen, line 313: Hi @emasab

Curious why are these ignored?

PR #781, opened 2022-05-25
Improvements to Go client examples

emasab:
1. Changed the producer example to use the Events delivery channel. 
2. Added another producer example  using a custom delivery channel
3. Moved channel producer and consumer examples to the `examples/legacy` folder
4. Added missing descriptions for other examples
5. Consumer example with at-least-once guarantee by disabling `enable.auto.offset.store` and using `StoreMessage`


emasab:
moved to #801

PR #792, opened 2022-06-08
Docker example for AWS lambda

jliunyu:
Example for using go clients with AWS lambda.

### DIFF
--- a/examples/docker_aws_lambda_example/Dockerfile
+++ b/examples/docker_aws_lambda_example/Dockerfile
@@ -1,0 +1,23 @@
+FROM fedora:latest
+
+# Install pre-reqs
+RUN dnf install wget -y
+RUN dnf install gcc -y
+
+# Install librdkafka
+RUN rpm --import https://packages.confluent.io/rpm/5.4/archive.key
+RUN dnf install librdkafka-devel -y
+
+# Install Go v1.14
+RUN wget https://dl.google.com/go/go1.14.linux-amd64.tar.gz && tar -xvf go1.14.linux-amd64.tar.gz && rm go1.14.linux-amd64.tar.gz
+RUN mv go /usr/local
+ENV GOROOT=/usr/local/go
+ENV PATH="${GOROOT}/bin:${PATH}"
+
+# Build the producer_example
+WORKDIR /kafka
+COPY examples/docker_aws_lambda_example/go.mod .
+COPY examples/docker_aws_lambda_example/producer_example.go .
+RUN go build -o producer_example .
+
+ENTRYPOINT ["./producer_example"]
--- a/examples/docker_aws_lambda_example/Dockerfile
+++ b/examples/docker_aws_lambda_example/Dockerfile
@@ -4,12 +4,10 @@ FROM fedora:latest
 RUN dnf install wget -y
 RUN dnf install gcc -y
 
-# Install librdkafka
 RUN rpm --import https://packages.confluent.io/rpm/5.4/archive.key
-RUN dnf install librdkafka-devel -y
 
 # Install Go v1.14
-RUN wget https://dl.google.com/go/go1.14.linux-amd64.tar.gz && tar -xvf go1.14.linux-amd64.tar.gz && rm go1.14.linux-amd64.tar.gz
+RUN wget https://dl.google.com/go/go1.14.linux-amd64.tar.gz && tar -xvzf go1.14.linux-amd64.tar.gz && rm go1.14.linux-amd64.tar.gz
 RUN mv go /usr/local
 ENV GOROOT=/usr/local/go
 ENV PATH="${GOROOT}/bin:${PATH}"
--- a/examples/docker_aws_lambda_example/README.md
+++ b/examples/docker_aws_lambda_example/README.md
@@ -1,0 +1,15 @@
+# Build Docker image
+    $ docker build -f examples/docker_aws_lambda_example/Dockerfile -t goclients .
+
+# Push the docker image to AWS Elastic Container Registry
+  Create Amazon Elastic Container Registry first
+  Push the docker image to AWS ECR according to the [AWS ECR user guide](https://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html), or using all the commands under the `View push commands` of the ECR repository.
+
+# Create AWS lambda function using image from AWS ECR
+  Choose the `Container Image` when create the lambda function, add the docker image URL from `Container image URL`.
+
+# Config Environment Variables
+  Add the environment variables under the `Configuration`, we can pass the parameters like `bootstrap_servers`, `ccloudAPIKey`, `ccloudAPISecret` as environment variables. 
+
+# Run the test
+  Click the `Test` button under `Test`.
--- a/examples/docker_aws_lambda_example/README.md
+++ b/examples/docker_aws_lambda_example/README.md
@@ -9,7 +9,7 @@
   Choose the `Container Image` when create the lambda function, add the docker image URL from `Container image URL`.
 
 # Config Environment Variables
-  Add the environment variables under the `Configuration`, we can pass the parameters like `bootstrap_servers`, `ccloudAPIKey`, `ccloudAPISecret` as environment variables. 
+  Add the environment variables under the `Configuration`, we can pass the parameters like `BOOTSTRAP_SERVERS`, `CCLOUDAPIKEY`, `CCLOUDAPISECRET`, `TOPIC` as environment variables.
 
 # Run the test
   Click the `Test` button under `Test`.
--- a/examples/docker_aws_lambda_example/go.mod
+++ b/examples/docker_aws_lambda_example/go.mod
@@ -1,0 +1,3 @@
+module docker_example
+
+go 1.14
--- a/examples/docker_aws_lambda_example/go.mod
+++ b/examples/docker_aws_lambda_example/go.mod
@@ -1,3 +1,8 @@
 module docker_example
 
+require (
+	github.com/aws/aws-lambda-go v1.27.0
+	github.com/confluentinc/confluent-kafka-go v1.9.0 // indirect
+)
+
 go 1.14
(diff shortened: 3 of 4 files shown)


examples/docker_aws_lambda_example/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM fedora:latest
+
+# Install pre-reqs
+RUN dnf install wget -y
+RUN dnf install gcc -y
+
+# Install librdkafka
+RUN rpm --import https://packages.confluent.io/rpm/5.4/archive.key
```
edenhill, line 8: That's quite an old version of CP, but maybe that doesn't matter.

examples/docker_aws_lambda_example/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN dnf install librdkafka-devel -y
```
edenhill, line 9: Why is librdkafka required? It looks like the build step below uses the Go-bundled librdkafka.

examples/docker_aws_lambda_example/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Install Go v1.14
+RUN wget https://dl.google.com/go/go1.14.linux-amd64.tar.gz && tar -xvf go1.14.linux-amd64.tar.gz && rm go1.14.linux-amd64.tar.gz
```
edenhill, line 12: missing -z to tar (-xvzf) since it is a gzipped archive, or perhaps it understands that automatically?

examples/docker_aws_lambda_example/go.mod
```diff
@@ -0,0 +1,3 @@
+module docker_example
+
+go 1.14
```
edenhill, line 3: Shouldn't there be a require of confluent-kafka-go/kafka here somewhere?

examples/docker_aws_lambda_example/producer_example.go
```diff
@@ -0,0 +1,107 @@
...
+		fmt.Printf("Failed to create topic: %v\n", err)
+		os.Exit(1)
+	}
+
+	// Print results
+	for _, result := range results {
+		fmt.Printf("%s\n", result)
+	}
+
+	a.Close()
```
edenhill, line 71: Instead of instantiating two clients (see Matt's previous note on startup costs), you can instantiate an admin client from the Producer with NewAdminClientFromProducer()

edenhill (changes requested):
Nicely done!

examples/docker_aws_lambda_example/Dockerfile
mhowlett: what is it for? i don't understand this, maybe add a comment.

examples/docker_aws_lambda_example/producer_example.go
```diff
@@ -0,0 +1,129 @@
...
+	"os"
+
+	"github.com/aws/aws-lambda-go/lambda"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
+)
+
+var p *kafka.Producer
+
+func main() {
+	// Use signal to kill the client instance.
```
mhowlett, line 35: maybe just remove lines 35 and 36 i think.

examples/docker_aws_lambda_example/producer_example.go
```diff
@@ -0,0 +1,129 @@
...
+		if p != nil {
+			p.Close()
+		}
+		fmt.Printf("done")
+	}()
+
+	lambda.Start(HandleRequest)
+}
+
+// HandleRequest handles creating producer and
```
mhowlett, line 53: i think the comments on these two lines are not useful

examples/docker_aws_lambda_example/producer_example.go
```diff
@@ -0,0 +1,129 @@
...
+			"sasl.password":     ccloudAPISecret,
+		})
+
+		if err != nil {
+			fmt.Printf("Failed to create producer: %s\n", err)
+			os.Exit(1)
+		}
+	}
+
+	fmt.Printf("Created Producer %v\n", p)
```
mhowlett, line 80: this should be inside the p == null block

examples/docker_aws_lambda_example/producer_example.go
jliunyu: The CI will complain without this comments:

```
examples/docker_aws_lambda_example/producer_example.go:53:1: exported function HandleRequest should have comment or be unexported
Found 1 lint suggestions; failing.
The command "if [[ -f .do_lint ]]; then golint -set_exit_status ./examples/... ./kafka/... ./kafkatest/... ./soaktest/... ; fi" exited with 1.
```

examples/docker_aws_lambda_example/producer_example.go
```diff
@@ -0,0 +1,129 @@
...
+	a, err := kafka.NewAdminClientFromProducer(p)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	maxDur, err := time.ParseDuration("60s")
+	if err != nil {
+		panic("ParseDuration(60s)")
+	}
+	results, err := a.CreateTopics(
```
mhowlett, line 91: since you expect this to be called more than once, you really want "create if doesn't exist" semantics. i think you want to explain that a bit here. i think what you have gives the behavior you want, because the error will be on the partition basis, but it would be good to look at the result, and printf whether the topic was created or not.

examples/docker_aws_lambda_example/producer_example.go
```diff
@@ -0,0 +1,129 @@
...
+		ctx,
+		[]kafka.TopicSpecification{{
+			Topic:         topic,
+			NumPartitions: numParts}},
+		kafka.SetAdminOperationTimeout(maxDur))
+	if err != nil {
+		fmt.Printf("Failed to create topic: %v\n", err)
+		os.Exit(1)
```
mhowlett, line 99: can AWS functions return an error? `os.Exit(1)` may be over reacting because the client instance should be still good.

mhowlett: or maybe printf an error and return like you do below

examples/docker_aws_lambda_example/producer_example.go
jliunyu: The result is printed below:)

examples/docker_aws_lambda_example/producer_example.go
jliunyu: Yes, the func HandleRequest() can return an error

examples/docker_aws_lambda_example/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+# Push the docker image to AWS Elastic Container Registry
+  Create Amazon Elastic Container Registry first
+  Push the docker image to AWS ECR according to the [AWS ECR user guide](https://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html), or using all the commands under the `View push commands` of the ECR repository.
+
+# Create AWS lambda function using image from AWS ECR
+  Choose the `Container Image` when create the lambda function, add the docker image URL from `Container image URL`.
+
+# Config Environment Variables
+  Add the environment variables under the `Configuration`, we can pass the parameters like `bootstrap_servers`, `ccloudAPIKey`, `ccloudAPISecret` as environment variables.
```
mhowlett, line 12: like `BOOTSTRAP_SERVERS`, `TOPIC`, `CCLOUDAPIKEY`, and `CCLOUDAPISECRET` as environment variables.

examples/docker_aws_lambda_example/producer_example.go
```diff
@@ -0,0 +1,133 @@
...
+	// Create topics if it doesn't exist
+	results, err := a.CreateTopics(
+		ctx,
+		[]kafka.TopicSpecification{{
+			Topic:         topic,
+			NumPartitions: 1}},
+		kafka.SetAdminOperationTimeout(maxDur))
+	if err != nil {
+		fmt.Printf("Failed to create topic: %v\n", err)
+		return err
```
mhowlett, line 98: you don't want to quit if there is a "topic already exists" error, otherwise this lambda function can only be used once.

mhowlett: need some sort of check like:

https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/examples/ExactlyOnce/Program.cs#L233

examples/docker_aws_lambda_example/producer_example.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	// Print results
+	for _, result := range results {
```
mhowlett, line 102: print if there is no error.

examples/docker_aws_lambda_example/producer_example.go
mhowlett: (if there is an error that isn't topic already exists, then print the error and exit)

mhowlett: ahh, ok, we did .NET differently.

maybe change this to `CreateTopics request failed: `

examples/docker_aws_lambda_example/producer_example.go
```diff
@@ -0,0 +1,133 @@
...
+		[]kafka.TopicSpecification{{
+			Topic:         topic,
+			NumPartitions: 1}},
+		kafka.SetAdminOperationTimeout(maxDur))
+	if err != nil {
+		fmt.Printf("Failed to create topic: %v\n", err)
+		return err
+	}
+
+	// Print results
```
mhowlett, line 101: seems like you could get rid of this comment ...

mhowlett (approved):
lgtm, after small suggestions.

emasab:
@jliunyu remember to update the `examples/README.md` and `examples/.gitignore` with the description and the binary file produced

emasab (approved):

PR #809, opened 2022-06-30
Import v1.9.1 rc1 with osx arm64/m1 support

edenhill:


edenhill:
This needs testing on OSX specifically, both x64 and arm64

PR #810, opened 2022-07-01
changed confluent cloud example

emasab:
* added SR serde with Confluent Cloud and authentication


mhowlett (approved):
yep, good idea. thanks!

PR #812, opened 2022-07-01
added copyright to some schema registry files and

emasab:
moved copyright position on top of the files
with one newline before package definition to avoid
being interpreted as the package's documentation


mhowlett (approved):
LGTM - what you are moving to makes more sense to me.

But I don't know if there is a good reason for the previous convention

PR #814, opened 2022-07-02
fix some typos

cuishuang:


kafka/producer.go
```diff
@@ -495,7 +495,7 @@ func NewProducer(conf *ConfigMap) (*Producer, error) {
 		// produce.offset.report is no longer used in librdkafka >= v1.0.0
 		v, _ = confCopy.extract("{topic}.produce.offset.report", nil)
 		if v == nil {
-			// Enable offset reporting by default, unless overriden.
+			// Enable offset reporting by default, unless override.
```
edenhill, line 498: This should be `overridden`, not `override`.

PR #815, opened 2022-07-04
avoid cache stampede if key is not found.

emasab:
When the key is not in cache and multiple goroutines
call the Serialize method at once, many of them could
start calls to SR to put in cache the same schema, this way
only the first one makes the call and the rest
read the cached result.

### DIFF
--- a/schemaregistry/schemaregistry_client.go
+++ b/schemaregistry/schemaregistry_client.go
@@ -265,14 +265,21 @@ func (c *client) Register(subject string, schema SchemaInfo, normalize bool) (id
 	metadata := SchemaMetadata{
 		SchemaInfo: schema,
 	}
-	err = c.restService.handleRequest(newRequest("POST", versionNormalize, &metadata, url.PathEscape(subject), normalize), &metadata)
-	if err != nil {
-		return -1, err
-	}
 	c.schemaCacheLock.Lock()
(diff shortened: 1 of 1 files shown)


mhowlett (approved):
thanks @emasab - LGTM!

schemaregistry/schemaregistry_client.go
```diff
@@ -265,14 +265,21 @@ func (c *client) Register(subject string, schema SchemaInfo, normalize bool) (id
...
-	err = c.restService.handleRequest(newRequest("POST", versionNormalize, &metadata, url.PathEscape(subject), normalize), &metadata)
-	if err != nil {
-		return -1, err
-	}
 	c.schemaCacheLock.Lock()
-	c.schemaCache.Put(cacheKey, metadata.ID)
+	// another goroutine could have already put it in cache
+	idValue, ok = c.schemaCache.Get(cacheKey)
+	if !ok {
+		err = c.restService.handleRequest(newRequest("POST", versionNormalize, &metadata, url.PathEscape(subject), normalize), &metadata)
```
edenhill, line 272: This will block all other threads/go-routines to access the schema cache during the duration of the REST call, which could take a long time. Is this intentional?

One way to avoid this is to also have a lock on each cache entry and only keep that lock held during the request, this way accesses to other schemas are not affected by retrieving this one. (Do pay attention to lock ordering though).

PR #821, opened 2022-07-07
Schema registry static api documentation

emasab:


mk/doc-gen.py
```diff
@@ -18,6 +19,10 @@ def convert_path(url, base_url, after):
 
 
 if __name__ == '__main__':
+    if len(sys.argv) < 2:
```
mhowlett, line 22: is there a README.md somewhere that needs to be updated to reflect this?

mhowlett (approved):

mk/doc-gen.py
emasab: no, the README.md just states to call the make target

Issue #778, opened 2022-05-05
Offset Set function should allow Offset as parameter.


MatthL-SwiftC:
Description
===========

When using the [Offset.Set](https://github.com/confluentinc/confluent-kafka-go/blob/v1.8.2/kafka/offset.go#L65-L73) function I'd expect to be able to pass a value of the Offset type such as the predefined constant [OffsetBeginning](https://github.com/confluentinc/confluent-kafka-go/blob/v1.8.2/kafka/offset.go#L38), however only a string or int64 representation is valid.

How to reproduce
================

Use  [Offset.Set](https://github.com/confluentinc/confluent-kafka-go/blob/v1.8.2/kafka/offset.go#L65-L73) with one of the [predefined Offset constants](https://github.com/confluentinc/confluent-kafka-go/blob/v1.8.2/kafka/offset.go#L38-L47).

Checklist
=========
Please provide the following information:

 - [X] confluent-kafka-go and librdkafka version (`LibraryVersion()`): master
 - [x] Apache Kafka broker version: N/A
 - [x] Client configuration: N/A
 - [x] Operating system: N/A
 - [x] Provide client logs (with `"debug": ".."` as necessary) N/A
 - [x] Provide broker log excerpts N/A
 - [ ] Critical issue

PR #844, opened 2022-08-14
add offset type to NewOffset func

jdockerty:
Closes https://github.com/confluentinc/confluent-kafka-go/issues/778

Adding the ability to use a concrete `Offset` type with the `NewOffset` func, by extension this also enables it through the `Offset.Set` func.

An `Offset` is an `int64` type in itself, so it can be returned as such.


PrasanthV454:
Closing this and created #962

Issue #806, opened 2022-06-25
Unreachable code in README docs


snosratiershad:
Description
===========
In "High-level balanced consumer" part of "Examples" section of README file, it's impossible to reach `c.Close` method because for loop doesn't have any conditions and will run for ever.

How to reproduce
================
I think it can be proved by linting code with `go vet` and `staticcheck`.

Checklist
=========
Please provide the following information:

 - [ ] confluent-kafka-go and librdkafka version (`LibraryVersion()`):
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `ConfigMap{...}`
 - [ ] Operating system:
 - [ ] Provide client logs (with `"debug": ".."` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

mhowlett:
thanks for pointing this out!

milindl:
Closed with #846

PR #846, opened 2022-08-22
docs: correct consumer example code in README

snosratiershad:
Signed-off-by: Salar Nosrati-Ershad <snosratiershad@gmail.com>

as discussed in [this issue](https://github.com/confluentinc/confluent-kafka-go/issues/806).


milindl (changes requested):
@snosratiershad Thank you for the PR.

While the defer approach seems to be a logical solution, it won't actually work, since the loop runs forever, and can only really be terminated by Ctrl-C. The best way to handle this would be to handle SIGINT, and on getting that signal, set a variable to terminate the loop. 

You can look at how it is done in the C code for an idea: https://github.com/edenhill/librdkafka/blob/2f331f7c2ece4b76d5e0f07c63c5342febd86db6/examples/consumer.c#L197

Please tell me if you require any help with the code.

snosratiershad:
@milindl Thanks for your review, I didn't knew about singals before in golang.
I think this change may fix it.

README.md
```diff
@@ -59,6 +63,15 @@ func main() {
 
 	c.SubscribeTopics([]string{"myTopic", "^aRegex.*[Tt]opic"}, nil)
 
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGINT)
+	go func() {
+		<-sig
+		fmt.Println("Keyboard interrupted.")
+		c.Close()
+		os.Exit(1)
```
milindl, line 72: Minor: If we use a `closed` boolean as given in the other comment, we won't need to use os.Exit at all

milindl (changes requested):
Thanks for the updated change, it looks much better. A few minor changes are required before we can push it.

snosratiershad:
@milindl, just recently applied your change requests. feel free to contact me if there is anything, that can I help about it.

milindl (approved):

README.md
```diff
@@ -59,8 +64,16 @@ func main() {
 
 	c.SubscribeTopics([]string{"myTopic", "^aRegex.*[Tt]opic"}, nil)
 
-	for {
-		msg, err := c.ReadMessage(-1)
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGINT)
```
milindl, line 68: Please change this to os.Interrupt instead of syscall.SIGINT, and remove the import for syscall, to ensure this code works across platforms including Windows. 

Here's more info for signals - https://pkg.go.dev/os#Signal - the part where it mentions that the we are guaranteed only 2 signal values across platforms (os.Interrupt and os.Kill)

Minor - if you can, add a comment above the `sig := ...` line explaining the purpose of this goroutine.

milindl (changes requested):
Thanks again for your changes, I had thought we could merge it, but spotted a small issue which might cause problems running it cross platform. Looks good otherwise.

snosratiershad:
@milindl Done.

README.md
```diff
@@ -59,8 +63,17 @@ func main() {
 
 	c.SubscribeTopics([]string{"myTopic", "^aRegex.*[Tt]opic"}, nil)
 
-	for {
-		msg, err := c.ReadMessage(-1)
+	sig := make(chan os.Signal, 1)
+	// get notified by signal when interrupted and break the consuming loop
```
milindl, line 67: nit: Capitalize 'G' to a match other comments in readme

README.md
```diff
@@ -59,8 +63,17 @@ func main() {
...
+	go func() {
+		<-sig
+		closed = true
+	}()
+
+	for !closed {
+		msg, err := c.ReadMessage(100 * time.Millisecond)
 		if err == nil {
 			fmt.Printf("Message on %s: %s\n", msg.TopicPartition, string(msg.Value))
 		} else {
```
milindl, line 79: I tested this and realized we should probably change this else clause to something like

`else if err.(kafka.Error).Code() != kafka.ErrTimedOut { ... }`

Reason for this is, by making ReadMessage have a timeout, the console will be filled with a ton of timeout messages

milindl (changes requested):
Thanks for your change again. I tested it locally, found one small issue.

README.md
```diff
@@ -59,12 +63,23 @@ func main() {
 
 	c.SubscribeTopics([]string{"myTopic", "^aRegex.*[Tt]opic"}, nil)
 
-	for {
-		msg, err := c.ReadMessage(-1)
+	sig := make(chan os.Signal, 1)
+	// Get notified by signal when interrupted and break the consuming loop.
+	signal.Notify(sig, os.Interrupt)
```
emasab, line 68: add ", syscall.SIGTERM" to exit gracefully from kill

README.md
```diff
@@ -42,6 +42,10 @@ High-level balanced consumer
 ```golang
 import (
 	"fmt"
+	"os"
+	"os/signal"
+	"time"
+
```
emasab, line 48: add "syscall"

emasab (dismissed):
Nice job, left a comment on handling SIGTERM too

edenhill (commented):
I'm not super-convinced this is very helpful; it bloats the first example people will see with stuff that is unrelated to the Kafka client.
I'd prefer if we don't go down this approach for the README, but keep those examples as minimal and easy to follow as possible. We have the examples/ directory for fully fleshed examples.

We shouldn't need to teach people how to break a loop in Go.

Maybe it is sufficient with something like:
```golang
run := true  // You could use a signal handler or similar to set this to false to break the loop
for run {
...
```
```

**Comment by emasab** [comment]
@edenhill what if we don't put a complete example into the README, we could leave snippets of code of the Kafka client, that don't compile alone, and say to look in the examples sections for complete ones.

**Comment by edenhill** [comment]
I think the current format is reasonable, a small buildable example that focuses on simple usage, followed by a link to more examples.

**Comment by milindl** [comment]
Changed the PR for changes to use "run" with the given comment.

File: README.md [hunk 7]
```diff
@@ -59,8 +61,9 @@ func main() {
 
 	c.SubscribeTopics([]string{"myTopic", "^aRegex.*[Tt]opic"}, nil)
 
-	for {
-		msg, err := c.ReadMessage(-1)
+	run := true // A signal handler or similar could be used to set this to false to break the loop.
+	for run {
+		msg, err := c.ReadMessage(5 * time.Second)
```
edenhill, line 66: If using a timeout you will now also need to handle the timeout case specifically below, since it is not an error.

edenhill (changes requested):

README.md
milindl: Thank @edenhill, I wasn't sure what would be the best way for this snippet.

I added the timeout because without the timeout, the loop never terminates if we are polling to a topic that messages aren't being produced to. 

I can remove it entirely and keep it -1, if that is what you suggest to keep the example simple.

edenhill: No I think it makes sense to have a timeout here as the run var would be a bit useless otherwise.
But you need to check for the timeout case, I think it is the msg==nil,err==nil, but please double check (and test).

milindl: I think with ReadMessage we give an `err.(kafka.Error).Code() != kafka.ErrTimedOut`, I'll confirm and change it

README.md
```diff
@@ -59,12 +61,17 @@ func main() {
...
-		msg, err := c.ReadMessage(-1)
+	// A signal handler or similar could be used to set this to false to break the loop.
+	run := true
+
+	for run {
+		msg, err := c.ReadMessage(time.Second)
 		if err == nil {
 			fmt.Printf("Message on %s: %s\n", msg.TopicPartition, string(msg.Value))
-		} else {
+		} else if err.(kafka.Error).Code() != kafka.ErrTimedOut {
```
edenhill, line 71: Perfect :+1:

We should make a convenience function on the Error type, so you can do:  `if err.(Kafka.Error).IsTimeout() {`

edenhill (approved):
Good stuff!

snosratiershad:
@milindl Thanks.

Issue #843, opened 2022-08-09
`avro_specific.Deserialize()` returns wrong things


falau:
Description
===========
The method `Deserialize()` of the specific Avro deserializer has the signature like this:
https://github.com/confluentinc/confluent-kafka-go/blob/1092e015153c87b7d282f224b9d3da597eaf2637/schemaregistry/serde/avro/avro_specific.go#L106

Although this method is not documented(nor tested) well, I expect the first return should be of the message type produced by the `MessageFactory` for the deserializer.
However, the first returned object is the return of `vm.Eval()`, and it's an `error` object:
https://github.com/confluentinc/confluent-kafka-go/blob/1092e015153c87b7d282f224b9d3da597eaf2637/schemaregistry/serde/avro/avro_specific.go#L142

This behavior basically makes `Deserialize()` unusable since usually there is no way for the caller of `Deserialize()` to access the message produced by the `MessageFactory`.

How to reproduce
================
- Init a Schema Registry client and prepare a gogen-avro generated struct
- with the following code:
```go
	// srClient: the Schema Registry client
	deser, _ := avro.NewSpecificDeserializer(srClient, serde.ValueSerde, avro.NewDeserializerConfig())
	deser.MessageFactory = func(subject, name string) (interface{}, error) {
		// MyMessageType: a specific type generated by gogen-avro
		var myMessage MyMessageType
		return &myMessage, nil
	}
	decodedValue, err := deser.Deserialize(*kafkaMsg.TopicPartition.Topic, kafkaMsg.Value)
        // decodedValue would be a nil if everything goes well :/
```

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-go and librdkafka version (`LibraryVersion()`): v1.9.2 / v1.9.2
 - [x] Apache Kafka broker version: irrelevant
 - [x] Client configuration: `ConfigMap{...}`: irrelevant
 - [x] Operating system: irrelevant
 - [x] Provide client logs (with `"debug": ".."` as necessary): irrelevant
 - [x] Provide broker log excerpts: irrelevant
 - [ ] Critical issue

milindl:
Thanks for reporting this, we're looking into it.

PR #849, opened 2022-08-29
Fix Deserialize (avro_specific) and improve tests for avro_specific/avro_generic

milindl:
1. Fixes avro_specific Deserialize method which did not return the
   right thing (it did not return anything useable).
2. Update tests for (1) - add Deserialize to tests which were only
   testing DeserializeInto
3. Add Deserialize to avro_generic tests as well.

Addresses  #843


rayokota (approved):
Thanks @milindl , good catch.  LGTM

PR #853, opened 2022-09-06
[consumer] Propagate errors for Pause/Resume APIs

### LINKED ISSUE #808: Consumer Resume/Pause Propagating Errors
Description
===========

Reading the documentation for the resume and pause functions for consumers in librdkafka [here](https://docs.confluent.io/5.0.4/clients/librdkafka/classRdKafka_1_1Handle.html#a322d16d6786da2b4b20505f19325fcec), it seems as though the resume and pause functions that this library wraps always returns a `NO_ERROR` and instead adds the error from resuming/pausing the consumer to the partitions list passed to the function. However it does not seem as though the implementation of the pause/resume function looks for the errors within the partitions list when considering whether or not to return an error. Is this a correct understanding of the library's implementation? Is there a way to propagate the error added to the partition list in librdkafka to users of this library?

Checklist
=========

 - [ ] confluent-kafka-go and librdkafka version (`LibraryVersion()`):
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `ConfigMap{...}`
 - [ ] Operating system:
 - [ ] Critical issue

milindl:
This propagates the individual errors for each TopicPartition
that is being Paused/Resumed, into the input slice itself.
An non-nil cumulative error is returned in case one or more of
the TopicPartitions has an error.
This approach is different from other such APIs, which return a
new slice of TopicPartitions, but that would break the API.
This method keeps the existing API to not break existing code, but
allows users to access errors which were not being made available
earlier through the input.
Fixes #808

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -671,6 +671,23 @@ func (c *Consumer) Position(partitions []TopicPartition) (offsets []TopicPartiti
 	return newTopicPartitionsFromCparts(cparts), nil
 }
 
+// populateErrorsFromTopicPartitionToTopicPartition sets the error from the
+// first i partitions of src into the first i partitions of dst, where i is
+// min(len(src), len(dst)).
+// Returns the total number of not-nil errors set on dst.
+func populateErrorsFromTopicPartitionToTopicPartition(
+	src []TopicPartition, dst []TopicPartition) int {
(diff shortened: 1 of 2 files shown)


chamilto:
Please 👍

PR #854, opened 2022-09-06
Implement callback for SASL PLAIN credentials

josvisser66:
# Introduction

This PR is the sibling of https://github.com/edenhill/librdkafka/pull/3955 which adds a feature to librdkafka that allows the specification of a callback to obtain the SASL PLAIN credentials.

The Golang implementation works as follows:

```
func credsCallback() (string, string, error) {
        // Magic here.
        return username, password, nil
}

func main() {
        c, err := kafka.NewConsumer(&kafka.ConfigMap{
                "bootstrap.servers": "pkc-XXXXXX.confluent.cloud:9092",
                "group.id":          "myGroup",
                "sasl.mechanism":    "PLAIN",
                "security.protocol": "SASL_SSL",
                "plain.creds.cb":    credsCallback,
        })
```

Whenever librdkafka needs the credentials it calls (through some intermediaries) the Golang callback specified in the "plain.creds.cb" key and uses whatever it gets back from there.

# Implementation details

There is some excitement in this PR because of the way the C/Go bridge works and the consequences of that. The PR provides a single callback function for all Golang Kafka clients (in C) which then branches out to the "right" one for this particular Kafka client (using the pointer address of the Kafka client structure as a key in a map). There is a drawback to this approach, which is that we can only update that map when we know that pointer address and this means that there is a short coverage gap where the client could call the callback and not get any results. Testing so far seems to indicate that this is not a big problem.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/confluentinc/confluent-kafka-go
+module github.com/josvisser66/confluent-kafka-go
 
 go 1.13
 
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/josvisser66/confluent-kafka-go
+module github.com/confluentinc/confluent-kafka-go
 
 go 1.13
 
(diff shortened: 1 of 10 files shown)


josvisser66:
We decided on another approach.

PR #862, opened 2022-09-27
[docs] Add instructions on building a static executable for Linux

milindl:
We can either make sure to ensure that the gcc version mismatch isn't there, or use musl for static builds.


emasab (approved):
Nice documentation!

PR #864, opened 2022-09-29
[producer] Fix race between Len() and produceChannel receive

### LINKED ISSUE #601: race condition with channel based producer when calling flush
I think there's a race condition when producing a message to `p.ProduceChannel()`, then immediately calling Flush. I think the following sequence is possible:

message take from `produceChannel` here: https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/producer.go#L556

length of channel is queried here and found to be zero, hence flush skipped: https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/producer.go#L328

`p.produce(m, C.RD_KAFKA_MSG_F_BLOCK, nil)` called here: https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/producer.go#L557

milindl:
At the current moment, the code to process the produceChannel does something like:

1. recv from channel, some message M
2. actually process M

The problem is that while the code is between steps 1 and 2, a call to Len() can give faulty results, since the message is not counted either in the part of the channel, the producer.events, or the librdkafka outq.

This will fix #601 .

The tradeoff here is that the Len() call will now need to write to a channel twice rather than being a just a sum of three things. Not only this, but while actually processing the Len() call (or simultaneous Len() calls), we will not make any progress in processing the produceChannel.

However - I think it's justified for the sake of correctness, and the secondly because (I hope) that Len is called much less frequently than actually producing messages (the actual production of messages is less affected, just one 'select' statement more). librdkafka uses a similar lock while getting the count of pending messages while flushing, in `rd_kafka_curr_msgs_cnt`

### DIFF
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -136,6 +136,12 @@ type Producer struct {
 
 	// Terminates the poller() goroutine
 	pollerTermChan chan bool
+
+	// To make sure we are not receiving from the produce channel the same
+	// time as reading its length, an intent to read the length must be
+	// communicated on this channel by sending `true`, and `false` when
+	// done. See Len().
+	lenIntentChan chan bool
 }
 
 // String returns a human readable name for a Producer instance
@@ -325,7 +331,12 @@ func (p *Producer) ProduceChannel() chan *Message {
 // as well as delivery reports queued for the application.
 // Includes messages on ProduceChannel.
 func (p *Producer) Len() int {
-	return len(p.produceChannel) + len(p.events) + int(C.rd_kafka_outq_len(p.handle.rk))
+	// This will block indefinitely if Close() has been called before calling Len(). But no
+	// method should be called after Close().
+	p.lenIntentChan <- true
+	sum := len(p.produceChannel) + len(p.events) + int(C.rd_kafka_outq_len(p.handle.rk))
+	p.lenIntentChan <- false
+	return sum
 }
 
 // Flush and wait for outstanding messages and requests to complete delivery.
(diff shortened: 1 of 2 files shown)


kafka/producer.go
```diff
@@ -325,7 +331,12 @@ func (p *Producer) ProduceChannel() chan *Message {
 // as well as delivery reports queued for the application.
 // Includes messages on ProduceChannel.
 func (p *Producer) Len() int {
-	return len(p.produceChannel) + len(p.events) + int(C.rd_kafka_outq_len(p.handle.rk))
+	// This will block indefinitely if Close() has been called before calling Len(). But no
+	// method should be called after Close().
+	p.lenIntentChan <- true
```
emasab, line 336: It's not ensured that Len is executed before the select, it could be that the channelProducer is at line 578 and the len terminates without blocking the read from channel that had already happened. The sequence is:
g1:578, g2:337, g2:338, g2:339, g1:581

edenhill:
This seems like an overly complex solution and it is hard to assert its correctness.

Instead of using channels, can't we simply have an atomic counter for these intermediary messages?

emasab:
@edenhill the thing is that it's not ensured that the Len is not executed between reading the message from the channel and incrementing the atomic counter. It reduces the likelihood for sure.

edenhill:
The atomic can be incremented prior to sending on the producerChannel and after produce() has returned.
It won't matter if messages are double accounted for in the flush case.

edenhill:
Didn't we deprecate the channel based clients btw?

milindl:
@edenhill : Yep, we did, that's the primary question Emanuele and I had to consider before looking at the PR -- if we've deprecated channel based clients, are we better off just documenting this race condition and letting things be?

edenhill:
Yeah I think so - document rather than fix.

But we need to make it clearer that the channel based clients are in fact deprecated.

milindl:
Not fixing this issue. Closing with #894 which emphasizes the deprecation.

PR #865, opened 2022-09-30
Draft for integration tests setup

milindl:
This is a draft PR intended to illustrate the usage of testcontainers and testify in the go integration tests.
For the purposes of this demonstration, only the AdminClient related tests have been changed for the correct setup.

Using a command line flag, it's also possible to make testcontainer usage optional - we can use them in CI but not in our local setup, for example. In this case, I have made the usage of testcontainers false by default.

Any changes to the things that need to be brought up can be made inside the docker compose file -- for now there's only kafka+zookeeper containers.

Also - besides the rough edges in code, I have not added my go.sum updates.

The commands to run it are:
```
# run all admin tests without anything extra, just like the normal setup
go test -run  ^TestAdminClient$
# run with automatic docker container up and down
go test -run  ^TestAdminClient$  -clients.docker true
# run selected test, in this case TestAdminTopics only
go test -run  ^TestAdminClient$  -clients.docker true -testify.m ^TestAdminTopics$
```

### DIFF
--- a/kafka/integration_test.go
+++ b/kafka/integration_test.go
@@ -27,6 +27,9 @@ import (
 	"sort"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/suite"
+	"github.com/testcontainers/testcontainers-go"
 )
 
 // producer test control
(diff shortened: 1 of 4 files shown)


kafka/integration_test.go
```diff
@@ -1208,6 +1211,38 @@ func TestProducerConsumerHeaders(t *testing.T) {
...
+	suite.Suite
+	compose *testcontainers.LocalDockerCompose
+}
+
+func (ats *AdminTestSuite) SetupSuite() {
+	testconfInit()
+	if !testconf.Docker {
+		return
+	}
+	ats.compose = testcontainers.NewLocalDockerCompose([]string{"testresources/docker-compose.yaml"}, "test-docker")
```
mhowlett, line 1227: the two other occasions i've needed to deal with the test code being responsible for managing the lifecycle of the docker containers it makes use of, it's brought more of a pain than a benefit.

that makes me wary of doing this.

it's not very hard to require the user manually do this, or make the script running the tests responsible for it.

milindl: Based on the time it took me to get this to a working state, I'm inclined to agree, it's probably a lot of effort to get it running, especially on the CI. We should just maybe leave the docker-compose file and make the rest of the work be done by a script.

milindl:
No longer needed, we are actually doing this (not as a draft)

PR #866, opened 2022-10-03
[DP-9090]add run_semaphore_public

lyao-77:
A new parameter "run_semaphore_public" is added into manifest and is used to whitelist project that are public on github but could run semaphore plugin.
For new project, if we want to avoid this kind of error and change the manifest manually, one possible solution is to add a new parameter of "run_semaphore_public" in Jenkins so that this parameter will be passed in to projects.json, which will add this parameter to manifest by default.

### DIFF
--- a/service.yml
+++ b/service.yml
@@ -8,3 +8,4 @@ github:
   repo_name: confluentinc/confluent-kafka-go
 semaphore:
   enable: true
+  run_semaphore_public: true


service.yml
```diff
@@ -8,3 +8,4 @@ github:
   repo_name: confluentinc/confluent-kafka-go
 semaphore:
   enable: true
+  run_semaphore_public: true
```
acourtneybrown, line 11: i don't think this is the right place to allow this configuration.  it might make sense in `projects.json`, but since the team owning this repo can approve/merge a PR adding this config, it wouldn't really provide any roadblock.

acourtneybrown (changes requested):

service.yml
lyao-77: From https://github.com/confluentinc/cc-service-bot/blob/master/plugin_semaphore.py#L53, I think all of the parameters to construct the Semaphore bot is from the manifest. I looked into the projects.json, and I feel like that file is used to create default manifest for the new project: https://github.com/confluentinc/cc-service-bot/blob/master/plugin_semaphore.py#L53 
And that's why I left a comment for this pr: "For new project, if we want to avoid this kind of error and change the manifest manually, one possible solution is to add a new parameter of "run_semaphore_public" in Jenkins so that this parameter will be passed in to projects.json, which will add this parameter to manifest by default."

PR #872, opened 2022-10-12
Added validations on Producer

PrasanthV454:
1. Added isClosed new field to Producer struct which acts as flag to know if Producer is open or closed.
2. Added new function verifyProducer() function which checks if producer is valid or not by doing null checks and closed checks to avoid seg faults for C-GO or go-panics.

### DIFF
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -76,6 +76,16 @@ import "C"
 type AdminClient struct {
 	handle    *handle
 	isDerived bool // Derived from existing client handle
+
+	isClosed bool // to check if Admin Client is closed or not.
+}
+
+// verifyAdminClient verifies if AdminClient can be used
+func (a *AdminClient) verifyAdminClient() bool {
+	if a == nil || a.isClosed {
+		return false
+	}
+	return true
 }
 
 func durationToMilliseconds(t time.Duration) int {
@@ -560,6 +570,10 @@ type DeleteACLsResult = DescribeACLsResult
 // The returned result event is checked for errors its error is returned if set.
 func (a *AdminClient) waitResult(ctx context.Context, cQueue *C.rd_kafka_queue_t, cEventType C.rd_kafka_event_type_t) (rkev *C.rd_kafka_event_t, err error) {
 
+	if a.verifyAdminClient() == false {
+		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	}
+
 	resultChan := make(chan *C.rd_kafka_event_t)
 	closeChan := make(chan bool) // never written to, just closed
 
@@ -623,6 +637,10 @@ func (a *AdminClient) waitResult(ctx context.Context, cQueue *C.rd_kafka_queue_t
 // cToTopicResults converts a C topic_result_t array to Go TopicResult list.
 func (a *AdminClient) cToTopicResults(cTopicRes **C.rd_kafka_topic_result_t, cCnt C.size_t) (result []TopicResult, err error) {
 
+	if a.verifyAdminClient() == false {
+		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	}
+
 	result = make([]TopicResult, int(cCnt))
 
 	for i := 0; i < int(cCnt); i++ {
@@ -638,6 +656,10 @@ func (a *AdminClient) cToTopicResults(cTopicRes **C.rd_kafka_topic_result_t, cCn
 
 // cConfigResourceToResult converts a C ConfigResource result array to Go ConfigResourceResult
 func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t, cCnt C.size_t) (result []ConfigResourceResult, err error) {
+
+	if a.verifyAdminClient() == false {
+		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	}
 
 	result = make([]ConfigResourceResult, int(cCnt))
 
@@ -671,6 +693,11 @@ func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t
 //
 // Requires broker version >= 0.10.0.
 func (a *AdminClient) ClusterID(ctx context.Context) (clusterID string, err error) {
+
+	if a.verifyAdminClient() == false {
+		return "", newErrorFromString(ErrState, "not a valid adminClient state.")
+	}
+
 	responseChan := make(chan *C.char, 1)
 
 	go func() {
@@ -703,6 +730,11 @@ func (a *AdminClient) ClusterID(ctx context.Context) (clusterID string, err erro
 //
 // Requires broker version >= 0.10.0.
 func (a *AdminClient) ControllerID(ctx context.Context) (controllerID int32, err error) {
+
+	if a.verifyAdminClient() == false {
+		return 0, newErrorFromString(ErrState, "not a valid adminClient state.")
+	}
+
 	responseChan := make(chan int32, 1)
 
 	go func() {
@@ -732,6 +764,11 @@ func (a *AdminClient) ControllerID(ctx context.Context) (controllerID int32, err
 //
 // Note: TopicSpecification is analogous to NewTopic in the Java Topic Admin API.
 func (a *AdminClient) CreateTopics(ctx context.Context, topics []TopicSpecification, options ...CreateTopicsAdminOption) (result []TopicResult, err error) {
+
+	if a.verifyAdminClient() == false {
+		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	}
+
 	cTopics := make([]*C.rd_kafka_NewTopic_t, len(topics))
 
 	cErrstrSize := C.size_t(512)
@@ -846,6 +883,11 @@ func (a *AdminClient) CreateTopics(ctx context.Context, topics []TopicSpecificat
 //
 // Requires broker version >= 0.10.1.0
 func (a *AdminClient) DeleteTopics(ctx context.Context, topics []string, options ...DeleteTopicsAdminOption) (result []TopicResult, err error) {
+
+	if a.verifyAdminClient() == false {
+		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	}
+
 	cTopics := make([]*C.rd_kafka_DeleteTopic_t, len(topics))
 
 	cErrstrSize := C.size_t(512)
@@ -904,6 +946,11 @@ func (a *AdminClient) DeleteTopics(ctx context.Context, topics []string, options
 
 // CreatePartitions creates additional partitions for topics.
 func (a *AdminClient) CreatePartitions(ctx context.Context, partitions []PartitionsSpecification, options ...CreatePartitionsAdminOption) (result []TopicResult, err error) {
+
+	if a.verifyAdminClient() == false {
+		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	}
+
 	cParts := make([]*C.rd_kafka_NewPartitions_t, len(partitions))
 
 	cErrstrSize := C.size_t(512)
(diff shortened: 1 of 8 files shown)


kafka/producer.go
```diff
@@ -148,10 +151,21 @@ func (p *Producer) gethandle() *handle {
 	return &p.handle
 }
 
+// verifyProducer verifies if Producer can be used
+func (p *Producer) verifyProducer() bool {
+	if p == nil || p.isClosed {
+		return false
```
vivek-ng, line 157: This is not thread safe and I don't think it's worth making it thread safe. What's the use case for this? I understand it's useful to know if the producer is closed but I don't see a real benefit to this.

emasab: @vivek-ng this is for cases where the Producer, Consumer or AdminClient have been closed, and are used after being closed, or a nil instance returned by New* is used without checking the err. It happens, and we want to give a meaningful error instead of a Go panic or a C segfault.

vivek-ng: @emasab that makes sense. Thanks for the context, this was super helpful 👍. Should we just use `if p == nil` for checking valid producer state (again, I'm not sure if this is thread safe either)? I'm concerned that checking `p.isClosed` is not thread safe  (same issue with consumer).

@PrasanthV454 I'd recommend adding detailed description to the PR explaining the reason for the change and what issue/problem this PR aims to solve (like what Emanuele mentioned in the comment) especially when you're contributing to open source projects used by many companies across the world. This will help people (maybe 2-3 years down the line) to debug issues faster when they look at PR and try to understand why this change was made. Please look at @milindl's PRs as examples on how to write good PR descriptions :)

kafka/adminapi.go
```diff
@@ -560,6 +570,10 @@ type DeleteACLsResult = DescribeACLsResult
 // The returned result event is checked for errors its error is returned if set.
 func (a *AdminClient) waitResult(ctx context.Context, cQueue *C.rd_kafka_queue_t, cEventType C.rd_kafka_event_type_t) (rkev *C.rd_kafka_event_t, err error) {
 
+	if a.verifyAdminClient() == false {
```
milindl, line 573: Maybe we need this check only for the public methods and not the private ones (since we can control when the private ones are called and take care never to call one after closure.) Public methods on the other hand are called by the external users.

kafka/consumer.go
```diff
@@ -47,13 +47,23 @@ type Consumer struct {
...
+
+	isClosed bool
 }
 
 // Strings returns a human readable name for a Consumer instance
 func (c *Consumer) String() string {
 	return c.handle.String()
 }
 
+// verifyConsumer verifies if Producer can be used
```
milindl, line 59: Producer -> Consumer change in comment

kafka/consumer_test.go
```diff
@@ -46,13 +51,38 @@ func TestConsumerAPIs(t *testing.T) {
 		t.Fatalf("%s", err)
 	}
 
+	closedConsumer, err := NewConsumer(&ConfigMap{
+		"group.id":                 "gotest",
+		"socket.timeout.ms":        10,
+		"session.timeout.ms":       10,
+		"enable.auto.offset.store": false, // permit StoreOffsets()
```
milindl, line 58: Keep only the required config options, the ones for example 'enable.auto.offset.store' is not needed for this particular consumer since we don't deal with any offsets at all.

kafka/consumer_test.go
```diff
@@ -46,13 +51,38 @@ func TestConsumerAPIs(t *testing.T) {
...
 	t.Logf("Consumer %s", c)
+	t.Logf("Consumer %s", closedConsumer)
+	t.Logf("Consumer %s", nilConsumer)
 
 	err = c.Subscribe("gotest", nil)
 	if err != nil {
 		t.Errorf("Subscribe failed: %s", err)
 	}
 
+	err = closedConsumer.Subscribe("gotest", nil)
```
milindl, line 76: Rather than repeating this everywhere, perhaps we can have a method like `testConsumerAPIsErr(consumer, expectedErr)` and from TestConsumerAPIs, we just call 
```
testConsumerAPIsWithErr(c, nil)
testConsumerAPIsWithErr(closedConsumer, stateErr)
.... and so on
```

I understand that for the case where err is nil (the "valid" consumer), the handling may be different. If that is so, the method testConsumerAPIsErr can at least be used for the nil and the closed consumer, because their handling should be the same

kafka/producer_test.go
```diff
@@ -179,6 +179,41 @@ func TestProducerAPIs(t *testing.T) {
 	if offsets != nil {
 		t.Errorf("OffsetsForTimes() failed but returned non-nil Offsets: %s\n", offsets)
 	}
+
+	p.Close()
+	err = p.Produce(&Message{TopicPartition: TopicPartition{Topic: &topic2, Partition: 0},
+		Value: []byte("Producer Validation"), Key: []byte("This is my key")},
+		nil)
+	expectedErr := Error{code: ErrState, str: "not a valid Producer state."}
+	if err != expectedErr {
```
milindl, line 188: The if-else clauses can be made into a function, will clean this up quite a bit.

kafka/producer_test.go
```diff
@@ -320,6 +374,37 @@ func TestProducerOAuthBearerConfig(t *testing.T) {
 	}
 
 	p.Close()
+	err = p.SetOAuthBearerToken(OAuthBearerToken{
```
milindl, line 377: For a lot of the producer tests, the code looks like

```
func() {
 // 1. handling for non-nil, non-closed producer
 // 2. handling for closed producer
 // 3. handling for nil producer
}
```

in most of these cases, while 1 is too different from 2 and 3, 2 and 3 are mostly the same except the value of the producer. It would be better if they can be made into a separate function and called with different value of producer to cut down on code duplication.

kafka/adminapi.go
```diff
@@ -639,6 +657,10 @@ func (a *AdminClient) cToTopicResults(cTopicRes **C.rd_kafka_topic_result_t, cCn
 // cConfigResourceToResult converts a C ConfigResource result array to Go ConfigResourceResult
 func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t, cCnt C.size_t) (result []ConfigResourceResult, err error) {
 
+	if a.verifyAdminClient() == false {
```
milindl, line 660: Minor - for this and everywhere - maybe `!verifyClient`is better than an explicit comparison to `false`.

kafka/adminapi_test.go
```diff
@@ -30,6 +30,11 @@ func TestAdminAPIWithDefaultValue(t *testing.T) {
 		t.Skipf("Missing testconf.json")
 	}
 
+	stateErr := Error{
```
milindl, line 33: Why don't we use `newErrorFromString` method here to create the error? Would be good since any changes in that method would otherwise require changing this value, too.

kafka/producer_test.go
milindl: Same goes for the changes in TestProducerAPIs

milindl (changes requested):
Some minor changes suggested for the implementation code.
Some changes suggested for the tests to cut down on duplication.

kafka/producer.go
```diff
@@ -148,10 +151,21 @@ func (p *Producer) gethandle() *handle {
 	return &p.handle
 }
 
+// verifyProducer verifies if Producer can be used
+func (p *Producer) verifyProducer() bool {
+	if p == nil || p.isClosed {
```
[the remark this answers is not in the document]
milindl: Nope, it's probably fine, you can ignore the minor stuff for if you feel like

kafka/consumer_test.go
PrasanthV454: We did use Store Offsets method on closedConsumer and nilConsumer.

kafka/producer_test.go
PrasanthV454: got it.

kafka/adminapi.go
```diff
@@ -76,6 +76,16 @@ import "C"
 type AdminClient struct {
 	handle    *handle
 	isDerived bool // Derived from existing client handle
+
+	isClosed bool // to check if Admin Client is closed or not.
+}
+
+// verifyAdminClient verifies if AdminClient can be used
+func (a *AdminClient) verifyAdminClient() bool {
+	if a == nil || a.isClosed {
```
edenhill, line 85: Can `a == nil` even happen? Sounds like a usage error:
```go
a := NewAdminClient(..)
... other stuff..

a = nil

a.DoThing()  // invalid, not something we need to support
```

kafka/adminapi.go
```diff
@@ -560,6 +570,10 @@ type DeleteACLsResult = DescribeACLsResult
 // The returned result event is checked for errors its error is returned if set.
 func (a *AdminClient) waitResult(ctx context.Context, cQueue *C.rd_kafka_queue_t, cEventType C.rd_kafka_event_type_t) (rkev *C.rd_kafka_event_t, err error) {
 
+	if a.verifyAdminClient() == false {
+		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
```
edenhill, line 574: Instead of repeating this in all uses, suggest having verifyAdminClient return this error.
And perhaps also rephrase it, the user won't know what a valid or invalid adminClient state is, perhaps say:
"Operation not allowed on closed client"

kafka/adminapi_test.go
```diff
@@ -64,6 +77,32 @@ func TestAdminAPIWithDefaultValue(t *testing.T) {
...
+		ctx,
+		[]TopicSpecification{
+			{
+				Topic:             topic,
+				NumPartitions:     -1,
+				ReplicationFactor: -1,
+			},
+		})
+	if err != stateErr {
+		t.Errorf("AdminClient should have throwed : %s", stateErr)
```
edenhill, line 90: Also print what it throwed

kafka/consumer.go
```diff
@@ -47,13 +47,23 @@ type Consumer struct {
 	rebalanceCb        RebalanceCb
 	appReassigned      bool
 	appRebalanceEnable bool // SerializerConfig setting
+
+	isClosed bool
```
edenhill, line 51: Remove empty line

kafka/consumer.go
```diff
@@ -47,13 +47,23 @@ type Consumer struct {
...
 }
 
 // Strings returns a human readable name for a Consumer instance
 func (c *Consumer) String() string {
 	return c.handle.String()
 }
 
+// verifyConsumer verifies if Producer can be used
+func (c *Consumer) verifyConsumer() bool {
```
edenhill, line 60: Since these functions are all the same across client types, I recommend putting this on the handle type instead and rename it to verifyClient

kafka/consumer_test.go
```diff
@@ -46,13 +51,38 @@ func TestConsumerAPIs(t *testing.T) {
...
+	})
+	if err != nil {
+		t.Fatalf("%s", err)
+	}
+	closedConsumer.Close()
+
+	var nilConsumer *Consumer = nil
+
 	t.Logf("Consumer %s", c)
+	t.Logf("Consumer %s", closedConsumer)
```
edenhill, line 68: For these to be helpful you need to have the log differentiate them: "closedConsumer %s" and "nilConsumer %s" (still don't think we need to support nils).

kafka/consumer_test.go
edenhill: I agree

edenhill (changes requested):
Consolidate code into common handle type.
Have it return an error object to avoid all the duplication.
Lots of duplication in tests now.
Make sure that no existing tests are changed (the code may change, but not the behaviour - in that case add a new test).
I don't think the check for nil is meaningful - if unsure, check what the golang stdlib does.
Add a CHANGELOG.md entry.

kafka/adminapi.go
milindl: I think what someone actually did was this -  
```
a, err := NewAdminClient(..)
// err is not nil, a is nil, now if I use a without checking err, it is a panic.
a.DoThing()
```

PrasanthV454: @milindl I think it's impossible to have a=nil when NewAdminClient has been called.

The only way I see this is usage error by assigning a = nil.

milindl: There are a couple of places in NewAdminClient where we return `nil, err` from, all those would indeed cause this.

PrasanthV454: Sorry, mistook it thought err is also nil. Agreed.

edenhill: Right, not checking the err return is a programming error and a huge golang antipattern, so not something we need to explicitly handle.

PrasanthV454: Moving ahead without nil check. 

cc: @edenhill @milindl

Issue #876, opened 2022-10-27
Panic when creating Schema Registry Client with SSL Ca certificate


HansK-p:
Description
===========
When  running schemaregistry.NewClient() with the default config + SslCaLocation set, this results in an panic due to a nil pointer. It might be this isn't supposed to work, but I assume the application shouldn't panic.

How to reproduce
================
<pre>
package main

import (
        "github.com/confluentinc/confluent-kafka-go/schemaregistry"
)

func main() {
        schemaRegistryConfig := schemaregistry.NewConfig("https://schemaregistry.test.com")
        schemaRegistryConfig.SslCaLocation = "certs/gandi.pem"
        _, _ = schemaregistry.NewClient(schemaRegistryConfig)
}
==

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x578803]

goroutine 1 [running]:
crypto/x509.(*CertPool).addCertFunc(...)
	/usr/lib/go-1.19/src/crypto/x509/cert_pool.go:189
crypto/x509.(*CertPool).AppendCertsFromPEM(0x0, {0xc0000ec900?, 0x7fa521729fff?, 0x7fa5203d8bb8?})
	/usr/lib/go-1.19/src/crypto/x509/cert_pool.go:227 +0x263
github.com/confluentinc/confluent-kafka-go/schemaregistry.configureTLS(0x0?, 0xc000108000)
	/home/hkno/go/pkg/mod/github.com/confluentinc/confluent-kafka-go@v1.9.2/schemaregistry/rest_service.go:175 +0x2a8
github.com/confluentinc/confluent-kafka-go/schemaregistry.configureTransport(0xc0000c5ec0)
	/home/hkno/go/pkg/mod/github.com/confluentinc/confluent-kafka-go@v1.9.2/schemaregistry/rest_service.go:192 +0x3b
github.com/confluentinc/confluent-kafka-go/schemaregistry.newRestService(0xc0000c5ec0)
	/home/hkno/go/pkg/mod/github.com/confluentinc/confluent-kafka-go@v1.9.2/schemaregistry/rest_service.go:126 +0x125
github.com/confluentinc/confluent-kafka-go/schemaregistry.NewClient(0xc0000c5ec0)
[...]

PR #856, opened 2022-09-07
fix: initialise schema-registry tlsConf.RootCAs value (invalid memory address or nil pointer)

ju-popov:
```
panic: runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation code=0x2 addr=0x20 pc=0x100fcd0e0]

goroutine 1 [running]:
crypto/x509.(*CertPool).addCertFunc(...)
	/opt/homebrew/Cellar/go/1.19.1/libexec/src/crypto/x509/cert_pool.go:189
crypto/x509.(*CertPool).AppendCertsFromPEM(0x0, {0x1400047a000?, 0x1400031f998?, 0x100dabf7c?})
	/opt/homebrew/Cellar/go/1.19.1/libexec/src/crypto/x509/cert_pool.go:227 +0x260
github.com/confluentinc/confluent-kafka-go/schemaregistry.configureTLS(0x1400046da01?, 0x140001cd500)
	/Users/ju/github.com/sumup/confluent-kafka-go/schemaregistry/rest_service.go:176 +0x1f0
github.com/confluentinc/confluent-kafka-go/schemaregistry.configureTransport(0x1400031fbf8)
	/Users/ju/github.com/sumup/confluent-kafka-go/schemaregistry/rest_service.go:190 +0x38
github.com/confluentinc/confluent-kafka-go/schemaregistry.newRestService(0x1400031fbf8)
	/Users/ju/github.com/sumup/confluent-kafka-go/schemaregistry/rest_service.go:126 +0x128
github.com/confluentinc/confluent-kafka-go/schemaregistry.NewClient(0x1400031fbf8)
	/Users/ju/github.com/sumup/confluent-kafka-go/schemaregistry/schemaregistry_client.go:229 +0x15c
```

PR #878, opened 2022-10-27
Ensure that Rest Service RootCA is set before use in TLS config

HansK-p:
This PR should solve #876. It's been lightly tested.


schemaregistry/rest_service.go
```diff
@@ -172,6 +173,9 @@ func configureTLS(conf *Config, tlsConf *tls.Config) error {
 		if err != nil {
 			return err
 		}
+		if tlsConf.RootCAs == nil {
```
edenhill, line 176: Good find!

@emasab @rayokota From what I can see this code never worked, and was thus never tested. Can we add a test case?

Also, the tls docs says RootCAs will default (if nil) to the system CAs, which makes me wonder if the suggested fix is correct, or if AppendCertsFromPEM.. call isn't correct?

milindl:
#856 is also fixing this. I think we should change the needless check for `err != nil` just after `AppendCertsFromPEM`, and rather, check the return value of `AppendCertsFromPEM` instead. 

I am also unsure if we should check `tlsConf.RootCAs == nil` first, because this method is an internal method called once while making a NewClient and it's supposed to initialize/populate the tlsConf.

edenhill:
I agree that we should also have error handling for the AppendCerts.. call

anchitj (approved):

PR #880, opened 2022-10-28
remove error return for consumer close

clyang82:
there is no any errors for `Close` so removing it.

Signed-off-by: clyang82 <chuyang@redhat.com>


clyang82:
That makes sense to keep the API backward compatibility. We do not want to break it. Thanks @edenhill for your comment.

Issue #884, opened 2022-11-02
Manually committing offset of messages causing consuming same message again


VipulZopSmart:
Description
===========
I've written a sample consumer app that's running indefinitely in which i've disabled the auto commit. I'm trying to manually commit the messages on the basis of booleans(created 2 booleans isCommit, isContinue) on the basis of which i decide if i want to commit the consumed message offset or not. I've tried to use https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/consumer.go#L249
https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/consumer.go#L268 by passing the message offsets explicitly but what's happening it's consuming the same message almost 4 times when it comes at the situation where i don't want to commit the offset. Please provide the sample code below.

```go
func(k *Kafka)SubscribeWithCommit(f pubsub.CommitFunc)(message,error){
isCommit, isContinue := f(msg)    // where msg is the consumed message from the topic.
// f is func passed through actual app
		if isCommit {
			if k.config.DisableAutoCommit {
				fmt.Printf("message %v , offset %d , partition %d\n", msg.Value, msg.Offset, msg.Partition)
				x := pubsub.TopicPartition{
					Topic:     msg.Topic,
					Partition: msg.Partition,
					Offset:    msg.Offset,
					Metadata:  nil,
				}

				k.CommitOffset(x)
}



func (k *Kafka) CommitOffset(offsets pubsub.TopicPartition) {
	fmt.Printf("Committed Offset, msg offset %d\n", offsets.Offset)
	go k.Consumer.Commit()
}

// f func passed by app
message, err := ctx.SubscribeWithCommit(func(message *pubsub.Message) (bool, bool) {
		count++
		fmt.Println(message)
		ctx.Logger.Infof("Consumed %v message(s), offset: %v, topic: %v", count, message.Offset, message.Topic)

		if count <= 2 {
			return true, true
		}

		if count <= 5 {
			return false, true
		}

		return false, false
	})



Output:
&{0 test-topic 0 100  {"ID":"3","Name":"Andrew","Email":"abc@gmail.com"} map[test1:test2] <nil>}
INFO [16:08:08]  Consumed 1 message(s), offset: 100, topic: test-topic
  CorrelationId: 70bcf4a45a9a11ed8dac1e00230ed139 (Memory: 29662688 GoRoutines: 8) 
[...]

PR #891, opened 2022-11-07
[consumer] Add example for rebalance callback

### LINKED ISSUE #886: rebalanceCb example [enhancement]
we don't have any good examples of using a rebalance callback - we need one, with plenty of commenting to explain what is going on and the possibilities. if the erroneous state error case is possible with the go client when storing / committing offsets, we should also document this here as well.

milindl:
Adds a consumer example for usage of rebalance cb. Includes enable.auto.commit and enable.auto.offset.store each as optional fields to set/unset, and how to write the callback in each of those cases.
Only for EAGER rebalancing strategy, for example, range or roundrobin.

Fixes #886 
Should help #884


VipulZopSmart:
Few questions @milindl 
- The rebalance callback will be triggered every time as it would be polling for messages, in that case there would be frequent rebalancing of consumer group. And as far as i know frequent rebalancing create issues in consumption from the assigned partitions.
- As a part of this [ticket](https://github.com/confluentinc/confluent-kafka-go/issues/884), you would see i've also tried to use the same sort of rebalance callback func in my consumer application that you're using in this example but i've got the remark from matt: `you seem to be re-subscribing to the group every time you want to retrieve a message. this will cause a group rebalance (extremely expensive, and affects all consumers in the group).`
- If i want to handle the offsets committing manually, there are issues like(if i don't commit it's not going for the next message consumption it's still consuming the same message) how can we achieve this?

milindl:
1. The callback is not triggered every time we poll for messages, it _may_ be created on Poll the moment after a rebalance. For example, say there are two long running clients A and B as a part of the consumer group "mygroup". They are calling Poll. The callback will not be triggered. Now, when a client C joins this group, and subscribes to some topic. Now, when A, B and C call Poll, then the callback will be triggered for all three clients once. After that trigger one time, Poll will again only fetch events/messages, till a client joins/leaves.

2. In the linked code, it seems like `subscribeMessage` is called every you need to fetch a message. This is because we are returning from the for loop when we get a message. This means, every time `subscribeMessage` is called, `SubscribeTopics` will be called, which is very expensive as it triggers a rebalance.
This is different from the example, in which I call `SubscribeTopics` exactly once, and then let the for loop run till I want to terminate the entire client.
In general, you should try to do something similar -- after subscribing to the topics you require, let the for loop run till the client needs to be terminated. You can use a function to process the message rather than returning it, kind of like `processEvent` in my example. Or else, if your processing is time taking, you can even push events to a channel and use another goroutine to process them. But the long running for loop should not be broken every message.

3. Looking at your attached code, you should be committing `Offset:    msg.Offset + 1,` rather than `Offset:    msg.Offset,`. The explanation is same as what you can see in `maybeStoreOffset` of my example.

edenhill:
Great work on this @milindl

I feel that combining all these different cases in the same example file makes it a bit daunting, especially for new users I imagine.
How about we separate each case into its own example to keep things as minimal and simple as possible?

milindl:
Thanks @edenhill!
Sure, I can do that - 
Which combinations do we need?
1. auto store = off, auto commit = on
2. auto store = on, auto commit = off
3. auto store = off, auto commit = off 

Or do we need one for a case where both of them are "on", and we're just doing some logging stuff in the rebalance cb?

edenhill:
Anything with auto commit enabled probably does not need a rebalance callback, unless they're writing offsets outside of Kafka - which I'm not sure we need to provide an example for?

And auto store is not really a thing that should affect the rebalance callback but rather the consume loop.

If people agree with that we can probably just bring this down to one example: disabled auto commits, right?

edenhill:
There is a strong case for having an enable.auto.offset.store=false example though, but that's not directly related to the rebalance callback so it should be its own example.

milindl:
Noted, I'll focus on just the auto.commit.offset=off example for now.

An example with enable.auto.offset.store=false already exists at examples/consumer_example/consumer_example.go (which is also, seemingly, the main consumer example)
I think I can add a few more comments to it, since it doesn't mention it in the file itself how we can get an atleast-once-guarantee by disabling this property.

edenhill:
Good plan :+1:

milindl:
Made the changes discussed above.

examples/consumer_example/consumer_example.go
```diff
@@ -50,10 +50,14 @@ func main() {
 		// when using localhost brokers on OSX, since the OSX resolver
 		// will return the IPv6 addresses first.
 		// You typically don't need to specify this configuration property.
-		"broker.address.family":    "v4",
-		"group.id":                 group,
-		"session.timeout.ms":       6000,
-		"auto.offset.reset":        "earliest",
+		"broker.address.family": "v4",
+		"group.id":              group,
+		"session.timeout.ms":    6000,
```
edenhill, line 55: Don't need to set this property

examples/consumer_example/consumer_example.go
```diff
@@ -81,11 +85,24 @@ func main() {
...
 			switch e := ev.(type) {
 			case *kafka.Message:
+				// Process the message received.
 				fmt.Printf("%% Message on %s:\n%s\n",
 					e.TopicPartition, string(e.Value))
 				if e.Headers != nil {
 					fmt.Printf("%% Headers: %v\n", e.Headers)
 				}
+
+				// We can store the offsets of the messages manually or let
```
edenhill, line 95: I think we also need to explain what storing the offset means, because this might give the idea that storing == committing.

examples/consumer_rebalance_example/consumer_rebalance_example.go
```diff
@@ -0,0 +1,206 @@
...
+			fmt.Printf("%% Caught signal %v: terminating\n", sig)
+			run = false
+		default:
+			ev := c.Poll(100)
+			if ev == nil {
+				continue
+			}
+
+			if err = processEvent(c, ev); err != nil {
+				fmt.Fprintf(os.Stderr, "Failed to process event: %s\n", err)
```
edenhill, line 80: I'd include the event too, e.g:
```suggestion
				fmt.Fprintf(os.Stderr, "Failed to process event: %v: %s\n", ev, err)
```

examples/consumer_rebalance_example/consumer_rebalance_example.go
```diff
@@ -0,0 +1,206 @@
...
+// rebalanceCallback is called on each group rebalance to assign additional
+// partitions, or remove existing partitions, from the consumer's current
+// assignment.
+//
+// The application may use this optional callback to inspect the assignment,
+// alter the initial start offset (the .Offset field of each assigned partition),
+// and read/write offsets to commit to an alternative store outside of Kafka.
+func rebalanceCallback(c *kafka.Consumer, event kafka.Event) error {
+	switch ev := event.(type) {
+	case kafka.AssignedPartitions:
```
edenhill, line 154: This needs to support incremental rebalancing too, see cooperative_consumer_example.go:

if the rebalance protocol is EAGER we use standard Assign() / Unassign(), otherwise Incremental*(). Also pay attention to AssignmentLost (which should cause an Unassign()).

edenhill (changes requested):

examples/consumer_example/consumer_example.go
milindl: Done, tried to explain the difference between storing and committing.

Just wanted to confirm if it's okay to use this line or if this is the recommended way of doing things:
"We generally follow a two-phase process, of storing offsets, and then committing them"

examples/consumer_rebalance_example/consumer_rebalance_example.go
milindl: Done for the most part - I have separated the code for Eager and Cooperative protocols despite there being a very large amount of repetition. The reason for this is that this being an example, people are likely to want to just copy and paste a bunch of stuff. 
And since we can't mix protocols, they're likely to want either just Eager, or just Cooperative, and there's no need to check on every rebalance.

> Also pay attention to AssignmentLost (which should cause an Unassign()).
Do we need to call this manually in both cases or is just the comment okay?

examples/consumer_rebalance_example/consumer_rebalance_example.go
```diff
@@ -0,0 +1,303 @@
...
+	// can happen if we haven't stored anything since the last commit.
+	// While this will never happen for this example since we call this method
+	// per-message, and thus, always have something to commit, the error
+	// handling is illustrative of how to handle it in cases we call Commit()
+	// in another way, for example, every N seconds.
+	if err != nil && err.(kafka.Error).Code() != kafka.ErrNoOffset {
+		return err
+	}
+
+	fmt.Printf("%% Commited offsets to Kafka: %v\n", commitedOffsets)
```
edenhill, line 161: Committed

examples/consumer_rebalance_example/consumer_rebalance_example.go
```diff
@@ -0,0 +1,303 @@
...
+// Note: as a reader of this example, it is likely that all your consumers will
+// use the same rebalance protocol (either EAGER or COOPERATIVE). In that case,
+// you only need to include code from that particular callback, and ignore the
+// other one.
+func rebalanceCallback(c *kafka.Consumer, event kafka.Event) error {
+	protocol := c.GetRebalanceProtocol()
+	if protocol == "EAGER" {
+		return rebalanceCallbackEager(c, event)
+	}
+	if protocol == "COOPERATIVE" {
```
edenhill, line 182: else if

examples/consumer_rebalance_example/consumer_rebalance_example.go
edenhill: On AssignmentLost you will not be able to commit offsets.

examples/consumer_rebalance_example/consumer_rebalance_example.go
```diff
@@ -0,0 +1,303 @@
...
+func rebalanceCallbackEager(c *kafka.Consumer, event kafka.Event) error {
+	switch ev := event.(type) {
+	case kafka.AssignedPartitions:
+		fmt.Printf("%% %s rebalance: %d new partition(s) assigned: %v\n",
+			c.GetRebalanceProtocol(), len(ev.Partitions), ev.Partitions)
+
+		// The application may update the start .Offset of each assigned
+		// partition and then call Assign(). It is optional to call Assign
+		// in case the application is not modifying any start .Offsets. In
+		// the case we don't, the library takes care of it.
```
edenhill, line 199: "consumer", not "library"

edenhill (changes requested):

examples/consumer_rebalance_example/consumer_rebalance_example.go
milindl: Understood, mentioned that in a comment while handling partitions revoked.

examples/consumer_rebalance_example/consumer_rebalance_example.go
```diff
@@ -0,0 +1,302 @@
...
+	// in another way, for example, every N seconds.
+	if err != nil && err.(kafka.Error).Code() != kafka.ErrNoOffset {
+		return err
+	}
+
+	fmt.Printf("%% Committed offsets to Kafka: %v\n", committedOffsets)
+	return nil
+}
+
+// rebalanceCallback is called on each group rebalance to assign additional
```
PrasanthV454, line 165: I think it's good to list down possible scenarios when rebalance happens. 

1. Consumer Group subscribes to new topics.
2. New partition added.

possible cases like this.

milindl:
Closing with PR on non-forked branch #942

examples/consumer_rebalance_example/consumer_rebalance_example.go
milindl: Done in the new PR

examples/consumer_rebalance_example/consumer_rebalance_example.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	sigchan := make(chan os.Signal, 1)
+	signal.Notify(sigchan, os.Interrupt)
+
+	c, err := kafka.NewConsumer(&kafka.ConfigMap{
+		"bootstrap.servers": bootstrapServers,
+		// Consumer group ID
+		"group.id": group,
+		// Start reading from the first message of each assigned
```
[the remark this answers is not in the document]
milindl: About this -- I think we should retain the duplicate, because the person reading the example is likely just starting out, and it's better if the relevant information is in one place rather than them having to jump around files

PR #894, opened 2022-11-09
Deprecate channel based clients

milindl:
Adds Deprecated: to the godoc for ProduceChannel, and consumer's Event method.

Adds BUG: to the godoc for Flush and Len to emphasize they don't play well with ProduceChannel.


milindl:
Does this need a changelog.md entry despite the fact that these were deprecated earlier?

kafka/producer.go
```diff
@@ -317,21 +317,25 @@ func (p *Producer) Logs() chan LogEvent {
...
+// of Produce.
+// Flush() and Len() are not guaranteed to work correctly with ProduceChannel.
 func (p *Producer) ProduceChannel() chan *Message {
 	return p.produceChannel
 }
 
 // Len returns the number of messages and requests waiting to be transmitted to the broker
 // as well as delivery reports queued for the application.
-// Includes messages on ProduceChannel.
+// BUG: Tries to include messages on ProduceChannel, but it's not guaranteed to work.
```
edenhill, line 330:
```suggestion
// BUG: Tries to include messages on ProduceChannel, but it's not guaranteed to be reliable.
```

kafka/consumer.go
```diff
@@ -346,6 +346,9 @@ func (c *Consumer) Poll(timeoutMs int) (event Event) {
 }
 
 // Events returns the Events channel (if enabled)
+//
+// Deprecated: Events (channel based consumer) is deprecated in favour
+// of Poll.
```
edenhill, line 351:
```suggestion
// of Poll().
```

kafka/producer.go
```diff
@@ -317,21 +317,25 @@ func (p *Producer) Logs() chan LogEvent {
 }
 
 // ProduceChannel returns the produce *Message channel (write)
+//
+// Deprecated: ProduceChannel (channel based producer) is deprecated in favour
+// of Produce.
```
edenhill, line 322:
```suggestion
// of Produce().
```

kafka/producer.go
```diff
@@ -317,21 +317,25 @@ func (p *Producer) Logs() chan LogEvent {
...
+// Flush() and Len() are not guaranteed to work correctly with ProduceChannel.
```
edenhill, line 323:
```suggestion
// Flush() and Len() are not guaranteed to be reliable with ProduceChannel.
```

kafka/producer.go
```diff
@@ -317,21 +317,25 @@ func (p *Producer) Logs() chan LogEvent {
...
+// BUG: Tries to include messages on ProduceChannel, but it's not guaranteed to work.
 func (p *Producer) Len() int {
 	return len(p.produceChannel) + len(p.events) + int(C.rd_kafka_outq_len(p.handle.rk))
 }
 
 // Flush and wait for outstanding messages and requests to complete delivery.
-// Includes messages on ProduceChannel.
 // Runs until value reaches zero or on timeoutMs.
 // Returns the number of outstanding events still un-flushed.
+// BUG: Tries to include messages on ProduceChannel, but it's not guaranteed to work.
```
edenhill, line 338:
```suggestion
// BUG: Tries to include messages on ProduceChannel, but it's not guaranteed to be reliable.
```

edenhill (changes requested):
Please also add deprecation warnings to the examples README and the deprecated examples themselves, as needed

edenhill:
Add a CHANGELOG entry to formally deprecate the channel based producer and consumer.
This will go well with our semver bump to 2.0

milindl:
Emanuele has already moved the channel based examples into a legacy folder inside examples/ and made the required changes to legacy/README.md
The only example which needed a major change was go-kafkacat, which has been rewritten to not use channels. 

Also - just to clarify this, Producer's Events() channel/deliveryChan in Produce() method is not being deprecated, right? Only ProduceChannel of the producer, and the Events() channel of the consumer is being deprecated?

edenhill:
> Also - just to clarify this, Producer's Events() channel/deliveryChan in Produce() method is not being deprecated, right? Only ProduceChannel of the producer, and the Events() channel of the consumer is being deprecated?

Yep, that's correct

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Confluent's Golang client for Apache Kafka
 
+# v2.0.0
+
+This is a maintenance release:
+
+* Channel based producer (`.ProduceChannel()`) and channel based consumer (`.Events()`) are deprecated.
```
edenhill, line 7: Since both Producer and Consumer has an Events() (iirc), I think it might be good to actually say something like
```suggestion
* Channel based producer (Producer `ProduceChannel()`) and channel based consumer (Consumer `Events()`) are deprecated.
```

edenhill (changes requested):

CHANGELOG.md
```diff
@@ -1,13 +1,13 @@
 # Confluent's Golang client for Apache Kafka
 
-## v1.9.3
+# v2.0.0
```
edenhill, line 3: Use level-2 heading, there's already a level-1 above.

CHANGELOG.md
```diff
@@ -1,13 +1,13 @@
...
 
 This is a maintenance release:
 
  * Added SetSaslCredentials. This new method (on the Producer, Consumer, and
    AdminClient) allows modifying the stored SASL PLAIN/SCRAM credentials that
    will be used for subsequent (new) connections to a broker.
-
+ * Channel based producer (Producer `ProduceChannel()`) and channel based consumer (Consumer `Events()`) are deprecated.
```
edenhill, line 10: Try to keep below 80 columns in width to make it palatable on terminals and GH code reviews.
Keep two newlines to separate from the next release.

edenhill (changes requested):

CHANGELOG.md
milindl: Noted, I'll keep this in mind for the future. Generally, I try to stick to what the file already seems to follow.

edenhill (approved):
LGTM

PR #900, opened 2022-11-17
Add info about arm64 statically built target

okulik:
It's worth noting in the README there's a new statically built Apple silicon image.


edenhill (approved):
LGTM

edenhill:
Thank you!

PR #901, opened 2022-11-18
Bug/producer consumer adminclient validity handling

PrasanthV454:
- Added isClosed new field to Producer struct which acts as flag to know if Producer is open or closed.

- Added new function verifyProducer() function which checks if producer is valid or not by doing null checks and closed checks to avoid seg faults for C-GO or go-panics.

### DIFF
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync/atomic"
 	"time"
 	"unsafe"
 )
@@ -1006,7 +1007,7 @@ func (a *AdminClient) ClusterID(ctx context.Context) (clusterID string, err erro
 func (a *AdminClient) ControllerID(ctx context.Context) (controllerID int32, err error) {
 	err = a.verifyClient()
 	if err != nil {
-		return 0, err
+		return -1, err
 	}
 
 	responseChan := make(chan int32, 1)
@@ -1927,21 +1928,20 @@ func (a *AdminClient) SetSaslCredentials(username, password string) error {
 
 // Close an AdminClient instance.
 func (a *AdminClient) Close() {
-	if a.isClosed {
+	var newValue uint32 = 1
+	currentValue := atomic.LoadUint32((*uint32)(unsafe.Pointer(&a.isClosed)))
+	if !atomic.CompareAndSwapUint32((*uint32)(unsafe.Pointer(&a.isClosed)), currentValue, newValue) {
 		return
 	}
-
 	if a.isDerived {
 		// Derived AdminClient needs no cleanup.
 		a.handle = &handle{}
-		a.isClosed = true
 		return
 	}
 
 	a.handle.cleanup()
 
 	C.rd_kafka_destroy(a.handle.rk)
-	a.isClosed = true
 }
 
 // ListConsumerGroups lists the consumer groups available in the cluster.
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -808,11 +808,6 @@ func (a *AdminClient) cToConsumerGroupResults(
 
 // cToTopicResults converts a C topic_result_t array to Go TopicResult list.
 func (a *AdminClient) cToTopicResults(cTopicRes **C.rd_kafka_topic_result_t, cCnt C.size_t) (result []TopicResult, err error) {
-	err = a.verifyClient()
-	if err != nil {
-		return nil, err
-	}
-
 	result = make([]TopicResult, int(cCnt))
 
 	for i := 0; i < int(cCnt); i++ {
@@ -1932,9 +1927,14 @@ func (a *AdminClient) SetSaslCredentials(username, password string) error {
 
 // Close an AdminClient instance.
 func (a *AdminClient) Close() {
+	if a.isClosed {
+		return
+	}
+
 	if a.isDerived {
 		// Derived AdminClient needs no cleanup.
 		a.handle = &handle{}
+		a.isClosed = true
 		return
 	}
 
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -104,16 +104,14 @@ import "C"
 type AdminClient struct {
 	handle    *handle
 	isDerived bool // Derived from existing client handle
-
-	isClosed bool // to check if Admin Client is closed or not.
+	isClosed  bool // to check if Admin Client is closed or not.
 }
 
-// verifyAdminClient verifies if AdminClient can be used
-func (a *AdminClient) verifyAdminClient() bool {
-	if a == nil || a.isClosed {
-		return false
+func (a *AdminClient) verifyClient() error {
+	if a.isClosed {
+		return getOperationNotAllowedErrorForClosedClient()
 	}
-	return true
+	return nil
 }
 
 func durationToMilliseconds(t time.Duration) int {
@@ -734,11 +732,6 @@ type DeleteACLsResult = DescribeACLsResult
 // first.
 // The returned result event is checked for errors its error is returned if set.
 func (a *AdminClient) waitResult(ctx context.Context, cQueue *C.rd_kafka_queue_t, cEventType C.rd_kafka_event_type_t) (rkev *C.rd_kafka_event_t, err error) {
-
-	if a.verifyAdminClient() == false {
-		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
-	}
-
 	resultChan := make(chan *C.rd_kafka_event_t)
 	closeChan := make(chan bool) // never written to, just closed
 
@@ -815,9 +808,9 @@ func (a *AdminClient) cToConsumerGroupResults(
 
 // cToTopicResults converts a C topic_result_t array to Go TopicResult list.
 func (a *AdminClient) cToTopicResults(cTopicRes **C.rd_kafka_topic_result_t, cCnt C.size_t) (result []TopicResult, err error) {
-
-	if a.verifyAdminClient() == false {
-		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	err = a.verifyClient()
+	if err != nil {
+		return nil, err
 	}
 
 	result = make([]TopicResult, int(cCnt))
@@ -942,9 +935,9 @@ func (a *AdminClient) cToErrorList(
 
 // cConfigResourceToResult converts a C ConfigResource result array to Go ConfigResourceResult
 func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t, cCnt C.size_t) (result []ConfigResourceResult, err error) {
-
-	if a.verifyAdminClient() == false {
-		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	err = a.verifyClient()
+	if err != nil {
+		return nil, err
 	}
 
 	result = make([]ConfigResourceResult, int(cCnt))
@@ -979,9 +972,9 @@ func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t
 //
 // Requires broker version >= 0.10.0.
 func (a *AdminClient) ClusterID(ctx context.Context) (clusterID string, err error) {
-
-	if a.verifyAdminClient() == false {
-		return "", newErrorFromString(ErrState, "not a valid adminClient state.")
+	err = a.verifyClient()
+	if err != nil {
+		return "", err
 	}
 
 	responseChan := make(chan *C.char, 1)
@@ -1016,9 +1009,9 @@ func (a *AdminClient) ClusterID(ctx context.Context) (clusterID string, err erro
 //
 // Requires broker version >= 0.10.0.
 func (a *AdminClient) ControllerID(ctx context.Context) (controllerID int32, err error) {
-
-	if a.verifyAdminClient() == false {
-		return 0, newErrorFromString(ErrState, "not a valid adminClient state.")
+	err = a.verifyClient()
+	if err != nil {
+		return 0, err
 	}
 
 	responseChan := make(chan int32, 1)
@@ -1050,9 +1043,9 @@ func (a *AdminClient) ControllerID(ctx context.Context) (controllerID int32, err
 //
 // Note: TopicSpecification is analogous to NewTopic in the Java Topic Admin API.
 func (a *AdminClient) CreateTopics(ctx context.Context, topics []TopicSpecification, options ...CreateTopicsAdminOption) (result []TopicResult, err error) {
-
-	if a.verifyAdminClient() == false {
-		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	err = a.verifyClient()
+	if err != nil {
+		return nil, err
 	}
 
 	cTopics := make([]*C.rd_kafka_NewTopic_t, len(topics))
@@ -1169,9 +1162,9 @@ func (a *AdminClient) CreateTopics(ctx context.Context, topics []TopicSpecificat
 //
 // Requires broker version >= 0.10.1.0
 func (a *AdminClient) DeleteTopics(ctx context.Context, topics []string, options ...DeleteTopicsAdminOption) (result []TopicResult, err error) {
-
-	if a.verifyAdminClient() == false {
-		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	err = a.verifyClient()
+	if err != nil {
+		return nil, err
 	}
 
 	cTopics := make([]*C.rd_kafka_DeleteTopic_t, len(topics))
@@ -1232,9 +1225,9 @@ func (a *AdminClient) DeleteTopics(ctx context.Context, topics []string, options
 
 // CreatePartitions creates additional partitions for topics.
 func (a *AdminClient) CreatePartitions(ctx context.Context, partitions []PartitionsSpecification, options ...CreatePartitionsAdminOption) (result []TopicResult, err error) {
-
-	if a.verifyAdminClient() == false {
-		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	err = a.verifyClient()
+	if err != nil {
+		return nil, err
 	}
 
 	cParts := make([]*C.rd_kafka_NewPartitions_t, len(partitions))
@@ -1327,9 +1320,9 @@ func (a *AdminClient) CreatePartitions(ctx context.Context, partitions []Partiti
 // resource of type ResourceBroker is allowed per call since these
 // resource requests must be sent to the broker specified in the resource.
 func (a *AdminClient) AlterConfigs(ctx context.Context, resources []ConfigResource, options ...AlterConfigsAdminOption) (result []ConfigResourceResult, err error) {
-
-	if a.verifyAdminClient() == false {
-		return nil, newErrorFromString(ErrState, "not a valid adminClient state.")
+	err = a.verifyClient()
+	if err != nil {
+		return nil, err
 	}
 
 	cRes := make([]*C.rd_kafka_ConfigResource_t, len(resources))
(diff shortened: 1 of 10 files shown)


kafka/adminapi.go
```diff
@@ -622,6 +629,10 @@ func (a *AdminClient) waitResult(ctx context.Context, cQueue *C.rd_kafka_queue_t
 
 // cToTopicResults converts a C topic_result_t array to Go TopicResult list.
 func (a *AdminClient) cToTopicResults(cTopicRes **C.rd_kafka_topic_result_t, cCnt C.size_t) (result []TopicResult, err error) {
+	err = a.verifyClient()
```
milindl, line 632: Not required to check this in non-exposed/non-public functions (here and elsewhere)

milindl: All the conversion c->go, go->c methods etc all don't need this

kafka/adminapi_test.go
```diff
@@ -30,26 +30,30 @@ func TestAdminAPIWithDefaultValue(t *testing.T) {
...
 	}
 
-	expDuration, err := time.ParseDuration("30s")
+	adminClient, err := NewAdminClient(&conf)
 	if err != nil {
-		t.Fatalf("Failed to Parse Duration: %s", err)
+		t.Fatalf("Failed to create AdminClient %v", err)
 	}
+	testAdminAPIWithDefaultValue(t, adminClient, nil)
+	testAdminAPIWithDefaultValue(t, adminClient, getOperationNotAllowedErrorForClosedClient())
```
milindl, line 43: Add comment here that `adminClient` is closed by the time we call the second one, because otherwise it requires one to read the whole other function.

kafka/adminapi_test.go
```diff
@@ -58,22 +62,27 @@ func TestAdminAPIWithDefaultValue(t *testing.T) {
...
+	if !a.isClosed {
+		t.Logf("Succeed to create topic %v\n", res)
 	}
-	t.Logf("Succeed to create topic %v\n", res)
 
 	ctx, cancel = context.WithTimeout(context.Background(), expDuration)
 	defer cancel()
-	res, err = adminClient.DeleteTopics(ctx, []string{topic})
+	res, err = a.DeleteTopics(ctx, []string{topic})
 	if err != nil {
```
milindl, line 79: `err != errCheck` here as well?

milindl (commented):
Looks much cleaner than last time

kafka/adminapi.go
```diff
@@ -703,6 +718,11 @@ func (a *AdminClient) ClusterID(ctx context.Context) (clusterID string, err erro
 //
 // Requires broker version >= 0.10.0.
 func (a *AdminClient) ControllerID(ctx context.Context) (controllerID int32, err error) {
+	err = a.verifyClient()
+	if err != nil {
+		return 0, err
```
edenhill, line 723: -1
I believe 0 is a valid broker id

kafka/adminapi.go
```diff
@@ -1555,15 +1641,21 @@ func (a *AdminClient) SetSaslCredentials(username, password string) error {
 
 // Close an AdminClient instance.
 func (a *AdminClient) Close() {
+	if a.isClosed {
+		return
+	}
+
```
edenhill, line 1647: To avoid a race condition this should be a cmp and exchange, or perhaps an increment (if they are atomic in go).
Two different threads calling Close() at the same time now risks the below code to execute in parallel and thus crash.

kafka/adminapi_test.go
```diff
@@ -58,22 +65,32 @@ func TestAdminAPIWithDefaultValue(t *testing.T) {
...
-		adminClient.Close()
-		t.Fatalf("Failed to create topics %v\n", err)
+	if err != errCheck {
+		t.Errorf("CreateTopics() failed to return expected result")
+		if !a.isClosed {
+			a.Close()
+			t.Fatalf("Failed to create topics %v\n", err)
+		}
+	}
+	if !a.isClosed {
```
edenhill, line 75: Why are you checking isClosed here and below?

Also, these tests should preferably only use the public APIs and isClosed is not.

kafka/consumer.go
```diff
@@ -308,25 +365,28 @@ func (c *Consumer) StoreMessage(m *Message) (storedOffsets []TopicPartition, err
...
+	err := c.verifyClient()
+	if err != nil {
+		return err
+	}
 	rkt := c.handle.getRkt(*partition.Topic)
 	cErr := C.rd_kafka_seek(rkt,
 		C.int32_t(partition.Partition),
 		C.int64_t(partition.Offset),
-		C.int(-1))
+		C.int(timeoutMs))
```
edenhill, line 389: these changes are not related to this PR, should not be here

kafka/consumer.go
```diff
@@ -442,6 +506,8 @@ func (c *Consumer) Close() (err error) {
 
 	C.rd_kafka_destroy(c.handle.rk)
 
+	c.isClosed = true
```
edenhill, line 509: same thing here, race condition

kafka/consumer.go
```diff
@@ -575,11 +642,19 @@ func consumerReader(c *Consumer, termChan chan bool) {
...
+		return nil, err
+	}
 	return getMetadata(c, topic, allTopics, timeoutMs)
 }
 
 // QueryWatermarkOffsets queries the broker for the low and high offsets for the given topic and partition.
 func (c *Consumer) QueryWatermarkOffsets(topic string, partition int32, timeoutMs int) (low, high int64, err error) {
+	err = c.verifyClient()
+	if err != nil {
+		return 0, 0, err
```
edenhill, line 656: use -1 for invalid value, 0 is a valid offset.

kafka/consumer.go
```diff
@@ -588,6 +663,10 @@ func (c *Consumer) QueryWatermarkOffsets(topic string, partition int32, timeoutM
 // The low offset is populated every statistics.interval.ms if that value is set.
 // OffsetInvalid will be returned if there is no cached offset for either value.
 func (c *Consumer) GetWatermarkOffsets(topic string, partition int32) (low, high int64, err error) {
+	err = c.verifyClient()
+	if err != nil {
+		return 0, 0, err
```
edenhill, line 668: use -1

kafka/consumer_test.go
```diff
@@ -189,45 +194,50 @@ func TestConsumerAPIs(t *testing.T) {
...
 
 	err = c.Close()
-	if err != nil {
+	if err != errCheck {
 		t.Errorf("Close failed: %s", err)
 	}
 
 	// Tests the SetSaslCredentials call to ensure that the API does not crash.
-	c.SetSaslCredentials("username", "password")
+	if !c.isClosed {
```
edenhill, line 238: Why isn't this tested when closed?

kafka/error.go
```diff
@@ -163,3 +163,7 @@ func getFatalError(H Handle) error {
 func testFatalError(H Handle, code ErrorCode, str string) ErrorCode {
 	return ErrorCode(C.rd_kafka_test_fatal_error(H.gethandle().rk, C.rd_kafka_resp_err_t(code), C.CString(str)))
 }
+
+func getOperationNotAllowedErrorForClosedClient() error {
+	return newErrorFromString(ErrState, "Operation not allowed on closed client.")
```
edenhill, line 168: error strings don't have ending "."

kafka/producer.go
```diff
@@ -368,6 +382,8 @@ func (p *Producer) Close() {
 	p.handle.cleanup()
 
 	C.rd_kafka_destroy(p.handle.rk)
+
+	p.isClosed = true
```
edenhill, line 386: this Close() should also check that isClosed is not already done, and fix that race condition

kafka/producer.go
```diff
@@ -643,12 +664,20 @@ func poller(p *Producer, termChan chan bool) {
...
+	}
 	return getMetadata(p, topic, allTopics, timeoutMs)
 }
 
 // QueryWatermarkOffsets returns the broker's low and high offsets for the given topic
 // and partition.
 func (p *Producer) QueryWatermarkOffsets(topic string, partition int32, timeoutMs int) (low, high int64, err error) {
+	err = p.verifyClient()
+	if err != nil {
+		return 0, 0, err
```
edenhill, line 679: -1

kafka/producer_test.go
```diff
@@ -180,10 +180,78 @@ func TestProducerAPIs(t *testing.T) {
 		t.Errorf("OffsetsForTimes() failed but returned non-nil Offsets: %s\n", offsets)
 	}
 
+	p.Close()
```
edenhill, line 183: The SetSaslCredentials call must be done before Close

edenhill (changes requested):

kafka/adminapi.go
PrasanthV454: got it. will use atomic.CompareAndSwapUint32()

kafka/adminapi_test.go
PrasanthV454: We are performing same tests on closed admin client and open admin client. In case of closed client we don't need to close the client but for other case we were closing the client and producing fatal error. So to differentiate a.isClosed helps.

kafka/consumer.go
PrasanthV454: Reverted.

kafka/adminapi.go
```diff
@@ -638,6 +645,10 @@ func (a *AdminClient) cToTopicResults(cTopicRes **C.rd_kafka_topic_result_t, cCn
 
 // cConfigResourceToResult converts a C ConfigResource result array to Go ConfigResourceResult
 func (a *AdminClient) cConfigResourceToResult(cRes **C.rd_kafka_ConfigResource_t, cCnt C.size_t) (result []ConfigResourceResult, err error) {
+	err = a.verifyClient()
```
milindl, line 648: Remove from this function, it's private to our usage

kafka/adminapi.go
```diff
@@ -1190,11 +1244,20 @@ func (a *AdminClient) SetOAuthBearerToken(oauthBearerToken OAuthBearerToken) err
...
+	err := a.verifyClient()
+	if err != nil {
+		return err
+	}
 	return a.handle.setOAuthBearerTokenFailure(errstr)
 }
 
 // aclBindingToC converts a Go ACLBinding struct to a C rd_kafka_AclBinding_t
 func (a *AdminClient) aclBindingToC(aclBinding *ACLBinding, cErrstr *C.char, cErrstrSize C.size_t) (result *C.rd_kafka_AclBinding_t, err error) {
+	err = a.verifyClient()
```
milindl, line 1256: Remove from this function, it's private to our usage

kafka/adminapi.go
```diff
@@ -1230,6 +1293,11 @@ func (a *AdminClient) aclBindingToC(aclBinding *ACLBinding, cErrstr *C.char, cEr
 
 // aclBindingFilterToC converts a Go ACLBindingFilter struct to a C rd_kafka_AclBindingFilter_t
 func (a *AdminClient) aclBindingFilterToC(aclBindingFilter *ACLBindingFilter, cErrstr *C.char, cErrstrSize C.size_t) (result *C.rd_kafka_AclBindingFilter_t, err error) {
+	err = a.verifyClient()
```
milindl, line 1296: Remove from this function, it's private to our usage

kafka/adminapi.go
```diff
@@ -1291,6 +1359,11 @@ func (a *AdminClient) cToACLBindings(cACLBindings **C.rd_kafka_AclBinding_t, acl
 
 // cToCreateACLResults converts a C acl_result_t array to Go CreateACLResult list.
 func (a *AdminClient) cToCreateACLResults(cCreateAclsRes **C.rd_kafka_acl_result_t, aclCnt C.size_t) (result []CreateACLResult, err error) {
+	err = a.verifyClient()
```
milindl, line 1362: Remove from this function, it's private to our usage

kafka/adminapi.go
```diff
@@ -1555,6 +1642,11 @@ func (a *AdminClient) SetSaslCredentials(username, password string) error {
 
 // Close an AdminClient instance.
 func (a *AdminClient) Close() {
+	var newValue uint32 = 1
+	currentValue := atomic.LoadUint32((*uint32)(unsafe.Pointer(&a.isClosed)))
```
milindl, line 1646: While this code is logical, I'm not sure about the cast to 4 byte int. Size of boolean in go (at least on my machine) is only 1 byte, and not 4. Since there is no CAS for booleans, maybe we can use uint8 as the type of isClosed, with a comment. 
Other alternatives also welcome on this, can't figure out what is the 100% correct way.

kafka/adminapi_test.go
```diff
@@ -58,22 +65,32 @@ func TestAdminAPIWithDefaultValue(t *testing.T) {
...
 
 	ctx, cancel = context.WithTimeout(context.Background(), expDuration)
 	defer cancel()
-	res, err = adminClient.DeleteTopics(ctx, []string{topic})
-	if err != nil {
-		adminClient.Close()
-		t.Fatalf("Failed to delete topic %v, err: %v", topic, err)
+	res, err = a.DeleteTopics(ctx, []string{topic})
+	if err != errCheck {
+		t.Errorf("DeleteTopics() failed to return expected result")
```
milindl, line 83: Print both err and errCheck (in this case, and other cases where expected error is being used)

kafka/consumer_test.go
```diff
@@ -45,11 +44,17 @@ func TestConsumerAPIs(t *testing.T) {
...
+	testConsumerAPIs(t, c, nil)
+
+	// testing the API's with closed consumer.
+	testConsumerAPIs(t, c, getOperationNotAllowedErrorForClosedClient())
+}
+
+func testConsumerAPIs(t *testing.T, c *Consumer, errCheck error) {
+	err := c.Subscribe("gotest", nil)
+	if err != errCheck {
 		t.Errorf("Subscribe failed: %s", err)
```
milindl, line 58: ... Failed to return expected result %v : %v\n", errCheck, err

^ something like this here and in other instances

milindl (changes requested):
Almost done, mostly minor things left.

kafka/consumer_test.go
milindl: It's already been done in case of producer tests, you can use that too

kafka/adminapi_test.go
PrasanthV454: @milindl need a feedback here.

milindl: I can't see a way around checking it, if we should only use the public APIs then we can add an `IsClosed() bool` for that.

kafka/consumer.go
```diff
@@ -494,6 +494,10 @@ func (c *Consumer) ReadMessage(timeout time.Duration) (*Message, error) {
 // Close Consumer instance.
 // The object is no longer usable after this call.
 func (c *Consumer) Close() (err error) {
+	if c.IsClosed() {
+		return getOperationNotAllowedErrorForClosedClient()
+	}
+
 	var newValue uint32 = 1
 	currentValue := atomic.LoadUint32(&c.isClosed)
 	if !atomic.CompareAndSwapUint32(&c.isClosed, currentValue, newValue) {
```
emasab, line 503: You should add `currentValue == 1 || ` here and remove isClosed before because if two threads reach line 501, then one of them reaches line 508 and the second executes line 502 and 503 will execute a compare and swap 1 -> 1 that returns true (the value hasn't changed since when it was read) and continues so it's closing two times

emasab (changes requested):

kafka/adminapi.go
```diff
@@ -103,7 +104,23 @@ import "C"
...
 type AdminClient struct {
 	handle    *handle
-	isDerived bool // Derived from existing client handle
+	isDerived bool   // Derived from existing client handle
+	isClosed  uint32 // to check if Admin Client is closed or not.
+}
+
+// IsClosed returns boolean representing if client is closed or not
+func (a *AdminClient) IsClosed() bool {
+	if a.isClosed == 0 {
```
emasab, line 113:
```suggestion
	return atomic.LoadUint32(&a.isClosed) == 1
```

also for other `IsClosed()`

kafka/consumer_test.go
```diff
@@ -176,10 +179,12 @@ func TestConsumerAPIs(t *testing.T) {
 	// OffsetsForTimes
 	offsets, err := c.OffsetsForTimes([]TopicPartition{{Topic: &topic, Offset: 12345}}, 100)
 	t.Logf("OffsetsForTimes() returned Offsets %s and error %s\n", offsets, err)
-	if err == nil {
+	if !c.IsClosed() && err == nil {
 		t.Errorf("OffsetsForTimes() should have failed\n")
+	} else if c.IsClosed() && err != errCheck {
+		t.Errorf("OffsetsForTimes() should have thrown %v but throwed %v", errCheck, err)
```
emasab, line 185: "threw"

kafka/consumer.go
PrasanthV454: does it still solve our problem? 
because  `currentValue == 1 || !atomic.CompareAndSwapUint32(&c.isClosed, currentValue, newValue)` is still not atomic. Can we instead introduce a mutex to perform this entirely in atomic fashion ?

emasab: Why it's not atomic? If it's already 1 it doesn't need to try to set to 1.

To further simplify, you can replace with:

```suggestion
	if !atomic.CompareAndSwapUint32(&c.isClosed, 0, 1) {
```

without anything else, it's the only case where it has to succeed.

PrasanthV454: got it.

kafka/consumer.go
```diff
@@ -420,6 +491,10 @@ func (c *Consumer) ReadMessage(timeout time.Duration) (*Message, error) {
 // Close Consumer instance.
 // The object is no longer usable after this call.
 func (c *Consumer) Close() (err error) {
+	currentValue := atomic.LoadUint32(&c.isClosed)
```
emasab, line 494: This line is not needed, just leave below:

```suggestion
	if !atomic.CompareAndSwapUint32(&c.isClosed, 0, 1) {
		return getOperationNotAllowedErrorForClosedClient()
	}
```

if isClosed is already 1 it'll return false

kafka/consumer_test.go
milindl: nit: 
thrown -> threw

milindl: nit:
throwed->threw, not thrown to threw. Or so I hope :sweat_smile:

kafka/adminapi.go
```diff
@@ -1841,6 +1913,9 @@ func (a *AdminClient) SetSaslCredentials(username, password string) error {
 
 // Close an AdminClient instance.
 func (a *AdminClient) Close() {
+	if !atomic.CompareAndSwapUint32(&a.isClosed, 0, 1) {
+		return
+	}
 	if a.isDerived {
 		// Derived AdminClient needs no cleanup.
 		a.handle = &handle{}
```
milindl, line 1921: Due to rebase I guess, some methods now lack the guard 
For example, ListConsumerGroups,  DescribeConsumerGroups and so on

milindl (approved):
LGTM but please add the missing prelude in the methods that were added after this PR

emasab (approved):
LGTM

nkostoulas:
hey - I think this is causing some causes when calling `Close` on the Consumer. The Consumer will `Poll` and receive a `RevokedPartitions` event and call the rebalance CB of the application. If the rebalance CB makes a call to `Unassign` this will fail as the consumer is already closed. cc @PrasanthV454 @emasab can you verify?

milindl:
Right, that seems to be the case. I'll file an issue and take a look at this.

nkostoulas:
@milindl is there any reason to call `Unassign` once the Consumer is closed? I figured we could just check in the rebalance CB whether the Consumer has closed (using `IsClosed`) and just ignore the call if that is the case.

milindl:
@nkostoulas in the case of Unassign, it is okay. Especially since Unassign doesn't need to be called explicitly (the library takes care of it if you don't call it), and you can use the check as you mentioned if you do want to call the method. 

However, the issue is bigger, since none of the Commit methods can be called during the closing rebalance, that's problematic for users depending on manual commits. 

Thanks for pointing this out

PR #903, opened 2022-11-24
Add IsTimeout method to Error

milindl:
This convenience method just checks if Error code is ErrTimedOut.

Also modifies the example in README.md.

See https://github.com/confluentinc/confluent-kafka-go/pull/846#discussion_r1030215346

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This is a maintenance release:
    will be used for subsequent (new) connections to a broker.
  * Channel based producer (Producer `ProduceChannel()`) and channel based
    consumer (Consumer `Events()`) are deprecated.
+ * Added `IsTimeout()` on Error type. This is a convenience method that checks
+   if the error is due to a timeout.
 
 
 ## v1.9.2
--- a/kafka/consumer_test.go
+++ b/kafka/consumer_test.go
@@ -204,6 +204,21 @@ func TestConsumerAPIs(t *testing.T) {
 		t.Errorf("Committed() failed but returned non-nil Offsets: %s\n", offsets)
 	}
 
+	// Test timeouts using ReadMessage.
+	msg, err := c.ReadMessage(time.Millisecond)
+	t.Logf("ReadMessage() returned message %s and error %s\n", msg, err)
+
+	// Check both ErrTimedOut and IsTimeout() to ensure they're consistent.
+	if err == nil || !err.(Error).IsTimeout() {
+		t.Errorf("ReadMessage() should time out, instead got %s\n", err)
+	}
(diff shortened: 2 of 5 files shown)


kafka/error.go
```diff
@@ -127,6 +127,12 @@ func (e Error) IsRetriable() bool {
 	return e.retriable
 }
 
+// IsTimeout returns true if the error is a timeout error.
+// A timeout error indicates that the operation timed out locally.
+func (e Error) IsTimeout() bool {
+	return e.code == ErrTimedOut
```
edenhill, line 133: Also check for ErrTimedOutQueue

edenhill (changes requested):
Add a CHANGELOG entry.
Also recommend updating one of the existing tests to use this too:
find a test that checks Code() == ..TimedOut,
in that test also add IsTimeout().

Why keep the old way? Because we want to make sure that we're not breaking existing functionality.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ This is a maintenance release:
    will be used for subsequent (new) connections to a broker.
  * Channel based producer (Producer `ProduceChannel()`) and channel based
    consumer (Consumer `Events()`) are deprecated.
+ * Added IsTimeout on Error. This is a convenience method that checks if the
```
edenhill, line 12:
```suggestion
 * Added `IsTimeout()` on Error type. This is a convenience method that checks if the
```

kafka/consumer_test.go
```diff
@@ -295,6 +295,9 @@ func TestConsumerAssignment(t *testing.T) {
 		if err.(Error).Code() != ErrTimedOut {
 			t.Errorf("Expected ReadMessage to fail with ErrTimedOut, not %v", err)
 		}
+		if !err.(Error).IsTimeout() {
+			t.Errorf("Expected ReadMessage to fail with a timeout error, not %v", err)
```
edenhill, line 299: Maybe add a case for the success case too. E.g., from the TestConsumerAPIs dry-test (doesn't require a broker), you could call ReadMessage() with a low timeout and verify it is a timeout.

edenhill (changes requested):

CHANGELOG.md
milindl: Done

kafka/consumer_test.go
milindl: I'm a bit unsure if I got this right since the case in `TestConsumerAssignment` is also a success case (ie it's a timeout). 
I added a test in TestConsumerAPIs, hopefully that captures the intent of this comment.

edenhill (approved):
LGTM

PR #904, opened 2022-11-29
Fix gofmt errors and broker check on integration test

milindl:
An integration test which should be checking for testconf.json, was not checking for it correctly, and wasn't being skipped due to that.

There were several issues with formatting, all of them in the generated error file. All those have been fixed along with the generation code.

Additionally, there are a few errors in the _build files, but they won't be fixed till we bump our go version to 1.17 (+build is replaced with go:build, and gofmt 1.18+ expects go:build, but we are going to stick to go 1.14 for now).

All this is needed for setting up a Github Actions workflow effectively.


edenhill (approved):
LGTM

PR #905, opened 2022-11-29
Add Github Actions setup

milindl:
1. build
2. test
3. gofmt
4. make-docs

Changes Makefile to have option to specify go tags for dynamic linking. We need to link dynamically against librdkafka whenever we run an action, because there might be changes in master which we haven't brought over to this repository yet.

Currently the last step (test) for the first job fails, because some recent changes in librdkafka cause the failure of TestTransactionalAPI. So, this PR should ideally be merged only after those have been fixed, else the built-test-doc job will always fail.

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -1,0 +1,75 @@
+name: check
+on: [push, pull_request]
+jobs:
+  # The build and test job builds librdkafka, a dynamically linked version of
+  # the go client, and then runs all the tests which don't require a broker.
+  # We also build the documentation (which also tests out error generation).
+  # We can't use prebuilt librdkafka since it might not contain changes from
+  # master, and so, we have to keep the build, test, and docs steps in one job
+  # to avoid needing to build librdkafka repeatedly.
+  build-test-doc:
+    runs-on: ubuntu-latest
+    steps:
+      # Setup repositories and dependencies for confluent-kafka-go
+      # Two checkouts are needed - the argumentless version which checks out
+      # confluent-kafka-go, and the other, which checks out librdkafka.
+      - uses: actions/checkout@v3
+      - uses: actions/checkout@v3
+        with:
+          repository: edenhill/librdkafka
+          path: './librdkafka'
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+      # We use go1.14, which is the lowest supported version.
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '=1.14'
+      - run: |
+          cd $HOME
+          go get golang.org/x/tools/cmd/godoc
+      - run: pip install beautifulsoup4
+      - run: |
+          sudo apt update
+          sudo apt install -y libcurl4-openssl-dev libssl-dev libsasl2-dev libzstd-dev
+      - run: |
+          cd librdkafka
+          ./configure --install-deps --enable-devel --disable-lz4-ext --prefix=/usr
+          make
+          sudo make install
+      # Build
+      - run: |
+          cd kafka
+          go build -tags dynamic ./...
+      # Docs
+      - run: |
+          export GOTAGS="-tags dynamic"
+          make -f mk/Makefile docs
+      # Tests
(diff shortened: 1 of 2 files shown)


milindl:
Sample run: https://github.com/milindl/confluent-kafka-go/actions/runs/3573381045/jobs/6007376755
Sample run, with the above mention test skipped: https://github.com/milindl/confluent-kafka-go/actions/runs/3573267179

.github/workflows/base.yml
```diff
@@ -0,0 +1,77 @@
...
+  # We also build the documentation (which also tests out error generation).
+  # We can't use prebuilt librdkafka since it might not contain changes from
+  # master, and so, we have to keep the build, test, and docs steps in one job
+  # to avoid needing to build librdkafka repeatedly.
+  build-test-doc:
+    runs-on: ubuntu-latest
+    steps:
+      # Setup repositories and dependencies for confluent-kafka-go
+      - uses: actions/checkout@v3 # checks out self-repo by default
+      - uses: actions/checkout@v3
```
edenhill, line 15: why is this duplicated?

.github/workflows/base.yml
```diff
@@ -0,0 +1,77 @@
...
+          go get golang.org/x/tools/cmd/godoc
+      - run: pip install beautifulsoup4
+      - run: |
+          sudo apt update
+          sudo apt install -y libcurl4-openssl-dev libssl-dev libsasl2-dev libzstd-dev
+      - run: |
+          cd librdkafka
+          ./configure --CFLAGS="-std=c99" --CXXFLAGS="-std=c++98" --install-deps --enable-devel --disable-lz4-ext --prefix=/usr/local
+      - run: |
+          cd librdkafka
```
edenhill, line 36: combine make in the run: .. above

.github/workflows/base.yml
```diff
@@ -0,0 +1,77 @@
...
+          make -j20
```
edenhill, line 37: drop 20, we don't know how many cores the build host has , just -j will make make figure it out.

.github/workflows/base.yml
```diff
@@ -0,0 +1,77 @@
...
+          sudo make install
+      # Build
+      - run: |
+          export LD_PRELOAD=/usr/local/lib/librdkafka.so
```
edenhill, line 41: Shouldn't use LD_PRELOAD, that forces a load, use LD_LIBRARY_PATH=/usr/local/lib so the standard linker is employed.

Or change --prefix to /usr when configuring librdkafka.

.github/workflows/base.yml
```diff
@@ -0,0 +1,77 @@
...
+      - run: |
+          cd $HOME
+          go get golang.org/x/tools/cmd/godoc
+      - run: pip install beautifulsoup4
+      - run: |
+          sudo apt update
+          sudo apt install -y libcurl4-openssl-dev libssl-dev libsasl2-dev libzstd-dev
+      - run: |
+          cd librdkafka
+          ./configure --CFLAGS="-std=c99" --CXXFLAGS="-std=c++98" --install-deps --enable-devel --disable-lz4-ext --prefix=/usr/local
```
edenhill, line 34: No need for the -std stuff.

edenhill (changes requested):

.github/workflows/base.yml
milindl: the first one checks out the repo (confluent-kafka-go), the second one checks out librdkafka. I'll make the comment more clear.

.github/workflows/base.yml
milindl: make, make install moved up

milindl:
Sample run with changes: https://github.com/milindl/confluent-kafka-go/actions/runs/3574366235/jobs/6009541012

milindl:
Thanks! Merging after failing test is fixed.

milindl:
Successful run with latest merges: https://github.com/milindl/confluent-kafka-go/actions/runs/3592174374/jobs/6047577322

Issue #888, opened 2022-11-03
Consumer.Seek or Consumer.Assign having issues


Tang8330:
Description
===========
This issue seems to have cropped up a couple of times, see:
* https://github.com/confluentinc/confluent-kafka-go/issues/121
* https://github.com/confluentinc/confluent-kafka-go/issues/748

In summary, if I try to do `consumer.Seek(...)` or `consumer.Assign(...)` to a `TopicPartition` where the offset doesn't exist anymore (maybe because the offset was older than retention); instead of having `Seek` or `Assign` error out, it returns `nil` and acts as if there was not an error. Meanwhile, I see a log line in my terminal.

```
%4|1667447031.006|OFFSET|rdkafka#consumer-1| [thrd:main]: topic_name [0]: offset reset (at offset 17011, broker 2) to END: fetch failed due to requested offset not available on the broker: Broker: Offset out of range
```

How to reproduce
================
* Try to run Seek or Assign on an offset that doesn't exist in the Broker anymore

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-go and librdkafka version (`LibraryVersion()`): v1.9.2
 - [x] Apache Kafka broker version: (Not sure how to check, I'm using Confluent Cloud)
 - [x] Client configuration: `ConfigMap{...}`
```go
	consumer, err := kafka.NewConsumer(&kafka.ConfigMap{
		"bootstrap.servers": config.GetSettings().Config.Kafka.BootstrapServer,
		"sasl.mechanisms":   "PLAIN",
		"security.protocol": "SASL_SSL",
		"group.id":          config.GetSettings().Config.Kafka.GroupID,
		"sasl.username":     config.GetSettings().Config.Kafka.Username,
		"sasl.password":     config.GetSettings().Config.Kafka.Password,
	})
```

 - [x] Operating system: Macbook M2 Monterey
 - [ ] Provide client logs (with `"debug": ".."` as necessary)
 - [ ] Provide broker log excerpts
 - [x] Critical issue

edenhill:
Seek() and Assign() are asynchronous and simply tell the consumer where to start fetching, the calls will not wait for the FetchRequests to be made to the broker.
If you set `"auto.offset.reset": "error"` you'll receive a consumer error later when the fetch has been attempted.

Tang8330:
[...]

PR #906, opened 2022-11-30
[consumer] Ignore Seek() timeout

milindl:
Seek timeout is ignored and an infinite timeout is used instead. The method will block till fetcher state is updated and the seek is registered.

See #888.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ This is a maintenance release:
    consumer (Consumer `Events()`) are deprecated.
  * Added `IsTimeout()` on Error type. This is a convenience method that checks
    if the error is due to a timeout.
+ * The timeout parameter on `Seek()` is now ignored and an infinite timeout is
+   used, the method will block until the fetcher state is updated (typically
+   within microseconds).
 
 
 ## v1.9.2
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -308,11 +308,12 @@ func (c *Consumer) StoreMessage(m *Message) (storedOffsets []TopicPartition, err
 
 // Seek seeks the given topic partitions using the offset from the TopicPartition.
 //
-// If timeoutMs is not 0 the call will wait this long for the
-// seek to be performed. If the timeout is reached the internal state
-// will be unknown and this function returns ErrTimedOut.
-// If timeoutMs is 0 it will initiate the seek but return
-// immediately without any error reporting (e.g., async).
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ This is a maintenance release:
    consumer (Consumer `Events()`) are deprecated.
  * Added `IsTimeout()` on Error type. This is a convenience method that checks
    if the error is due to a timeout.
+ * The timeout parameter on `Seek()` is ignored and an infinite timeout is used,
+   and the method blocks till fetcher state is updated.
```
edenhill, line 15:
```suggestion
 * The timeout parameter on `Seek()` is now ignored and an infinite timeout is used,
   the method will block until the fetcher state is updated (typically within microseconds).
```

kafka/consumer.go
```diff
@@ -308,24 +308,24 @@ func (c *Consumer) StoreMessage(m *Message) (storedOffsets []TopicPartition, err
 
 // Seek seeks the given topic partitions using the offset from the TopicPartition.
 //
-// If timeoutMs is not 0 the call will wait this long for the
-// seek to be performed. If the timeout is reached the internal state
-// will be unknown and this function returns ErrTimedOut.
-// If timeoutMs is 0 it will initiate the seek but return
-// immediately without any error reporting (e.g., async).
+// The ignoredTimeoutMs parameter is ignored. Instead, this method blocks till
```
edenhill, line 311: until

kafka/consumer.go
```diff
@@ -308,24 +308,24 @@ func (c *Consumer) StoreMessage(m *Message) (storedOffsets []TopicPartition, err
...
+// the fetcher state is updated for the given partition with the new offset.
+// This guarantees that no previously fetched messages for the old offset (or
+// fetch position) will be passed to the application.
```
edenhill, line 314:
```suggestion
// fetch position) will be passed to the application once this call returns.
```

kafka/consumer.go
```diff
@@ -308,24 +308,24 @@ func (c *Consumer) StoreMessage(m *Message) (storedOffsets []TopicPartition, err
...
+// It still takes some time after the method returns to fetch the new offset.
```
edenhill, line 315:
```suggestion
// It will still take some time after the method returns until messages are fetched at the new offset.
```

edenhill (changes requested):

edenhill (approved):
LGTM

Issue #902, opened 2022-11-22
Add SeekPartitions wrapper for rd_kafka_seek_partitions function to support seeking multiple partitions at a time


james-johnston-thumbtack:
Description
===========

The [Seek](https://github.com/confluentinc/confluent-kafka-go/blob/561a26092b1678e2ddcddbff6b2720ee4e85aaf9/kafka/consumer.go#L323) function only supports seeking one partition at a time.  This is due to the underlying limitation of the [rd_kafka_seek](https://github.com/edenhill/librdkafka/blob/51985c7eb287d4a7df93487c90372d239cf0d572/src/rdkafka.h#L3673) function also only seeking one partition at a time.

However, the rd_kafka_seek function was marked as deprecated, with [rd_kafka_seek_partitions](https://github.com/edenhill/librdkafka/blob/51985c7eb287d4a7df93487c90372d239cf0d572/src/rdkafka.h#L3712-L3713) being the new preferred alternative.  This function takes a list of partitions.

Please consider adding a new function named `SeekPartitions` or similar to the Go consumer, which wraps `rd_kafka_seek_partitions`.  This will allow us to avoid having to write loops like this:

```
	assigned, _ := consumer.Assignment()
	for i := range assigned {
		assigned[i].Offset = 0
	}
	for _, a := range assigned {
		_ = consumer.Seek(a, 0)
	}
```
and instead write:
```
	assigned, _ := consumer.Assignment()
	for i := range assigned {
		assigned[i].Offset = 0
	}
	_ = consumer.SeekPartitions(assigned, 0)
```

Checklist
=========
Please provide the following information:

 - [X] confluent-kafka-go and librdkafka version (`LibraryVersion()`): `master` branch: a reference to `rd_kafka_seek_partitions` cannot be found anywhere indicating that there is no existing wrapper function
 - [ ] Apache Kafka broker version: N/A
 - [ ] Client configuration: `ConfigMap{...}` N/A
 - [ ] Operating system: N/A
 - [ ] Provide client logs (with `"debug": ".."` as necessary): N/A
 - [ ] Provide broker log excerpts: N/A
 - [ ] Critical issue

milindl:
Assigning it to myself as we're taking a look at this in our internal tasks loop.

milindl:
Closing with #940, PR merged to master. It will be there in next release.

PR #907, opened 2022-11-30
[consumer] Add SeekPartitions to Consumer

milindl:
Allows seeking multiple partitions.
Deprecates Seek(), which supports seeking single partition.

See #902

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -321,6 +321,7 @@ func (c *Consumer) StoreMessage(m *Message) (storedOffsets []TopicPartition, err
 // a starting offset for each partition.
 //
 // Returns an error on failure or nil otherwise.
+// Deprecated: Seek is deprecated in favour of SeekPartitions().
 func (c *Consumer) Seek(partition TopicPartition, ignoredTimeoutMs int) error {
 	rkt := c.handle.getRkt(*partition.Topic)
 	cErr := C.rd_kafka_seek(rkt,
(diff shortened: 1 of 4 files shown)


kafka/consumer.go
```diff
@@ -332,6 +333,31 @@ func (c *Consumer) Seek(partition TopicPartition, timeoutMs int) error {
 	return nil
 }
 
+// SeekPartitions seeks the given topic partitions to the per-partition offset
+// stored in the .Offset field of each partition.
+//
+// The offset may be either absolute (>= 0) or a logical offset.
```
edenhill, line 339: provide an example for logical offsets: ` (e.g., ....OffsetEnd)`

kafka/consumer.go
```diff
@@ -332,6 +333,31 @@ func (c *Consumer) Seek(partition TopicPartition, timeoutMs int) error {
...
+//
+// SeekPartitions() may only be used for partitions already being consumed
+// (through Assign() or implicitly through a self-rebalanced Subscribe()).
+// To set the starting offset it is preferred to use Assign() and provide
```
edenhill, line 343:
```suggestion
// To set the starting offset it is preferred to use Assign() in a kafka.PartitionsAssigned handler and provide
```

edenhill: (or wahtever the event is called)

edenhill (changes requested):

kafka/integration_test.go
```diff
@@ -1213,6 +1213,48 @@ func TestProducerConsumerHeaders(t *testing.T) {
...
+		t.Errorf(
+			"SeekPartitions should return result for %d partitions, %d returned",
+			len(tps), len(seekedPartitions))
+	}
+	for _, seekedPartition := range seekedPartitions {
+		if seekedPartition.Error != nil {
+			t.Errorf("Seek error for partition %v", seekedPartition)
+		}
+	}
+}
```
edenhill, line 1256: Since this is an integration test, it should also check that it indeed receives messages starting at the seeked offset.
I recommend you produce 10 messages to a single partition, set auto.offset.reset=end on the consumer, assign(), and then call SeekPartitions() with offset 5, then do a ReadMessage and verify that the message's offset is indeed 5.

edenhill (changes requested):

kafka/integration_test.go
milindl: Done

PrasanthV454:
LGTM Just need to add the validation on SeekPartitions.

PrasanthV454 (approved):

milindl:
Closing with #940, since it's on an internal branch, and this is on a fork.

PR #913, opened 2022-12-14
feat: consumer with context

nicolascb:


kafka/consumer.go
```diff
@@ -335,10 +336,11 @@ func (c *Consumer) Seek(partition TopicPartition, ignoredTimeoutMs int) error {
 
 // Poll the consumer for messages or events.
 //
-// Will block for at most timeoutMs milliseconds
+// # Will block for at most timeoutMs milliseconds
```
devdrops, line 339: Do we really need this `#` here?

devdrops (commented):
@nicolascb LGTM! I've left a comment, but everything looks great :wink:

Strainy:
Supporting context on `Poll` is something the team at Zendesk put a fair amount of work into over a year ago:
https://github.com/confluentinc/confluent-kafka-go/pull/626

☝️ This approach would allow the timeout value to be dropped entirely on `Poll` in this PR and for the native context timeout to be relied upon. 

We've been running this change on our confluent-kakfa-go fork across multiple services processing a high volume of data for well over a year now without issue.

It would be fantastic if we could revisit this important change 🙏 

I'm happy to work getting that PR updated if it will get renewed attentions from the maintainers here ❤️

marcosvfranca (approved):

ggmariano22 (approved):

KJTsanaktsidis:
Huge 👍 to what @Strainy said - as written, I don't think this PR actually supports correct context cancellation. If `ReadMessageWithContext(ctx, 0)` is called, and `ctx` is not yet canceled when entering the `select` at line 909, then execution falls into the default branch, `c.Poll(0)` gets called, which will eventually wind up calling `C.rd_kafka_queue_poll(_, 0)`. That'll block in native code waiting for an event to become available and the goroutine gets parked.

If `ctx` gets cancelled whilst the goroutine is parked in `rd_kafka_queue_poll` (or really, any time after line 909 executes), nothing will unblock the call to `ReadMessageWithContext` and make it return, which is what context cancellation should achieve.

If there's an appetite to try and get this functionality merged, I'm definitely still happy to join forces with @Strainy to bring our patches up-to-date and get them in! As you said @Strainy , context support is really important for making confluent-kafka-go be a good "go citizen".

Otherwise, people are of course welcome to use our patches in their own apps too, if they're comfortable with the "zero warranties" disclaimer here - https://github.com/confluentinc/confluent-kafka-go/pull/626#issuecomment-988287180

PR #914, opened 2022-12-20
Reintroduce static linking

Strainy:
Basically vendoring the latest librdkafka and documenting some zendesk-specific guidance on how to do this in the future.


Strainy:
🤦

PR #918, opened 2023-01-06
Update go version and add developer instructions

milindl:
The go version is updated because 1.11 has been long out of support, and parts of the code are already using features only available in 1.13+. 

The oldest version of go officially supported (by the golang team) is 1.18, but that is a bit too aggressive for us, and so we go for 1.14. 

And as for preferring go1.18 for development, it's because go1.19 has a bunch of changes in gofmt etc, and the release has been out only recently. So we will stick to 1.18 till it's out of support, and then make the changes to all files with the new gofmt.

### DIFF
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -1,13 +1,21 @@
 # Information for confluent-kafka-go developers
 
-Whenever librdkafka error codes are updated make sure to run generate
+## Development process
+
+1. Use go1.18 (and related tooling) for development on confluent-kafka-go.
+2. Make sure to run `gofmt` and `go vet` on your code.
+3. While there is no hard-limit, try to keep your line length under 80
+   characters.
+3. [Test](#testing) your changes and create a PR.
+
+
+NOTE: Whenever librdkafka error codes are updated make sure to run generate
 before building:
 
 ```
   $ make -f mk/Makefile generr
   $ go build ./...
 ```
-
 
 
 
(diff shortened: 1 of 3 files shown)


kafka/README.md
```diff
@@ -1,6 +1,14 @@
 # Information for confluent-kafka-go developers
 
-Whenever librdkafka error codes are updated make sure to run generate
+## Development process
+
+1. Use go1.18 (and related tooling) for development on confluent-kafka-go.
+2. Make sure to run `gofmt` and `go vet` on your code.
+3. While there is no hard-limit, try to keep your code under 80 characters.
```
anchitj, line 7: code width?

milindl: done, used the same language as librdkafka

anchitj (approved):

kafka/README.md
```diff
@@ -1,6 +1,15 @@
 # Information for confluent-kafka-go developers
 
-Whenever librdkafka error codes are updated make sure to run generate
+## Development process
+
+1. Use go1.18 (and related tooling) for development on confluent-kafka-go.
```
pranavrth, line 5: Instead of "Use go1.18 (and related tooling)", we can write "Use Go (preferred v1.18+) and related tooling"

milindl: We don't want people to use 1.19, because gofmt in that version is not backwards compatible. We will need to make one gofmt commit before we are able to encourage people to use 1.19.

As for versions before 1.18, they would work, except that they are no longer supported by golang developers.

PR #928, opened 2023-01-17
Rename GroupResult to ConsumerGroupResult,

emasab:
rename DeleteConsumerGroupResult to DeleteConsumerGroupsResult


PrasanthV454 (approved):
Shouldn't we update this on change log?

PR #932, opened 2023-01-19
librdkafka static bundle v2.0.1

emasab:


kafka/librdkafka_vendor/rdkafka.h
```diff
@@ -2056,7 +2056,7 @@ void rd_kafka_conf_set_log_cb(rd_kafka_conf_t *conf,
  *                 rd_kafka_conf_set_opaque().
  *
  * For more information on the format of \p json, see
- * https://github.com/confluentinc/librdkafka/wiki/Statistics
+ * https://github.com/edenhill/librdkafka/wiki/Statistics
```
milindl, line 2059: Ideally, this should remain as https://github.com/confluentinc/librdkafka/wiki/Statistics , right?

milindl (approved):
LGTM after the Semaphore run is done

kafka/librdkafka_vendor/rdkafka.h
emasab: This is taken from the bundle, we should change there first. For the moment let's keep the redirect

PR #940, opened 2023-02-08
 [consumer] Add SeekPartitions to Consumer  

milindl:
Allows seeking multiple partitions.
Deprecates Seek(), which supports seeking single partition.

See https://github.com/confluentinc/confluent-kafka-go/issues/902

Added for replacing #907 since it's on a fork.


PrasanthV454 (approved):
LGTM

PR #942, opened 2023-02-10
 [consumer] Add example for rebalance callback 

milindl:
Creating for #891 on an internal branch for the checks to run. 

---

Adds a consumer example for usage of rebalance cb. Includes enable.auto.commit and enable.auto.offset.store each as optional fields to set/unset, and how to write the callback in each of those cases.
Only for EAGER rebalancing strategy, for example, range or roundrobin.

Fixes https://github.com/confluentinc/confluent-kafka-go/issues/886
Should help https://github.com/confluentinc/confluent-kafka-go/issues/884


examples/consumer_example/consumer_example.go
```diff
@@ -81,11 +85,19 @@ func main() {
...
 				fmt.Printf("%% Message on %s:\n%s\n",
 					e.TopicPartition, string(e.Value))
 				if e.Headers != nil {
 					fmt.Printf("%% Headers: %v\n", e.Headers)
 				}
+
+				// We can store the offsets of the messages manually or let
+				// the library do it automatically based on the setting
+				// enable.auto.offset.store. Once an offset is stored, the
+				// library takes care of periodically committing it to broker.
```
PrasanthV454, line 98: Can we make this as 
```suggestion
				// library takes care of periodically committing it to broker
				// when enable.auto.commit isn't set to false, default is true.
```

PrasanthV454 (approved):
LGTM

examples/consumer_example/consumer_example.go
milindl: Done, but changed language just a little

milindl:
I don't think we should add a CHANGELOG.md entry, because that should be for changes affecting the end-user of the library, like new API, or updates/deprecations in the API, or bug fixes (which might potentially change the behaviour in existing API).

PrasanthV454:
> I don't think we should add a CHANGELOG.md entry, because that should be for changes affecting the end-user of the library, like new API, or updates/deprecations in the API, or bug fixes (which might potentially change the behaviour in existing API).

got it.

milindl:
However, I'd missed updating the README.md inside the examples/ directory, so just added that.

PR #945, opened 2023-02-15
[WIP]implemented incremental alter configs (KIP-339)

PrasanthV454:
PR for [KIP-339](https://cwiki.apache.org/confluence/display/KAFKA/KIP-339%3A+Create+a+new+IncrementalAlterConfigs+API) implementation. With this PR new AlterOperations are introduced -> 

1. Append
2. Subtract
3. Remove

There is no necessity in passing all config values for changing certain configs. 
Support for list type Config Values is also added.


kafka/config.go
```diff
@@ -91,13 +93,29 @@ func value2string(v ConfigValue) (ret string, errstr string) {
...
 		ret = x
+	case types.Slice:
+		ret = ""
+		arr := v.([]ConfigValue)
+		for _, i := range arr {
+			temp, err := value2string(i)
+			if err != "" {
+				ret = ""
+				errstr = fmt.Sprintf("Invalid value type %T", v)
+				break
```
milindl, line 104: nit:why not just return here? 
return nil, fmt.Sprintf("Invalid value type %T", v)

and avoid the whole variable errstr as it was earlier

kafka/config.go
```diff
@@ -91,13 +93,29 @@ func value2string(v ConfigValue) (ret string, errstr string) {
 		ret = fmt.Sprintf("%d", x)
 	case string:
 		ret = x
+	case types.Slice:
```
milindl, line 96: Question: I didn't get where we're using this yet, or why we've added this case.

milindl (commented):
Minor changes only

kafka/adminapi.go
```diff
@@ -434,6 +471,19 @@ func StringMapToConfigEntries(stringMap map[string]string, operation AlterOperat
 	return ceList
 }
 
+// StringMapToIncrementalConfigEntries creates a new map of ConfigEntry objects from the
+// provided string map an operationMap. The AlterConfigOpType is set on each created entry.
```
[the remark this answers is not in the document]
emasab: Preferred changing operationMap to be consistent with `StringMapToConfigEntries` documentation

kafka/integration_test.go
```diff
@@ -1363,6 +1361,44 @@ func (its *IntegrationTestSuite) TestAdminConfig() {
...
+	opsMap["message.timestamp.type"] = AlterConfigOpTypeDelete
+
+	configResources = []ConfigResource{{Type: ResourceTopic, Name: topic,
+		Config: StringMapToIncrementalConfigEntries(newConfig, opsMap)}}
+	alterRes, err = a.IncrementalAlterConfigs(ctx, configResources)
+	if err != nil {
+		t.Fatalf("Incremental Alter Configs request failed: %v", err)
+	}
+
+	expResources[0].Config["retention.ms"] = ConfigEntryResult{Name: "retention.ms", Value: "86000000"}
```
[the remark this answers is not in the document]
emasab: It's not possible here as the expected values are different from the values in newConfig, because of different types of operations.

kafka/config.go
emasab: If we set a slice as config value for a list type configuration

kafka/config.go
emasab: I thought reducing the number of returns would make it more readable and predictable.

milindl (approved):

kafka/config.go
milindl: ack

kafka/integration_test.go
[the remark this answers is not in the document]
milindl: ack

milindl:
LGTM, can merge after librdkafka (CI will not pass until then)

PR #948, opened 2023-02-21
Update docs to ensure librdkafka release notes are consulted when vendoring

Strainy:
Basic docs update to make sure we check the librdkafka release notes for any potential or actual breaking changes to zendesk_kafka_go before we vendor.


Strainy:
whoops

Issue #943, opened 2023-02-13
SchemaRegistryClient handleRequest not honoring original schema registry URL with path


huizhilu:
Description
===========

https://github.com/confluentinc/confluent-kafka-go/blob/ff4c2c5a6f2c6555ed53f208c173954bb2bcc243/schemaregistry/rest_service.go#L275

If the schema registry URL has a path: `https://api-proxy/schema-registry`, the final endpoint url for GetSubjects() becomes `https://api-proxy:8082/subjects`, which doesn't include `/schema-registry` path.

Is this expected? Using `url.Parse()` will replace the path. If `JoinPath()` is used, it'll give the correct result. Shall we use `JoinPath()`?

How to reproduce
================

1. Set the schema registry URL to have a path like `https://api-proxy/schema-registry`
2. Init the schema registry client
3. call GetSubjects() API

Checklist
=========
Please provide the following information:

 - [ ] confluent-kafka-go and librdkafka version (`LibraryVersion()`):
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `ConfigMap{...}`
 - [ ] Operating system:
 - [ ] Provide client logs (with `"debug": ".."` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

huizhilu:
Add the owner of the original PR #776  
cc @rayokota  Can you take a look if this can be fixed? Thanks!

huizhilu:
+ @jordaniversen  @mhowlett @perdue @emasab for more visibility. Thank!

PrasanthV454:
Hi @huizhilu, thanks for reporting this. We will make required changes and commit them in the next release.

JoinPath() do give a correct result but requires go 1.19 or later versions.

milindl:
Closing with #950, it will make its way to the next release since it's merged.

PR #950, opened 2023-02-21
Fix Url Parsing in Schema Registry

PrasanthV454:
#943 - Should fix this(Supports all schema-registry base urls which have existing path).

### DIFF
--- a/schemaregistry/rest_service.go
+++ b/schemaregistry/rest_service.go
@@ -30,6 +30,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"path"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


milindl:
While the new code seems fine to me, it seems we were deliberately going up one level due to the addition of base (`..`). 
(discounting url.Parse behaviour, which seems more like a bug)
.
So, I wanted to know the rationale behind that, and if is it okay to remove the `..`.
Maybe @rayokota  can weigh in here with the original context.

rayokota (approved):
LGTM.  The original code came from https://github.com/confluentinc/confluent-kafka-go/pull/231/files#diff-569ea523a93f2963a2e9262bc3350ed168e8b330f0a3e9794057ee5eb3f00600R282 so I don't know if it was originally a bug or on purpose.

PR #955, opened 2023-02-23
Run Integration Tests on Semaphore per PR basis

mahajanadhitya:


.semaphore/semaphore.yml
```diff
@@ -38,96 +40,4 @@ blocks:
             - name: EXPECT_LINK_INFO
               value: dynamic
           commands_file: semaphore_commands.sh
-  - name: "Go 1.19 linux bundled librdkafka"
-    dependencies: [ ]
-    task:
-      agent:
```
mahajanadhitya, line 44: These will be reverted now they are only for the quick feedback cycle

kafka/integration_test.go
```diff
@@ -2376,20 +2386,23 @@ func (its *IntegrationTestSuite) TestProducerConsumerHeaders() {
 
 func TestIntegration(t *testing.T) {
 	its := new(IntegrationTestSuite)
+	its.skipFlaky = true
```
milindl, line 2389: Please only set true by default if testconf.Semaphore == true. It's okay if these tests fail locally and in fact better since we are aware that something is flaky. This step will effectively remove the tests

kafka/integration_test.go
```diff
@@ -2376,20 +2386,23 @@ func (its *IntegrationTestSuite) TestProducerConsumerHeaders() {
...
 		if err := execErr.Error; err != nil {
-			its.T().Fatal(execErr)
+			fmt.Fprintf(os.Stderr, "up -d command failed with the error message %s\n", err)
+			t.Fatal(execErr)
 		}
 		// It takes some time after the containers come up for them to be ready.
 		time.Sleep(20 * time.Second)
 	}
-
+	fmt.Fprintf(os.Stdout, "We are going to run the suite\n")
```
milindl, line 2406: Remove this, it's not required

kafka/integration_test.go
```diff
@@ -2376,20 +2386,23 @@ func (its *IntegrationTestSuite) TestProducerConsumerHeaders() {
...
 	}
-	if testconf.Docker {
-		its.compose = testcontainers.NewLocalDockerCompose([]string{"testresources/docker-compose.yaml"}, "test-docker")
+	fmt.Fprintf(os.Stdout, "TestconfRead was successful!\n")
+	if testconf.Docker && !testconf.Semaphore {
+		its.compose = testcontainers.NewLocalDockerCompose([]string{"./testresources/docker-compose.yaml"}, "test-docker")
 		execErr := its.compose.WithCommand([]string{"up", "-d"}).Invoke()
 		if err := execErr.Error; err != nil {
-			its.T().Fatal(execErr)
+			fmt.Fprintf(os.Stderr, "up -d command failed with the error message %s\n", err)
```
milindl, line 2400: Use t.Fatalf and combine this Fprintf  with t.Fatal

kafka/integration_test.go
```diff
@@ -2376,20 +2386,23 @@ func (its *IntegrationTestSuite) TestProducerConsumerHeaders() {
...
+	its.skipFlaky = true
 	testconfInit()
 	if !testconfRead() {
-		t.Skipf("testconf not provided or not usable")
+		t.Skipf("testconf not provided or not usable\n")
 		return
 	}
-	if testconf.Docker {
-		its.compose = testcontainers.NewLocalDockerCompose([]string{"testresources/docker-compose.yaml"}, "test-docker")
+	fmt.Fprintf(os.Stdout, "TestconfRead was successful!\n")
```
milindl, line 2395: Remove this line too, since debugging is done, it's not required

kafka/testhelpers.go
```diff
@@ -52,26 +53,31 @@ var defaulttestconfConfig = [1]string{"api.version.request=true"}
 
 const defaulttestconfBrokers = "localhost:9092"
 
+// flag for semaphore job
+var semaphoreJob = flag.Bool("onSemaphore", false, "Tells if the job is running on Semaphore")
```
milindl, line 57: Rename the flag to clients.semaphore

That's the naming convention we've chosen for flags, for example, testify flags are called "testify.<flag>" and so on.

kafka/testhelpers.go
```diff
@@ -83,6 +89,11 @@ func testconfRead() bool {
 	testconf.Topic = defaulttestconfTopic
 	testconf.Brokers = ""
 
+	if testconf.Docker || testconf.Semaphore {
+		testconf.Brokers = defaulttestconfBrokers
+		return true
```
milindl, line 94: We can't return true here, since the user might be using testconf.Docker = true in their local setup. In that case we still need to consider the value of the other fields.

We can perhaps do 
```
if testconf.Docker || testconf.Semaphore {
		testconf.Brokers = defaulttestconfBrokers
		if cf == nil {
                  return true

**Review by milindl — COMMENTED** [review]
Mostly good, needs some small changes before merge

File: kafka/testhelpers.go (hunk 7 above)
Comment (mahajanadhitya) [reply kafka/testhelpers.go:94]: Did not get you! If the docker is true then all the default values are taken and no other value is entertained even if the testconf has some different ones

File: kafka/integration_test.go (hunk 4 above)
Comment (mahajanadhitya) [reply kafka/integration_test.go]: sure

File: kafka/testhelpers.go (hunk 6 above)
Comment (mahajanadhitya) [reply kafka/testhelpers.go]: done !

File: kafka/integration_test.go (hunk 2 above)
Comment (milindl) [reply kafka/integration_test.go:2389]: Why don't we remove this variable completely in that case?
Just use 

```
if  testconf.Semaphore {
                t.Skipf("Skipping TestConsumerCommitted since it is flaky[Does not run when tested with all the other integration tests]")
		return
```

File: kafka/testhelpers.go (hunk 7 above)
Comment (milindl) [reply kafka/testhelpers.go:94]: In that case, please close cf if it was opened. Currently, cf.Close() is never called corresponding to os.Open()

I suggest we do something like this:
``` 
       cf, err := os.Open("./testconf.json")
	if err != nil && !testconf.Docker && !testconf.Semaphore {
		fmt.Fprintf(os.Stderr, "%% testconf.json not found and docker compose not setup - ignoring test\n")
		return false
	}
	if err == nil {
	    defer cf.Close()
	}

milindl: And remove line 103 which is currently closing cf.

milindl (approved):
Approved modulo some small changes, please make these changes and merge the PR

PR #964, opened 2023-03-21
[admin] KIP-430: Add authorized operations to describe API

jainruchir:


examples/admin_describe_topics/admin_describe_topics.go
```diff
@@ -0,0 +1,95 @@
...
+	"time"
+
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
+)
+
+func main() {
+	if len(os.Args) < 4 {
+		fmt.Fprintf(
+			os.Stderr,
+			"Usage: %s <bootstrap-servers> <include_authorized_operations"+
```
pranavrth, line 34:
```suggestion
			"Usage: %s <bootstrap-servers> <include_authorized_operations>"+
```

kafka/integration_test.go
```diff
@@ -1093,6 +1095,417 @@ func (its *IntegrationTestSuite) TestAdminClient_ListAndDescribeConsumerGroups()
...
+			Host:                "*",
+			Operation:           ACLOperationRead,
+			PermissionType:      ACLPermissionTypeAllow,
+		},
+	}
+
+	ctx, cancel = context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+
+	resultCreateACLs, err := ac.CreateACLs(ctx, newACLs,
```
pranavrth, line 1474: Delete these acls as well

kafka/integration_test.go
```diff
@@ -1093,6 +1095,417 @@ func (its *IntegrationTestSuite) TestAdminClient_ListAndDescribeConsumerGroups()
...
+			ResourcePatternType: ResourcePatternTypeMatch,
+			Principal:           "User:*",
+			Host:                "*",
+			Operation:           ACLOperationAlter,
+			PermissionType:      ACLPermissionTypeAllow,
+		},
+	}
+	ctx, cancel = context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+	_, err = ac.DeleteACLs(ctx, aclBindingFilters,
```
pranavrth, line 1335: cant we do this using defer while creating the ACL?

pranavrth (changes requested):
First Pass comments

kafka/adminapi.go
```diff
@@ -990,6 +1070,55 @@ func (a *AdminClient) cToTopicResults(cTopicRes **C.rd_kafka_topic_result_t, cCn
...
+	}
+
+	return authorizedOperations
+}
+
+// cToNode converts a C Node_t* to a Go Node.
+func (a *AdminClient) cToNode(cNode *C.rd_kafka_Node_t) Node {
+	node := Node{ID: -1}
+	if cNode == nil {
+		return node
```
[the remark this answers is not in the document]
milindl: Because we're returning a struct, not the pointer type `*Node`
As to why I don't return a pointer, rather than a struct, it's because I'm using this method in cToNodes which returns (and should return) []Node. 
I can make a separate one returning a pointer cToNodePtr or something

kafka/adminoptions.go
```diff
@@ -302,6 +306,51 @@ func (ao AdminOptionMatchConsumerGroupStates) apply(cOptions *C.rd_kafka_AdminOp
 	return nil
 }
 
+// AdminOptionIncludeAuthorizedOperations decides if the broker should return
+// authorized operations.
+//
+// Default: false
+//
+// Valid for DescribeConsumerGroups, DescribeTopics, DescribeCluster.
+type AdminOptionIncludeAuthorizedOperations struct {
```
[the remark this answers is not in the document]
milindl: In case the librdkafka default ever changes, we don't need to change anything here if we use two variables, see similar admin options in the file, too

kafka/testhelpers.go
```diff
@@ -33,15 +34,19 @@ import (
...
-	Config       []string
-	conf         ConfigMap
+	Docker        bool
+	Semaphore     bool
+	Brokers       string
+	BrokersSasl   string
+	SaslUsername  string
+	SaslPassword  string
+	SaslMechanism string
+	Topic         string
```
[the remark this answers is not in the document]
milindl: This will break existing test configuration files that everyone is using
It's not a part of any API guarantees or anything, though, I suppose it's mostly internal... so should I make this change regardless of that?

kafka/adminapi.go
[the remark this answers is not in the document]
milindl: I've changed it to create a new method, and removed the nil check from this one

kafka/adminapi.go
```diff
@@ -1054,11 +1185,111 @@ func (a *AdminClient) cToConsumerGroupDescriptions(
...
+	cPartitionId := C.rd_kafka_TopicPartitionInfo_partition(partitionInfo)
+	info := TopicPartitionInfo{
+		Partition: int(cPartitionId),
+	}
+
+	cLeader := C.rd_kafka_TopicPartitionInfo_leader(partitionInfo)
+	if cLeader != nil {
+		leader := a.cToNode(cLeader)
+		info.Leader = &leader
+	}
```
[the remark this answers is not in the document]
milindl: Handled with cNodeToPtr

kafka/integration_test.go
milindl: Good idea, did the same for describe groups also

kafka/testhelpers.go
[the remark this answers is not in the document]
pranavrth: If it is internal only, then we should make the change.

kafka/adminoptions.go
```diff
@@ -394,9 +443,25 @@ type DescribeConsumerGroupsAdminOption interface {
 	apply(cOptions *C.rd_kafka_AdminOptions_t) error
 }
 
+// DescribeTopicsAdminOption - see setter.
+//
+// See SetAdminRequestTimeout, SetAdminOptionIncludeAuthorizedOperations.
+type DescribeTopicsAdminOption interface {
+	supportsDescribeTopics()
```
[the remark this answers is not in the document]
pranavrth: I would still like to understand this just for my understanding. Maybe over call.

kafka/adminapi.go
```diff
@@ -1012,11 +1158,7 @@ func (a *AdminClient) cToConsumerGroupDescriptions(
 			C.rd_kafka_ConsumerGroupDescription_state(cGroup))
 
 		cNode := C.rd_kafka_ConsumerGroupDescription_coordinator(cGroup)
-		coordinator := Node{
-			ID:   int(C.rd_kafka_Node_id(cNode)),
-			Host: C.GoString(C.rd_kafka_Node_host(cNode)),
-			Port: int(C.rd_kafka_Node_port(cNode)),
-		}
+		coordinator := a.cToNode(cNode)
```
pranavrth, line 1161: We are adding rack information for coordinator in ConsumerGroupDescription. Shall we add this as well to the Changelog?

kafka/adminapi.go
```diff
@@ -271,6 +299,66 @@ type DescribeConsumerGroupsResult struct {
...
+// Topic Partition information
+type TopicPartitionInfo struct {
+	// Partition id.
+	Partition int
+	// Leader broker.
+	Leader *Node
+	// Replicas of partition.
+	Replicas []Node
+	// In-Sync-Replicas of partition.
+	Isr []Node
```
pranavrth, line 325: Shouldn't all be pointers or all be concrete type? I know that leader can be null and hence pointer but just a thought for discussion.

kafka/integration_test.go
```diff
@@ -1093,6 +1095,465 @@ func (its *IntegrationTestSuite) TestAdminClient_ListAndDescribeConsumerGroups()
...
+				Operation:           ACLOperationRead,
+				PermissionType:      ACLPermissionTypeAllow,
+			},
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+		defer cancel()
+		_, err = ac.DeleteACLs(ctx, aclBindingFilters,
+			SetAdminRequestTimeout(time.Second*30))
+		assert.Nil(err, "DeleteACLs should not throw an error")
+	}()
```
pranavrth, line 1520: Better move this after createAcl call

kafka/integration_test.go
```diff
@@ -1093,6 +1095,465 @@ func (its *IntegrationTestSuite) TestAdminClient_ListAndDescribeConsumerGroups()
...
```
pranavrth, line 1217: Same

pranavrth (changes requested):
Looks good. Just few nits.

kafka/adminapi.go
milindl: I like it specifically different because it indicates that Leader can be nullable, but others cannot be nullable. The list can be empty for replicas or isr, but nil cannot be contained in the list.

kafka/adminoptions.go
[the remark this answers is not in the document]
milindl: Sure, I might also add a thing in the file itself, this is a common point of confusion (for me also, when I first wrote an admin option)

kafka/adminapi.go
```diff
@@ -271,6 +299,66 @@ type DescribeConsumerGroupsResult struct {
...
+type TopicDescription struct {
+	// Topic name.
+	Name string
+	// Error, if any, of result. Check with `Error.Code() != ErrNoError`.
+	Error Error
+	// Is the topic is internal to Kafka?
+	IsInternal bool
+	// Partitions' information list.
+	Partitions []TopicPartitionInfo
+	// Operations allowed for topic.
```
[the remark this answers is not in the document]
milindl: Done in other places also

kafka/adminapi.go
milindl: Added a comment in CHANGELOG, it's a bit more generic

milindl:
Thanks for the review @pranavrth ! I've addressed all the comment, either through code or replied on them

pranavrth (approved):
LGTM! Great work!

PR #966, opened 2023-03-23
fix nil pointer error on schemaregistry/serde/protobuf/protobuf.go

baganokodo2022:
an easy fix for potential nil pointer error.
we have hit nil pointer when desc.LoadMessageDescriptorForMessage() returns error.


milindl:
Thanks a lot for your contribution @baganokodo2022! 
We need to run the CI on it, which can only run on an internal branch. I've rebased this commit and created a PR for this here: #997. I will merge it as soon as the CI runs completely.

milindl:
Thanks @baganokodo2022 , your change is merged into master.

PR #968, opened 2023-03-29
KIP-320: Allow fetchers to detect

emasab:
and handle log truncation. Go changes

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -6,7 +6,7 @@ agent:
 global_job_config:
   env_vars:
     - name: LIBRDKAFKA_VERSION
-      value: v2.0.2
+      value: v2.1.0-RC3
   prologue:
     commands:
       - checkout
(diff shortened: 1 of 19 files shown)


kafka/kafka.go
```diff
@@ -276,11 +276,12 @@ const PartitionAny = int32(C.RD_KAFKA_PARTITION_UA)
...
-	Partition int32
-	Offset    Offset
-	Metadata  *string
-	Error     error
+	Topic       *string
+	Partition   int32
+	Offset      Offset
+	Metadata    *string
+	Error       error
+	LeaderEpoch *int32
```
milindl, line 284: Can this just be int32, we can use a negative value (-1) similar to C case (as illustrated in`if cLeaderEpoch >= 0`).
We can change if it makes sense.

kafka/message.go
```diff
@@ -77,6 +77,7 @@ type Message struct {
 	TimestampType  TimestampType
 	Opaque         interface{}
 	Headers        []Header
+	LeaderEpoch    *int32 // LeaderEpoch or nil if not available
```
milindl, line 80: Ditto, the thing about this not being a pointer here, too

milindl (approved):
Some comments about type, but otherwise okay.

PR #970, opened 2023-03-29
bugfix: update protobuf.Deserializer to use base class implementation of ConfigureDeserializer

ideasculptor:
Copying the functionality of a base implementation into the extending struct's implementation is a recipe for later bugs when changes are made in the base implementation that don't get copied into the extending implementation.  It is far better to call the base implementation and then just perform actions that are extension-specific in the extending implementation.


rayokota (approved):
Thanks @ideasculptor , LGTM

PR #984, opened 2023-04-18
Fix schema registry validation - FF-10738

mauricioPisabarro:
Currently it seems as if pre serialization, the serializer does not check against metadata registered on the schema. It looks like it just checks whether it is a well formed json.

I made a few changes to make it as I expected this to work (not sure what the intended behavior is), by comparing directly against metadata that's registered on the schema registry.


mauricioPisabarro:
BTW, had to add a way of getting the subject from the topic. I used the same name strategy as the one used in the `BaseSerializer`'s `GetID`, otherwise it doesn't find the registered schema.

If I don't do thus, but in turn I register `topic1` instead of `topic1-value` in the initialization of the test, then serialization fails, because the aforementioned `GetID` function from the base serializer does not find the id.

milindl:
We're looking into this internally, too, to keep the behavior/interfaces across all the clients the same. We do intend on having the validation work properly, it's just that we might want to do it in a slightly different way (possibly by passing the schema during the initialization of the serializer). I'll provide more comments on this PR as soon as we arrive at some agreement across clients.
Meanwhile, thanks for the PR @mauricioPisabarro !

Nilos:
@milindl this is blocking us from deploying proper schema validation at the moment. Would be really good if we can get this merged and it is fixing the currently buggy behaviour where the schema is not validated at all.

milindl:
Hi - we've been discussing this change internally. The reason why we can't merge this as-is, is because, with different settings of the schema registry's `CacheCapacity`, and if we're serializing for a bunch of topics, we might end up in a situation  where the client is making a network call for many of the serializations, which isn't desired.

One possible solution is to modify the Serializer to accept/fetch a JSON schema at the initialization time, and validate against that for each serialization. However, that needs an API change, and because any API we expose has to be supported for a long time, we're discussing it so we have some parity across clients.

PR #985, opened 2023-04-20
Checking if partitions list is empty before allocating a new list

alexandredantas:
This check makes it possible to list all partitions assigned to a consumer group. In librdkafka, the method `rd_kafka_ListConsumerGroupOffsets` expects the partition to be NULL or having some content. However, in confluent-kafka-go a list is always created


kafka/adminapi.go
```diff
@@ -2197,8 +2197,12 @@ func (a *AdminClient) ListConsumerGroupOffsets(
 	for i, groupPartitions := range groupsPartitions {
 		// We need to destroy this list because rd_kafka_ListConsumerGroupOffsets_new
 		// creates a copy of it.
-		cPartitions := newCPartsFromTopicPartitions(groupPartitions.Partitions)
-		defer C.rd_kafka_topic_partition_list_destroy(cPartitions)
+		var cPartitions *C.rd_kafka_topic_partition_list_t = nil
+
+		if len(groupPartitions.Partitions) > 0 {
```
milindl, line 2202: Change this check to 
`groupPartitions.Partitions != nil`.
This is because the underlying library differentiates between a 0 sized list and a nil topic partition list. The former is an error, but the latter stands for all the partitions.

milindl (changes requested):
Thanks a lot for submitting this PR!

Please also add unit tests to adminapi_test.go, one test case for `nil` topic partition list, which should fail in the unit test with "context deadline exceeded", and one test with 0-sized topic partition list, which should fail with "NULL or non-empty topic partition list must be passed".

Please let me know if you require any help with the changes.

kafka/adminapi.go
alexandredantas: @milindl thanks for your review. I'll change this and add some unit tests for this case

frankgrimes97:
@milindl I believe the requested changes have been addressed... are any further changes required before this can be reviewed, approved and merged? Thanks!

frankgrimes97:
@milindl Any update on this?

milindl:
The internal PR is merged. Thanks for this! It'll be released with 2.2.0

PR #1012, opened 2023-06-20
Deprecate msg.LeaderEpoch

emasab:
msg.TopicPartition.LeaderEpoch is used when storing offsets, that field now is also used when returning the leader epoch associated to the message, similar to what happens with the offset.


milindl (approved):
lgtm. added PR#

PR #1015, opened 2023-06-23
Implement OAuthBearer mechanism for creating Kafka Producer by communicating with SPIRE agent

chang-you:
The goal of these changes is to create Kafka Producer with OAUTHBEARER for Go clients by communicating with the SPIRE agent.

- Create `handleJWTTokenRefreshEvent` function to handle token refresh events
- Create `retrieveJWTToken` function to fetch JWT from SPIRE agent
- Integrate OAuth Bearer mechanism into Kafka Producer to enable secure communication

### DIFF
--- a/examples/spire_producer.example/spire_producer.example.go
+++ b/examples/spire_producer.example/spire_producer.example.go
@@ -1,0 +1,171 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"github.com/spiffe/go-spiffe/v2/svid/jwtsvid"
+	"github.com/spiffe/go-spiffe/v2/workloadapi"
+	"os"
+	"os/signal"
+	"regexp"
+	"syscall"
+	"time"
+
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
+	_ "github.com/spiffe/go-spiffe/v2/spiffeid"
+	_ "github.com/spiffe/go-spiffe/v2/svid/jwtsvid"
+)
+
+var (
+	// Regex for sasl.oauthbearer.config, which constrains it to be
+	// 1 or more name=value pairs with optional ignored whitespace
+	oauthbearerConfigRegex = regexp.MustCompile("^(\\s*(\\w+)\\s*=\\s*(\\w+))+\\s*$")
+	// Regex used to extract name=value pairs from sasl.oauthbearer.config
+	oauthbearerNameEqualsValueRegex = regexp.MustCompile("(\\w+)\\s*=\\s*(\\w+)")
+)
+
+const (
+	principalClaimNameKey = "principalClaimName"
+	principalKey          = "principal"
+	joseHeaderEncoded     = "eyJhbGciOiJub25lIn0" // {"alg":"none"}
+)
+
+type tokenAuth struct {
+	audience    []string
+	tokenSource *workloadapi.JWTSource
+}
+
+// handleJWTTokenRefreshEvent retrieves JWT from the SPIRE workload API and
+// sets the token on the client for use in any future authentication attempt.
+// It must be invoked whenever kafka.OAuthBearerTokenRefresh appears on the client's event channel,
+// which will occur whenever the client requires a token (i.e. when it first starts and when the
+// previously-received token is 80% of the way to its expiration time).
+func handleJWTTokenRefreshEvent(ctx context.Context, client kafka.Handle, principal, socketPath string, audience []string) {
+	fmt.Fprintf(os.Stderr, "Token refresh\n")
+	oauthBearerToken, retrieveErr := retrieveJWTToken(ctx, principal, socketPath, audience)
+	if retrieveErr != nil {
+		fmt.Fprintf(os.Stderr, "%% Token retrieval error: %v\n", retrieveErr)
+		client.SetOAuthBearerTokenFailure(retrieveErr.Error())
+	} else {
+		setTokenError := client.SetOAuthBearerToken(oauthBearerToken)
+		if setTokenError != nil {
+			fmt.Fprintf(os.Stderr, "%% Error setting token and extensions: %v\n", setTokenError)
+			client.SetOAuthBearerTokenFailure(setTokenError.Error())
+		}
+	}
+}
+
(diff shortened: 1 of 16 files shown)


milindl (commented):
Thanks for this PR @chang-you . I gave it a first-pass review, it's a good example to add.

Apart from the comments on the code, there are a few other things:

1. Is it possible to add a README into the folder (like inside examples/spire_consumer_example/README). It can talk about:
  a. how custom OAuth handlers are to be implemented, 
  b. how to run this example locally. 
  c. make it more clear that the client doesn't provide support for SPIRE/SPIFFE, but rather, that this is supposed to be an example implementation of a custom OAuth handler. 
  I can help with writing this README if you prefer, especially part a. and c. 

2. Is there an open guide on how to set up a SPIRE server locally? As the linked guide is inside Confluent's internal page, and external users of the library can't access it. 

3. Add the binary to the examples/.gitignore and the example link to examples/README. Similar to other examples. And add an entry into the CHANGELOG.md, too.

examples/spire_consumer_example/spire_consumer_example.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
+	"github.com/spiffe/go-spiffe/v2/svid/jwtsvid"
+	"github.com/spiffe/go-spiffe/v2/workloadapi"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+)
+
+// handleJWTTokenRefreshEvent retrieves JWT from the SPIRE workload API and
```
arvindth, line 32:
```suggestion
// handleJWTTokenRefreshEvent retrieves JWT from the SPIFFE workload API and
```
for correct terminology.

examples/spire_producer.example/spire_producer.example.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/spiffe/go-spiffe/v2/workloadapi"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
+)
+
+// handleJWTTokenRefreshEvent retrieves JWT from the SPIRE workload API and
```
arvindth, line 32:
```suggestion
// handleJWTTokenRefreshEvent retrieves JWT from the SPIFFE workload API and
```

arvindth (dismissed):

arvindth:
@chang-you what's the status of this PR?

examples/spire_producer.example/spire_producer.example.go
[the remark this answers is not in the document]
chang-you: Thanks for the comments! That makes sense too. 
@arvindth what do you think? Thank you!

examples/spire_consumer_example/spire_consumer_example.go
```diff
@@ -0,0 +1,179 @@
...
+	} else {
+		setTokenError := client.SetOAuthBearerToken(oauthBearerToken)
+		if setTokenError != nil {
+			fmt.Fprintf(os.Stderr, "%% Error setting token and extensions: %v\n", setTokenError)
+			client.SetOAuthBearerTokenFailure(setTokenError.Error())
+		}
+	}
+}
+
+func retrieveJWTToken(ctx context.Context, principal, socketPath string, audience []string) (kafka.OAuthBearerToken, func() error, error) {
```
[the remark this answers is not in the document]
chang-you: Thanks for the comments. To clarify:
> We don't need to describe the exact logic that the method use 

Here do you suggest that we don't need to include a description of this method in the readme, or is there something specific I should modify within this method?

examples/spire_producer.example/spire_producer.example.go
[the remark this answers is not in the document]
arvindth: I agree, a single producer or consumer example is probably good enough.

examples/spire_consumer_example/spire_consumer_example.go
```diff
@@ -0,0 +1,165 @@
...
+	"syscall"
+	"time"
+)
+
+// handleJWTTokenRefreshEvent retrieves JWT from the SPIFFE workload API and
+// sets the token on the client for use in any future authentication attempt.
+// It must be invoked whenever kafka.OAuthBearerTokenRefresh appears on the client's event channel,
+// which will occur whenever the client requires a token (i.e. when it first starts and when the
+// previously-received token is 80% of the way to its expiration time).
+func handleJWTTokenRefreshEvent(ctx context.Context, client kafka.Handle, principal, socketPath string, audience []string) {
```
arvindth, line 37: @chang-you I don't think we're using the principal here correctly. In the oauth producer case, the principal needs to be passed in because it's the principal for which we get a token, and the token would then contain the principal in the `sub` claim.

For spire though, the principal value should be a spiffe id, and we should be using this spiffe id to look through all the svids returned by the spire agent to see if one of them matches the given principal, and use that one.

examples/spire_consumer_example/spire_consumer_example.go
```diff
@@ -0,0 +1,165 @@
...
+	topic := os.Args[2]
+	principal := os.Args[3]
+	socketPath := os.Args[4]
+	audience := []string{"audience1", "audience2"}
+
+	config := kafka.ConfigMap{
+		"bootstrap.servers":       bootstrapServers,
+		"security.protocol":       "SASL_SSL",
+		"sasl.mechanisms":         "OAUTHBEARER",
+		"sasl.oauthbearer.config": principal,
```
arvindth, line 107: I'm not sure if this parameter is even required. Could we remove it and see if it works as expected?

examples/spire_consumer_example/spire_consumer_example.go
```diff
@@ -0,0 +1,165 @@
...
+		"identityPoolId": "pool-W9j5",
+	}
+	oauthBearerToken := kafka.OAuthBearerToken{
+		TokenValue: jwtSVID.Marshal(),
+		Expiration: jwtSVID.Expiry,
+		Principal:  principal,
+		Extensions: extensions,
+	}
+
+	return oauthBearerToken, jwtSource.Close, nil
```
jshahc, line 88: Doesn't seem necessary to return  `jwtSource.Close`, as `defer jwtSource.Close()` is called above

PR #1020, opened 2023-07-04
removed unused import in serde.go

timmyb32r:
Found unused import "C", which makes some tooling think package 'serde' has CGO (but it don't need CGO).


milindl (approved):
Thank you for this PR. Let me get the CI running on this so I can merge it.

PR #1021, opened 2023-07-07
Call rd_kafka_flush from within Producer.Flush

### LINKED ISSUE #1013: Flush should not wait for queue buffering time
Description
===========

This is basically https://github.com/confluentinc/librdkafka/issues/3489 for `confluent-kakfa-go`.

`Flush()` should ignore `queue.buffering.max.ms` and send queued messages immediately. The Go wrapper manually implements `Producer.Flush()` instead of using `rd_kafka_flush`, so it didn't benefit from the above fix.

How to reproduce
================

```
package main

import (
        "fmt"
        "github.com/confluentinc/confluent-kafka-go/v2/kafka"
        "time"
)

func main() {
        _, version := kafka.LibraryVersion()
        fmt.Printf("librdkafka %s\n", version)

        mockCluster, err := kafka.NewMockCluster(2)
        if err != nil {
                panic(err)
        }

        cfg := kafka.ConfigMap{
                "bootstrap.servers":      mockCluster.BootstrapServers(),
                "queue.buffering.max.ms": 1000,
        }
        p, err := kafka.NewProducer(&cfg)
        if err != nil {
                panic(err)
        }

        topic := "test-topic"
        msg := kafka.Message{
                TopicPartition: kafka.TopicPartition{
                        Topic: &topic,
                },
        }
        err = p.Produce(&msg, nil)
        if err != nil {
                panic(err)
        }

        // Consume all producer events
        go func() {
                for range p.Events() {
                }
        }()

        startTime := time.Now()
        n := p.Flush(2000)
        elapsed := time.Since(startTime)
        fmt.Printf("flush took %.3fs, %d messages remaining\n", elapsed.Seconds(), n)
}
```

This outputs something like
```
flush took 1.102s, 0 messages remaining
```
indicating that it's waited for the `queue.buffering.max.ms` time to pass before queued messages were actually flushed.

Checklist
=========

 - [x] confluent-kafka-go and librdkafka version (`LibraryVersion()`): 2.1.1
 - [x] Apache Kafka broker version: librdkafka mock cluster
 - [x] Client configuration: see code above, use a large value for `queue.buffering.max.ms`
 - [x] Operating system: GNU/Linux
 - [ ] Critical issue

milindl:
If rd_kafka_flush isn't called from Producer.Flush, queue.buffering.max.ms is taken into account while Flushing, which isn't desirable behaviour.

Also adds a tests.

Also cleans up some needless string parsing to find flush timeout.

Fixes #1013.


kafka/integration_test.go
```diff
@@ -2519,6 +2519,57 @@ func (its *IntegrationTestSuite) TestProducerConsumerHeaders() {
 
 }
 
+// TestImmediateFlush tests that producer.Flush ignores
+// "queue.buffering.max.ms". (Issue #1013).
+func (its *IntegrationTestSuite) TestImmediateFlush() {
```
jdemeyer, line 2524: This test doesn't seem to depend on the `IntegrationTestSuite` fixtures since it's using the mock cluster.

So I suggest to make a normal test, not a method of `IntegrationTestSuite`. This will make it run by default, without needing to setup a `testconf.json`.

kafka/producer.go
```diff
@@ -129,6 +128,15 @@ rd_kafka_resp_err_t err;
 */
 import "C"
 
+// minInt64 finds the minimum of two int64s.
+// Required until we start using Go 1.18 with generic functions.
```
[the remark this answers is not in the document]
milindl: We're a long way from Go 1.21, unfortunately... we will probably update to 1.17 for the next release, but we generally keep compatibility with older go versions more than the language itself.

kafka/integration_test.go
milindl: Good thought, I made a different file for mock-cluster based tests.

milindl:
Thanks for the review @jdemeyer

milindl:
Please ignore the last commit while reviewing, it's a merge from another PR (I needed to use the same testing related files)

kafka/integration_mock_test.go
```diff
@@ -0,0 +1,220 @@
...
+		wg.Done()
+	}()
+	go func() {
+		err1 = consumer.Close()
+		wg.Done()
+	}()
+
+	wg.Wait()
+
+	assert.True((err0 == nil) || (err1 == nil), "Exactly one of the errors must be nil")
```
emasab, line 211: Can use a single condition
```suggestion
	assert.True((err0 == nil) != (err1 == nil), "Exactly one of the errors must be nil")
```

emasab (approved):
Approved! Just one comment. Great fixes and tests!

Issue #712, opened 2021-11-22
Are there any plans to support an open telemetry hook to the library? 


naveenchlsn:
We need to write custom wrappers over producer and consumer to support open telemetry instrumentation. 
Libraries like [go-redis ](https://pkg.go.dev/github.com/go-redis/redis/extra/redisotel)support open telemetry hooks.
Are there plans to implement a similar wrapper for confluent?

jliunyu:
Can you please provide more details about what is expected to be reported?

kurennoy16:
Hey there
Any news from the year-ago + perspective? 

There is an example for inspiration from DataDog -> https://github.com/DataDog/dd-trace-go/blob/main/contrib/confluentinc/confluent-kafka-go/kafka/kafka.go

That would be great if a commercially supported library, like this one, provides the ability to integrate with OpenTelemetry out of the box.

limoges:
I'd like to request support for that features as well.

As of today, Semantic Convention 1.27.0 suggests the following metrics be implemented, think of Kafka as a "messaging client":
https://opentelemetry.io/docs/specs/semconv/messaging/kafka/

jurabek:
Please give it a try https://github.com/jurabek/otelkafka, feel free to open an issue if you have any suggestions. Currently, the version can trace only function-based producer and consumer.

jurabek:
Hi guys, please give a try https://pkg.go.dev/github.com/jurabek/otelkafka it also can be found on [otel-registry](https://opentelemetry.io/ecosystem/registry/?s=kafka) and feel free to open an issue, if you have any suggestions.

PR #1030, opened 2023-07-20
feat(schemaregistry): allow custom http client

liveFreeOrCode:
This patch introduces the ability to set a custom base http client that will be used for the rest calls to the Schema Registry. 

Common use cases include using traced http clients from telemetry providers.  (This may be a viable solution for #712)

```golang
// example
package main

import (
	"net/http"

	"github.com/confluentinc/confluent-kafka-go/schemaregistry"
	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
)

func main() {
	conf := schemaregistry.NewConfig()
	conf.BaseClient = httptrace.WrapClient(new(http.Client))
	client, err := schemaregistry.NewClient(conf)
}
```

### DIFF
--- a/schemaregistry/config.go
+++ b/schemaregistry/config.go
@@ -54,10 +54,10 @@ type Config struct {
 	// CacheCapacity positive integer or zero for unbounded capacity
 	CacheCapacity int
 
-	// BaseClient will override the default http.Client.
+	// HTTPClient will override the default http.Client.
 	//
 	// The TLS and Timeout configuration above will still be applied.
-	BaseClient *http.Client
+	HTTPClient *http.Client
 }
 
 // NewConfig returns a new configuration instance with sane defaults.
--- a/schemaregistry/config.go
+++ b/schemaregistry/config.go
@@ -16,7 +16,10 @@
 
 package schemaregistry
 
-import "fmt"
+import (
+	"fmt"
+	"net/http"
+)
 
 // Config is used to pass multiple configuration options to the Schema Registry client.
 type Config struct {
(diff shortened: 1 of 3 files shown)


schemaregistry/config.go
```diff
@@ -50,6 +53,11 @@ type Config struct {
 	RequestTimeoutMs int
 	// CacheCapacity positive integer or zero for unbounded capacity
 	CacheCapacity int
+
+	// BaseClient will override the default http.Client.
+	//
+	// The TLS and Timeout configuration above will still be applied.
+	BaseClient *http.Client
```
sagikazarmark, line 60: Maybe call this `HTTPClient` instead?

schemaregistry/rest_service.go
```diff
@@ -133,13 +133,18 @@ func newRestService(conf *Config) (*restService, error) {
 
 	timeout := conf.RequestTimeoutMs
 
+	client := conf.BaseClient
+	if client == nil {
+		client = new(http.Client)
+	}
+
+	client.Transport = transport
+	client.Timeout = time.Duration(timeout) * time.Millisecond
```
sagikazarmark, line 142: I wonder if the "old" config should be deprecated. Also, it might make sense to check if the client has any existing timeout and use that instead if so.

sagikazarmark:
This is a requirement for proper observability, but context propagation would also be necessary (ie. add a `context.Context` first parameter to each method or introduce new `SomethingContext` variants to keep backwards compatibility)

rayokota:
Subsumed by https://github.com/confluentinc/confluent-kafka-go/pull/1099

PR #1031, opened 2023-07-24
[internal branch] mock schema registry client does not return correct id from cache

milindl:
See #971, moved to internal branch to run CI.


milindl:
Closed with #971

PR #1053, opened 2023-08-25
Fix SchemaRegistry cache issue when value addition

djedjethai:
Fix SchemaRegistry cache issue when value addition, lruElements entry deletion does not occur(when needed), nothing much just a word.

take care.


djedjethai:
Hi there, is there someone ? there is a bug in your implementation, so whatever you merge my PR or not at least fix it(some people use it).

rayokota (approved):
Thanks for the PR @djedjethai , LGTM

Issue #1064, opened 2023-09-25
Confluent Schema Registry Client does not honor HTTPS_PROXY environment variable


finncolman:
Description
===========

The Confluent Schema Registry Client does not honor the HTTPS_PROXY environment variable. This means that if you have a firewall rule only allowing outbound HTTPS traffic via a proxy them the schema registry client will break if you switch to using the confluent schema registry client instead of third party schema registry clients such as https://github.com/riferrei/srclient.

This is happening because the http transport is being set here:

(https://github.com/confluentinc/confluent-kafka-go/blob/master/schemaregistry/rest_service.go#L206)

This overrides the default transport with a custom transport but it does not set the Proxy attribute.

In the DefaultTransport this is set to:

`Proxy: ProxyFromEnvironment`

How to reproduce
================

Set environment variable HTTPS_PROXY. This will not be honored. The request will go direct instead.

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-go and librdkafka version (`LibraryVersion()`): v2.2.0 and ibrdkafka-static-bundle-v2.2.0.tgz
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `ConfigMap{...}`
 - [x] Operating system: linux (glibc and musl)
 - [ ] Provide client logs (with `"debug": ".."` as necessary)
 - [ ] Provide broker log excerpts
 - [x] Critical issue

fspinolo:
I am also running into this issue. It looks like there hasn't been any movement on https://github.com/confluentinc/confluent-kafka-go/pull/1065 for a couple of weeks, can we get it merged?

milindl:
Should be closed with #1065 being merged, it'll be available in the next release

PR #1065, opened 2023-09-25
honor HTTPS_PROXY environment variable

finncolman:
This resolves issue [1064](https://github.com/confluentinc/confluent-kafka-go/issues/1064)


finncolman:
See: [ProxyFromEnvironment](https://pkg.go.dev/net/http#ProxyFromEnvironment)

milindl:
Hi @rayokota , asking for your review here, as you would know this much better than I.

Maybe the proxy can be specified via the configuration (as a function?), rather than only being set to being picked from the environment.

rayokota (approved):
Thanks for the PR @finncolman , LGTM

PR #1068, opened 2023-10-02
Added topic id to describeTopics Response

pranavrth:


kafka/adminapi.go
```diff
@@ -1097,6 +1099,16 @@ func (a *AdminClient) cToAuthorizedOperations(
 	return authorizedOperations
 }
 
+// cToUuid converts a C rd_kafka_uuid_t to a Go Uuid.
+func (a *AdminClient) cToUuid(cUuid *C.rd_kafka_uuid_t) Uuid {
```
milindl, line 1103: if rd_kafka_uuid_t is null in the argument, then what happens here? 
we should add a null check and return the 'zero' UUID, or else make this return a *Uuid and return a nil pointer

(I think for current situation it won't ever be nil but this method will also be used for future methods, possibly, or behavior might change)

kafka/testconf-example.json
```diff
@@ -1,5 +1,5 @@
 {
-    "Brokers": "mybroker or $BROKERS env",
+    "Brokers": "localhost:9092",
```
milindl, line 2: change isn't needed

kafka/kafka.go
```diff
@@ -334,6 +334,28 @@ func (n Node) String() string {
...
+	leastSignificantBits int64
+	// Base64 representation
+	base64str string
+}
+
+func (uuid Uuid) String() string {
+	return fmt.Sprintf("%s", uuid.base64str)
+}
+
+func (uuid Uuid) getMostSignificantBits() int64 {
```
milindl, line 351: make this public

```suggestion
func (uuid Uuid) GetMostSignificantBits() int64 {
```

milindl (commented):
mostly just minor stuff

kafka/adminapi.go
pranavrth: Currently, we won't get nil. I think let's update this method later for the nil case as we are not even making any constructor for this field for now.

kafka/integration_test.go
```diff
@@ -1420,6 +1420,7 @@ func (its *IntegrationTestSuite) TestAdminClient_DescribeTopics() {
 		"Expected correct error for nonexistent topic")
 
 	topicDesc := topicDescs[0]
+	assert.NotNil(topicDesc.TopicId)
```
milindl, line 1423: Since the type of topicDesc.TopicId is Uuid, it's a struct and not a nullable type, so this condition will always be true. 
It might be better to add a check like, assert.NotZero(topicDesc.TopicId.GetMostSignificantBytes())  and similar for least significant bytes.
Or that assert.NotEmpty(topicDesc.TopicId.String())
(I'm just taking a guess here, would this condition be true?)

Or if it depends on the broker version, and we can't always guarantee that this will be present, then we should omit this altogether, since this check cannot be false.

milindl (approved):
Approved with one minor comment - also need to merge the latest changes to the kip-430 branch.

kafka/integration_test.go
pranavrth: I added this check just to make sure that we are not removing TopicId from the code by mistake. I think we should keep it here.

kafka/integration_test.go
```diff
@@ -1420,6 +1420,7 @@ func (its *IntegrationTestSuite) TestAdminClient_DescribeTopics() {
 		"Expected correct error for nonexistent topic")
 
 	topicDesc := topicDescs[0]
+	assert.NotNil(topicDesc.TopicID)
```
emasab, line 1423: assert.NotNil asways returns true for structs according to this issue https://github.com/stretchr/testify/issues/570
Suggest:
```suggestion
	assert.NotZero(topicDesc.TopicID.GetLeastSignificantBits())
	assert.NotZero(topicDesc.TopicID.GetMostSignificantBits())
	assert.NotEmpty(topicDesc.TopicID.String())
```

Issue #1062, opened 2023-09-22
Methods cannot be called during final rebalance callback while closing consumer


milindl:
Description
===========

See https://github.com/confluentinc/confluent-kafka-go/pull/901#issuecomment-1688151851
Due to the closed flag being set for the consumer, methods like Commit...() cannot be called during the final rebalance callback.

Issue #1016, opened 2023-06-26
rd_kafka_rebalance_protocol() can hang when called from rebalance callback after consumer close


jdemeyer:
Description
===========

In rare cases, it can happen that the consumer `Close()` hangs forever. The stack trace points to `rd_kafka_rebalance_protocol`. So this sounds very related to https://github.com/confluentinc/librdkafka/issues/4041

The relevant part of the stacktrace:
```
goroutine 83 [syscall, 9 minutes]:
github.com/confluentinc/confluent-kafka-go/v2/kafka._Cfunc_rd_kafka_rebalance_protocol(0x7f817c001890)
        _cgo_gotypes.go:3749 +0x4d
github.com/confluentinc/confluent-kafka-go/v2/kafka.(*Consumer).GetRebalanceProtocol.func1(0xc007a41698?)
        /home/jdemeyer/go/pkg/mod/github.com/confluentinc/confluent-kafka-go/v2@v2.1.1/kafka/consumer.go:224 +0x46
github.com/confluentinc/confluent-kafka-go/v2/kafka.(*Consumer).GetRebalanceProtocol(0xc001300000?)
        /home/jdemeyer/go/pkg/mod/github.com/confluentinc/confluent-kafka-go/v2@v2.1.1/kafka/consumer.go:224 +0x19
github.com/confluentinc/confluent-kafka-go/v2/kafka.(*Consumer).handleRebalanceEvent(0xc007874fc0, 0x0, 0xc0012e14a0?)
        /home/jdemeyer/go/pkg/mod/github.com/confluentinc/confluent-kafka-go/v2@v2.1.1/kafka/consumer.go:1042 +0xeb
github.com/confluentinc/confluent-kafka-go/v2/kafka.(*handle).eventPoll(0xc007874fc8, 0x0, 0x0, 0xc0012e1778?, 0x0)
        /home/jdemeyer/go/pkg/mod/github.com/confluentinc/confluent-kafka-go/v2@v2.1.1/kafka/event.go:182 +0x609
github.com/confluentinc/confluent-kafka-go/v2/kafka.(*Consumer).Poll(...)
        /home/jdemeyer/go/pkg/mod/github.com/confluentinc/confluent-kafka-go/v2@v2.1.1/kafka/consumer.go:450
github.com/confluentinc/confluent-kafka-go/v2/kafka.(*Consumer).Close(0x95d502?)
[...]

PR #1073, opened 2023-10-06
Allow consumer methods to run while closing

milindl:
See #1062
https://github.com/confluentinc/confluent-kafka-go/pull/901#issuecomment-1688151851

With #901, such flows were no longer working after the final Close(). 
```
rebalanceCb() {
   ...
   ...
   if (partitions_are_being_revoked) { 
     commit(some_offsets) 
   }
```

This was because the consumer would be counted as closed and commit would return an error.

it's based on the dev_go_flush_ignore_queue_buffering_time since I'm using the same file

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -553,6 +553,9 @@ func (c *Consumer) Close() (err error) {
 		c.Poll(100)
 	}
 
+	// After this point, no more consumer methods may be called.
+	atomic.StoreUint32(&c.isClosed, 1)
+
 	// Destroy our queue
 	C.rd_kafka_queue_destroy(c.handle.rkq)
 	c.handle.rkq = nil
@@ -561,7 +564,6 @@ func (c *Consumer) Close() (err error) {
 
 	C.rd_kafka_destroy(c.handle.rk)
 
-	atomic.StoreUint32(&c.isClosed, 1)
 	return nil
 }
 
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -49,7 +49,8 @@ type Consumer struct {
 	appReassigned      bool
 	appRebalanceEnable bool // SerializerConfig setting
 
-	isClosed uint32
+	isClosed  uint32
+	isClosing uint32
 }
 
 // IsClosed returns boolean representing if client is closed or not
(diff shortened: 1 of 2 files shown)


kafka/consumer.go
```diff
@@ -528,8 +529,15 @@ func (c *Consumer) ReadMessage(timeout time.Duration) (*Message, error) {
...
-	if !atomic.CompareAndSwapUint32(&c.isClosed, 0, 1) {
-		return getOperationNotAllowedErrorForClosedClient()
+	// Check if the client is already closed.
+	err = c.verifyClient()
+	if err != nil {
+		return err
+	}
+
+	// Client is in the process of closing.
+	if !atomic.CompareAndSwapUint32(&c.isClosing, 0, 1) {
```
pranavrth, line 539: It is advised to use `UInt32.CompareAndSwap` instead. https://pkg.go.dev/sync/atomic#CompareAndSwapUint32

Though this function is added in Go 1.19. We should figure out a way to make this happen at least in future when min Go is 1.19.

kafka/consumer.go
```diff
@@ -528,8 +529,15 @@ func (c *Consumer) ReadMessage(timeout time.Duration) (*Message, error) {
...
+		return newErrorFromString(ErrState, "Consumer is already closing")
```
pranavrth, line 540: Added new error code changes the contract. Can we use old error?

kafka/consumer.go
milindl: Agreed, there are a few more places where we have things to do pending Go version upgrades. I'll note them internally.

kafka/consumer.go
milindl: The error code is the same (ErrState), but the error message is different. The user is expected to check the error code if doing any error-specific handling.
The string "Operation not allowed on closed client" is not correct for this error, what do you think?

pranavrth: If the errorcode is same I think it should be fine.

kafka/consumer.go
```diff
@@ -545,6 +553,9 @@ func (c *Consumer) Close() (err error) {
 		c.Poll(100)
 	}
 
+	// After this point, no more consumer methods may be called.
+	atomic.StoreUint32(&c.isClosed, 1)
```
pranavrth, line 557: Should this be just before returning? `rd_kafka_destroy` can stuck due to various reasons.

milindl: We're destroying an important queue below, which is necessary for most public API to run (it's the queue on which we're getting all the events from, from librdkafka side).

So most functions called after this point will panic due to segfault

pranavrth: I was just talking about `atomic.StoreUint32(&c.isClosed, 1)`

pranavrth (approved):

milindl:
It's merged but only in the other branch. I did it on top of that because of test file organization related changes.

jdemeyer:
Is this by chance related to https://github.com/confluentinc/confluent-kafka-go/issues/1016 ?

milindl:
I'm guessing it would not be. I saw the stack traces in those functions, we've already made it to the librdkafka functions and are waiting for those to return, the internal state of the Go consumer wouldn't come into the picture

Issue #425, opened 2020-02-03
Delivery report from events channel does not have timestamp and headers


jasimsulaiman:
Description
===========
Delivery report of message from events channel, when using channel producer does not have timestamp or headers. This behavior is same for successfully delivered, as well as on timeout on failure to deliver to kafka. 

Not having headers makes it very hard to identify which message failed.

How to reproduce
================
Send message with timestamp:
```msg := &kafka.Message{
		TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny},
		Value:          out,
		Headers:        mapToHeaders(headers),
		Timestamp:      time.Date(2018, 1, 1, 1, 1, 1, 1, time.UTC),
		TimestampType:  kafka.TimestampCreateTime,
	}

producer.ProduceChannel() <- msg
```

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-go and librdkafka version (`LibraryVersion()`): confluent-kafka-go:v1.3.0, librdkafka-1.3.0-r0
 - [x] Apache Kafka broker version: kafka_2.12-2.4.0
 - [x] Client configuration: `ConfigMap{...}` `kafka.ConfigMap{"bootstrap.servers": serverIp,
		"broker.version.fallback": "2.1.1", "message.timeout.ms": "1000", "api.version.request": true}`
 - [x] Operating system: `alpine:3.11`
 - [x] Provide client logs (with `"debug": ".."` as necessary)
code: 
```go func() {
		for e := range producer.Events() {
			switch ev := e.(type) {
			case *kafka.Message:
				m := ev
				log.Print("mes t: ", m.Timestamp)
			default:
				log.Print("default", e)
			}
		}
	}()
```
log: `2020/02/02 14:58:46 mes t: 0001-01-01 00:00:00 +0000 UTC`

- [x] Critical: yes

silencev:
I also set metadata in headers and would like to know the delivery result from headers.
for the timestamp, I'm using it to calculate the delivery latency.

edenhill:
@silencev Sorry, timestamps and headers are not available in the delivery report. You can pass a user object of your choice with each produced message which you can retrieve in the delivery report, you can use that to measure latency, et.al.

In librdkafka there's a `rd_kafka_message_latency()` which can be used in delivery reports to get the total time from produce() to delivery, it could easily be exposed in the Go client as well.

none-da:
> user object of your choice
[...]

PR #1076, opened 2023-10-09
Adding producer latency to Message Delivery Report

AdrielVelazquez:
Originally in response to this message, and better metrics internally. 

https://github.com/confluentinc/confluent-kafka-go/issues/425#issue-559017454

### Background:

Our internal applications measure producer latency, by consuming directly from the topic and producing a diff of the message timestamp, and the event timestamp. Exposing the Producer Latency directly to the delivery report will help prevent having an unnecessary consumer to report on Producer Latency.

### DIFF
--- a/kafka/message.go
+++ b/kafka/message.go
@@ -70,14 +70,15 @@ func (t TimestampType) String() string {
 
 // Message represents a Kafka message
 type Message struct {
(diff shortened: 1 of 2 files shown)


ivaaaan:
Are there any plans to merge this from Kafka team? We also faced this problem in our systems

PR #1084, opened 2023-10-24
Fix calls with empty lists

emasab:
DescribeTopics and ListOffsets should return an empty output if and empty input is passed as parameter.
Consistent with Java and user friendly for users using filter and map.


milindl (approved):

milindl:
CI will not be green for this PR without the merge of #1085 as they're linked.

PR #1085, opened 2023-10-24
librdkafka v2.3.0-RC4

emasab:


milindl (approved):
Approved, but only merge at the same time with #1084 - CI will not be green for either of these PRs without the other, and we will have to merge one with a red CI.

PR #1095, opened 2023-11-06
Move doc from InitTransactions to BeginTransaction

jdemeyer:
The way I understand it, the text about `transaction.timeout.ms` is relevant for `BeginTransaction` and not for `InitTransactions` (but please correct me if I'm misunderstanding this).


milindl:
Yes, that's right. Thanks @jdemeyer !

milindl (approved):

PR #1097, opened 2023-11-07
ci: Update mac build agents versions

xli1996:
- Update mac agent to 13.5 which is close to developer's local versions.
- new mac agent syntax is same as ubuntu agents


PrasanthV454 (approved):
LGTM

PR #1103, opened 2023-11-09
[DSG-8949] support Schema Registry bearer auth with static token

yangpei1214:
Flink Cell Operator wants to use bearer access token to connect to SR and it already has logic to get the token.

### DIFF
--- a/schemaregistry/config.go
+++ b/schemaregistry/config.go
@@ -88,6 +88,18 @@ func NewConfig(url string) *Config {
 	return c
 }
 
+// NewConfigWithAuthentication returns a new configuration instance using basic authentication.
+// For Confluent Cloud, use the API key for the username and the API secret for the password.
+// This method is deprecated.
+func NewConfigWithAuthentication(url string, username string, password string) *Config {
+	c := NewConfig(url)
+
+	c.BasicAuthUserInfo = fmt.Sprintf("%s:%s", username, password)
+	c.BasicAuthCredentialsSource = "USER_INFO"
+
+	return c
+}
+
 // NewConfigWithBasicAuthentication returns a new configuration instance using basic authentication.
 // For Confluent Cloud, use the API key for the username and the API secret for the password.
 func NewConfigWithBasicAuthentication(url string, username string, password string) *Config {
--- a/schemaregistry/config.go
+++ b/schemaregistry/config.go
@@ -44,8 +44,8 @@ type Config struct {
 	BearerAuthCredentialsSource string
 	// BearerAuthLogicalCluster specifies the target SR logical cluster id. It is required for Confluent Cloud Schema Registry
 	BearerAuthLogicalCluster string
-	// BearerAuthIdentityPoolId specifies the identity pool ID. It is required for Confluent Cloud Schema Registry
-	BearerAuthIdentityPoolId string
+	// BearerAuthIdentityPoolID specifies the identity pool ID. It is required for Confluent Cloud Schema Registry
+	BearerAuthIdentityPoolID string
 
 	// SslCertificateLocation specifies the location of SSL certificates.
 	SslCertificateLocation string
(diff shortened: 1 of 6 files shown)


schemaregistry/rest_service.go
```diff
@@ -251,28 +254,71 @@ func configureUSERINFOAuth(conf *Config, header http.Header) error {
...
+			" specified with STATIC_TOKEN")
+	}
+	header.Add("Authorization", fmt.Sprintf("Bearer %s", bearerToken))
+	setBearerAuthExtraHeaders(conf, header)
+	return nil
+}
+
+func setBearerAuthExtraHeaders(conf *Config, header http.Header) {
+	targetIdentityPoolID := conf.BearerAuthIdentityPoolID
+	if len(targetIdentityPoolID) > 0 {
```
akhileshm1, line 270: should we throw an error if len == 0 since `BearerAuthIdentityPoolID` is a required config?

schemaregistry/rest_service.go
```diff
@@ -251,28 +254,71 @@ func configureUSERINFOAuth(conf *Config, header http.Header) error {
...
+		header.Add(targetIdentityPoolIDKey, targetIdentityPoolID)
+	}
+
+	targetSr := conf.BearerAuthLogicalCluster
+	if len(targetSr) > 0 {
```
akhileshm1, line 275: should we throw an error if len == 0 since `BearerAuthLogicalCluster` is a required config?

akhileshm1 (dismissed):
thanks @yangpei1214 . LGTM. left a couple of minor comments.

schemaregistry/rest_service_test.go
```diff
@@ -83,3 +84,75 @@ func TestConfigureTLS(t *testing.T) {
...
+	_, err = newAuthHeader(url, config)
+	if err == nil {
+		t.Errorf("Should not work if bearer auth token is empty")
+	}
+
+	config.BearerAuthToken = "token"
+	headers, err := newAuthHeader(url, config)
+	if err != nil {
+		t.Errorf("Should work with bearer auth token, got %s", err)
+	} else if val, exists := headers["Authorization"]; !exists || len(val) == 0 ||
```
akhileshm1, line 115: can we also check for the presence of these headers?
```
targetSRClusterKey      = "target-sr-cluster"
targetIdentityPoolIDKey = "Confluent-Identity-Pool-Id"
```

schemaregistry/config.go
```diff
@@ -82,13 +88,28 @@ func NewConfig(url string) *Config {
 	return c
 }
 
-// NewConfigWithAuthentication returns a new configuration instance using basic authentication.
+// NewConfigWithBasicAuthentication returns a new configuration instance using basic authentication.
 // For Confluent Cloud, use the API key for the username and the API secret for the password.
-func NewConfigWithAuthentication(url string, username string, password string) *Config {
+func NewConfigWithBasicAuthentication(url string, username string, password string) *Config {
```
rayokota, line 93: Can we leave the old method `NewConfigWithAuthentication` and deprecate it?  It can just call this new method

schemaregistry/rest_service.go
yangpei1214: TargetSr is only required for cloud SR. if it is used by CP, I guess we don't need it.

schemaregistry/rest_service.go
yangpei1214: targetIdentityPoolID is only required by cloud SR. Let me add a comment.

schemaregistry/rest_service.go
akhileshm1: ah, I missed that. thanks!

schemaregistry/rest_service_test.go
yangpei1214: Those are only required for Cloud SR

yangpei1214: added it. Thanks @akhileshm1

akhileshm1 (approved):
thanks @yangpei1214

rayokota (approved):
Thanks @yangpei1214 , LGTM

PR #1106, opened 2023-11-15
Add CacheLatestTtlSecs to allow expiration of latest schemas

rayokota:
Add CacheLatestTtlSecs to allow expiration of latest schemas

### DIFF
--- a/schemaregistry/cache/lrucache.go
+++ b/schemaregistry/cache/lrucache.go
@@ -131,6 +131,19 @@ func (c *LRUCache) Delete(key interface{}) {
 	}
 }
 
+// Clear clears the cache
+func (c *LRUCache) Clear() {
+	c.cacheLock.Lock()
+	for key, value := range c.lruElements {
+		delete(c.lruElements, key)
+		c.lruKeys.Remove(value)
+	}
+	for key := range c.entries {
+		delete(c.entries, key)
+	}
+	c.cacheLock.Unlock()
+}
+
 // ToMap returns the current cache entries copied into a map
 func (c *LRUCache) ToMap() map[interface{}]interface{} {
 	ret := make(map[interface{}]interface{})
--- a/schemaregistry/schemaregistry_client.go
+++ b/schemaregistry/schemaregistry_client.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"runtime"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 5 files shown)


schemaregistry/cache/lrucache.go
```diff
@@ -131,6 +131,19 @@ func (c *LRUCache) Delete(key interface{}) {
 	}
 }
 
+// Clear clears the cache
+func (c *LRUCache) Clear() {
+	c.cacheLock.Lock()
+	for key, value := range c.lruElements {
```
atxalex8, line 137: Is there a more efficient way to clear than iterating through all the elements?

schemaregistry/schemaregistry_client.go
```diff
@@ -260,18 +264,34 @@ func NewClient(conf *Config) (Client, error) {
...
+			return nil, err
+		}
 	} else {
 		schemaToIDCache = cache.NewMapCache()
 		idToSchemaCache = cache.NewMapCache()
 		schemaToVersionCache = cache.NewMapCache()
 		versionToSchemaCache = cache.NewMapCache()
+		latestToSchemaCache = cache.NewMapCache()
+	}
+	if conf.CacheLatestTTLSecs > 0 {
```
atxalex8, line 278: I'm not familiar with the use case here, but we can have a LruCache with TTL on individual entries, why do we want to clear the entire cache vs having TTLs on each cache entry

schemaregistry/cache/lrucache.go
rayokota: I believe this is idomatic go before go 1.21, where they added a `clear()` method.  In any case, these maps are not expected to get that large as they store one entry per subject

schemaregistry/schemaregistry_client.go
rayokota: The TTL is just to ensure that the latest entries are refreshed every so often.  It's not necessary to have a TTL per entry.

schemaregistry/schemaregistry_client.go
```diff
@@ -260,18 +264,34 @@ func NewClient(conf *Config) (Client, error) {
...
 	}
 	handle := &client{
 		restService:          restService,
 		schemaToIDCache:      schemaToIDCache,
 		idToSchemaCache:      idToSchemaCache,
 		schemaToVersionCache: schemaToVersionCache,
 		versionToSchemaCache: versionToSchemaCache,
+		latestToSchemaCache:  latestToSchemaCache,
 	}
 	return handle, nil
```
emasab, line 296: Coulse use `runtime.SetFinalizer` with `handle` that stops the ticker and a done channel to exit the for loop in the spawned goroutine

rayokota: Yes, let me make that change

rayokota: @emasab , I added the finalizer

emasab (approved):
LGTM. Thanks @rayokota !

PR #1109, opened 2023-12-01
godoc: Change const groups so enum values are grouped with type

evanj:
Use the `const Name Type = expression` syntax for all typed enum constants. This causes the godoc tool to group the consts with the type definition, which makes it easier to find in the package API documentation.

The godoc tool does not group consts if the type comes from the right hand side, as `const Name = Type(expression)`. This is the syntax the kafka package uses. This means most enum const values are not easily found. For example, type OffsetSpec does not show the defined constant values:

https://pkg.go.dev/github.com/confluentinc/confluent-kafka-go/v2@v2.3.0/kafka#OffsetSpec

Instead, the constants are in the package Constants section at the top:

https://pkg.go.dev/github.com/confluentinc/confluent-kafka-go/v2@v2.3.0/kafka#pkg-constants

The ErrorCode type uses the correct syntax:

https://pkg.go.dev/github.com/confluentinc/confluent-kafka-go/v2@v2.3.0/kafka#ErrorCode

This is much easier to understand. This change removes a redundant type from the generated ErrorCode syntax since it is unneeded. This makes the syntax consistent with the other const groups.

For more details, see Go's documentation about const grouping: https://tip.golang.org/doc/comment#const

### DIFF
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -1004,7 +1004,7 @@ const (
 	// EarliestOffsetSpec is used to describe the earliest offset for the TopicPartition.
 	EarliestOffsetSpec OffsetSpec = C.RD_KAFKA_OFFSET_SPEC_EARLIEST
 	// LatestOffsetSpec is used to describe the latest offset for the TopicPartition.
-	LatestOffsetSpec = C.RD_KAFKA_OFFSET_SPEC_LATEST
+	LatestOffsetSpec OffsetSpec = C.RD_KAFKA_OFFSET_SPEC_LATEST
 )
 
 // NewOffsetSpecForTimestamp creates an OffsetSpec corresponding to the timestamp.
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -217,11 +217,11 @@ const (
 	// ConsumerGroupStateCompletingRebalance - completing rebalance
 	ConsumerGroupStateCompletingRebalance ConsumerGroupState = C.RD_KAFKA_CONSUMER_GROUP_STATE_COMPLETING_REBALANCE
 	// ConsumerGroupStateStable - stable
-	ConsumerGroupStateStable = ConsumerGroupState(C.RD_KAFKA_CONSUMER_GROUP_STATE_STABLE)
+	ConsumerGroupStateStable ConsumerGroupState = C.RD_KAFKA_CONSUMER_GROUP_STATE_STABLE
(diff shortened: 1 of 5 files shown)


kafka/adminapi.go
```diff
@@ -211,11 +211,11 @@ type ConsumerGroupState int
...
-	ConsumerGroupStateUnknown = ConsumerGroupState(C.RD_KAFKA_CONSUMER_GROUP_STATE_UNKNOWN)
+	ConsumerGroupStateUnknown ConsumerGroupState = C.RD_KAFKA_CONSUMER_GROUP_STATE_UNKNOWN
 	// ConsumerGroupStatePreparingRebalance - preparing rebalance
-	ConsumerGroupStatePreparingRebalance = ConsumerGroupState(C.RD_KAFKA_CONSUMER_GROUP_STATE_PREPARING_REBALANCE)
+	ConsumerGroupStatePreparingRebalance ConsumerGroupState = C.RD_KAFKA_CONSUMER_GROUP_STATE_PREPARING_REBALANCE
 	// ConsumerGroupStateCompletingRebalance - completing rebalance
-	ConsumerGroupStateCompletingRebalance = ConsumerGroupState(C.RD_KAFKA_CONSUMER_GROUP_STATE_COMPLETING_REBALANCE)
+	ConsumerGroupStateCompletingRebalance ConsumerGroupState = C.RD_KAFKA_CONSUMER_GROUP_STATE_COMPLETING_REBALANCE
 	// ConsumerGroupStateStable - stable
 	ConsumerGroupStateStable = ConsumerGroupState(C.RD_KAFKA_CONSUMER_GROUP_STATE_STABLE)
```
milindl, line 220: Please make the change for the remaining 3 ConsumerGroupStates, too

milindl (changes requested):
Thanks for this PR @evanj !

Just one thing missing, the rest looks great, I'll approve and run the CI once that's done.

kafka/adminapi.go
evanj: Oops thank you clearly my search/replace/edit missed these thank you!

evanj:
Thanks for the review! I've made that fix, and in re-reviewing my change, I noticed one case where I accidentally dropped the type. Since this is a "typeless integer constant" everything continued to compile, but we do want these consts to be typed correctly.

milindl (approved):
Thanks for this PR!

PR #1110, opened 2023-12-01
Run gofmt 1.19; Fix bundle-import.sh to generate correct output

evanj:
Go 1.19 introduced gofmt formatting changes. This change runs gofmt from Go 1.19, since commit 09f2d70be7d3 on 2023-07-14 updated to use that version. It also fixes bundle-import.sh to generate the correct output. The changes are:

* Add the new "//go:build" format lines
* Put "// +build" tags on one line
* Ensure the correct number of newlines

With these changes, future runs of bundle-import.sh should generate output that is not changed by running gofmt, with one exception: The libs line for Windows has a \r\n line ending, since it comes from the pkg-config .pc file which has \r\n line endings. I have left this unmodified.

`kafka/README.md`: Update link to correct location

See Go 1.19 release notes for details about gofmt changes: https://tip.golang.org/doc/go1.19#go-doc


milindl:
Thanks for this PR!
I think we'll soon need to update to go1.21 due to certain dependencies requiring it to be updated.
I'll update this PR and merge it then after the 1.21 and dependency update PR is in.

milindl:
I'll merge #1136 first, and then, this. It updates the go version (both the supported and development versions).

Issue #1146, opened 2024-03-04
Extremely slow protobuf deserialization and serialization for schema registry


fzmoment:
Description
===========
When using protobuf schemas we've noticed that the drop in performance is much worse than expected. For example, compared to the baseline of just a regular protobuf Unmarshal call, which takes on the order of <10 us:

![Screenshot 2024-03-01 at 11 16 20 AM](https://github.com/confluentinc/confluent-kafka-go/assets/146033899/72ceda7d-b5b7-4962-963f-7a88e9f3022b)

Using the confluent kafka go library, we see a over 2500x increase in processing time:

![Screenshot 2024-03-01 at 12 41 35 PM](https://github.com/confluentinc/confluent-kafka-go/assets/146033899/f421ba9d-92a1-43f1-bac2-dbf35f13716d)

Furthermore, we know that this decrease is specifically **not** due to communication with the schema registry. In this example, we are deserializing the same schema over and over again and thus only make a request to the registry once before the result is cached - in general, the schema ID lookup takes ~2us:

![Screenshot 2024-03-01 at 12 47 11 PM](https://github.com/confluentinc/confluent-kafka-go/assets/146033899/e18d1a5b-40b3-464e-8e98-9f821ea82c46)

The bottleneck is the [`ToFileDesc`](https://github.com/confluentinc/confluent-kafka-go/blob/d8d41367d60ed909847c82112755d412725ec7c8/schemaregistry/serde/protobuf/protobuf.go#L417) function which is parsing the schema string back to a file descriptor.

![Screenshot 2024-03-01 at 12 44 45 PM](https://github.com/confluentinc/confluent-kafka-go/assets/146033899/1a155f88-a999-4c9b-8a3e-b300b1dd5579)

this can take up on the order of seconds:

![Screenshot 2024-03-01 at 12 49 58 PM](https://github.com/confluentinc/confluent-kafka-go/assets/146033899/d481f9f1-f526-4aba-91d2-f29b0054365a)

Is this a known issue/are there any workarounds? We see similar behavior on the write side (where the bottleneck is the conversion from the file descriptor to a string, **not** looking up the schema ID for a schema given the caching), though not quite as drastic as the above examples. Any pointers would be appreciated.

How to reproduce
================

Run a consumer that uses the confluent kafka go library to deserialize all the messages on a given topic. In our setup, the topic schema only contains one message. Can provide a minimal repro example if this gains some traction.

Checklist
=========
Please provide the following information:

[...]

PR #1128, opened 2024-01-18
Cache parsed file descriptors

rayokota:
Fixes https://github.com/confluentinc/confluent-kafka-go/issues/969

PR #1151, opened 2024-03-14
add CacheSchemas option to Serializer

fzmoment:
This adds code to the Serializer that caches the `SchemaInfo` for each `FileDescriptor`, which improves performance by over an order of magnitude by caching the string manipulation bottleneck (see flamegraph [here](https://github.com/confluentinc/confluent-kafka-go/issues/1146#issuecomment-1983975280)). In order to do this, there was also some refactoring to isolate the "cachable" code (getting SchemaInfo from a proto).

Benchmark results:

```
go test -bench . -benchmem
goos: darwin
goarch: arm64
pkg: github.com/confluentinc/confluent-kafka-go/v2/schemaregistry/serde/protobuf
BenchmarkProtobufSerWithReference-8         	   30640	     38625 ns/op	  21778 B/op	     696 allocs/op
BenchmarkProtobufSerWithReferenceCached-8   	 1000000	      1047 ns/op	   1089 B/op	      10 allocs/op
BenchmarkProtobufDeserWithReference-8       	 2469417	       490.1 ns/op	    304 B/op	       8 allocs/op
PASS
```

This is analogous to the approach take in https://github.com/confluentinc/confluent-kafka-go/pull/1128, but the main difference is that this caching is gated behind a config which defaults to false (i.e. default no behavior change). I believe this optimization can be used by most use cases, but there is a specific scenario where it is not appropriate which is why it is gated behind a config:
* The caching uses the file descriptor as key, which reflects the state of the protobuf file a message comes from
* Thus, the exact same protobuf file/message with a different version of a dependency will result in a duplicate key
* So caching should not be used for applications that are not creating new protobuf messages themselves (e.g. proxying proto messages from some data source), or that explicitly include multiple versions of the same set of bindings

Since this caching can be enabled for most users (generally users are generating one set of protobuf bindings and creating messages based on that and thus there won't be multiple versions of protobufs within one run of the application) I think it's appropriate to have this upstreamed into the confluent-kafka-go library

Test: ran existing tests with config enabled; added benchmarks

### DIFF
--- a/schemaregistry/serde/protobuf/config.go
+++ b/schemaregistry/serde/protobuf/config.go
@@ -21,12 +21,21 @@ import "github.com/confluentinc/confluent-kafka-go/v2/schemaregistry/serde"
 // SerializerConfig is used to pass multiple configuration options to the serializers.
 type SerializerConfig struct {
 	serde.SerializerConfig
+	// CacheSchemas will cache serialization results based on the name of the protobuf file
+	// corresponding to the message being serialized. This will drastically improve serialization
+	// performance if you are only ever using a _single_ version of a specific protobuf schema
+	// during any given run of your application. This should be the case for most applications,
+	// but might not apply if you're not creating proto messages based on generated files (e.g.
+	// you are proxying or reading raw protobuf messages from a data source), or if for some reason
+	// you are including multiple versions of the same schema/protobuf in your application.
+	CacheSchemas bool
 }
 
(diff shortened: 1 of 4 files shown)


rayokota:
Thanks @fzmoment , I'll take a look

rayokota (approved):
Thanks @fzmoment , LGTM!

PR #1162, opened 2024-04-01
return more clearly topic error message

darrkz:
topic error was catch in  ErrInvalidArg
but it can show more clearly

### DIFF
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -173,8 +173,10 @@ func (p *Producer) gethandle() *handle {
 }
 
 func (p *Producer) produce(msg *Message, msgFlags int, deliveryChan chan Event) error {
-	if msg == nil || msg.TopicPartition.Topic == nil || len(*msg.TopicPartition.Topic) == 0 {
+	if msg == nil || len(*msg.TopicPartition.Topic) == 0 {
 		return newErrorFromString(ErrInvalidArg, "")
+	}else if msg.TopicPartition.Topic == nil{
+		return newErrorFromString(ErrUnknownTopic, "topic is null, please check")
 	}
 
 	crkt := p.handle.getRkt(*msg.TopicPartition.Topic)


darrkz:
@milindl  please review , thanks...

kafka/producer.go
```diff
@@ -173,8 +173,10 @@ func (p *Producer) gethandle() *handle {
 }
 
 func (p *Producer) produce(msg *Message, msgFlags int, deliveryChan chan Event) error {
-	if msg == nil || msg.TopicPartition.Topic == nil || len(*msg.TopicPartition.Topic) == 0 {
+	if msg == nil || len(*msg.TopicPartition.Topic) == 0 {
 		return newErrorFromString(ErrInvalidArg, "")
```
milindl, line 177:
```suggestion
	if msg == nil {
		return newErrorFromString(ErrInvalidArg, "msg cannot be nil")
```

kafka/producer.go
```diff
@@ -173,8 +173,10 @@ func (p *Producer) gethandle() *handle {
...
+	}else if msg.TopicPartition.Topic == nil{
+		return newErrorFromString(ErrUnknownTopic, "topic is null, please check")
```
milindl, line 179:
```suggestion
	} else if msg.TopicPartition.Topic == nil || len(*msg.TopicPartition.Topic) == 0 {
		return newErrorFromString(ErrInvalidArg, "topic cannot be nil or empty")
```

milindl: We must keep in mind, 2 things, 
1. msg.TopicPartition.Topic == nil  must be made before len(*msg.TopicPartition.Topic) == 0 , else it causes a nil pointer dereference and will panic.
2. We can't change the type of the error, only the message of it (since the type is a part of the public API so people might be relying on that)

milindl (changes requested):
Thanks a lot for following up with a PR!

I have suggested some changes. Once you are done, could you also run `go fmt` on the changed files? Let me know if you have any questions.

milindl (commented):
Looks pretty good, thanks! Just one more thing, we should run `go fmt` on the changed files

darrkz:
@milindl  please review again and it can be merged

milindl:
Hi @darrkz , please run the `go fmt` command on the codebase to fix the format issues automatically. I can run the CI and merge it after that.

PR #1166, opened 2024-04-05
fix unhandled error in consumer_example.go thank @TheAnthonyGG for this.

nathanwienand:
https://www.twitch.tv/theanthonygg - Subscribe!


Desquaredp (approved):
Looks good.

PR #1178, opened 2024-04-18
Use unsafe.SliceData()

jdemeyer:
Since we require Go 1.21 now, we can use `unsafe.SliceData` which simplifies quite a bit of code involving passing slices to C. I focused on the most important places in the code where such a conversion was done. If this PR is approved, I can do the other places also.

While I was at it, I also replaced `UnixNano()/1000000` by `UnixMilli()` in code I was editing anyway.


milindl:
Great stuff, thanks for this PR. I only skimmed the PR yet, but I do believe it makes sense.

PR #1181, opened 2024-04-20
Add missing API operations to Schema Registry client

alexhein:
1. Add the API path to list contexts: https://docs.confluent.io/platform/current/schema-registry/develop/api.html#get--contexts 
2. Add the API path to retrieve all subject-version pairs associated with a given schema ID: https://docs.confluent.io/platform/current/schema-registry/develop/api.html#get--schemas-ids-int-%20id-versions


alexhein:
@rayokota Can you please take a look?

PR #1185, opened 2024-04-24
[KIP-848] Run integration tests with both the "classic" and "consumer" consumer groups

emasab:


kafka/testresources/kraft/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM debian:bookworm
+
+RUN apt update
+RUN apt install -y git curl
+RUN apt install -y openjdk-17-jdk
+
+RUN git clone https://github.com/apache/kafka.git
```
milindl, line 7: I think we should checkout to a branch or a tag after this for making sure tests don't break at some random point in the future

milindl (approved):
Looks okay except one comment

kafka/testresources/kraft/Dockerfile
emasab: At the moment there's no tag to checkout as 3.7.0 needed some fixes I could leave current commit in trunk and we update it when needed before 3.8.0

emasab:
@milindl sent a commit, please check again

milindl (approved):
The PR itself looks OK barring the KRaft+classic failures

PR #1202, opened 2024-06-03
Add data-governance to code owners for schema registry clients

emasab:


.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
 * @confluentinc/clients
+* @confluentinc/data-governance
```
milindl, line 2: I think it should be `* @confluentinc/clients @confluentinc/data-governance` since the last one takes precedence, so we need to include both at once.

milindl (approved):

PR #1209, opened 2024-06-26
Clarify statement

joel-hamill:


milindl (approved):
approving assuming this is a general change we're doing

PR #1241, opened 2024-07-04
Fix issue in admin_describe_cluster.go

CaelRowley:
The results print was printing a pointer to clusterDesc.ClusterID and not the cluster id itself


rayokota (approved):
LGTM

PR #1250, opened 2024-07-04
Update protobuf_producer_encryption_example.go

CaelRowley:
Remove redundant types schemaregistry.Rule is inferred from `[]schemaregistry.Rule{` Prevent data leaks by ensuring producer and deliveryChan are always closed


rayokota (approved):
LGTM

PR #1267, opened 2024-08-09
[KIP-848 EA] Admin API for listing consumer groups now has 

mahajanadhitya:
an optional filter to return only groups of given types

### DIFF
--- a/kafka/adminapi_test.go
+++ b/kafka/adminapi_test.go
@@ -444,19 +444,6 @@ func testAdminAPIsListConsumerGroups(
 		t.Fatalf("Expected ConsumerGroupStateFromString to work for Stable state")
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), expDuration)
-	defer cancel()
-	listres, err := a.ListConsumerGroups(
-		ctx, SetAdminRequestTimeout(time.Second),
-		SetAdminMatchConsumerGroupStates([]ConsumerGroupState{state}))
-	if err == nil {
-		t.Fatalf("Expected ListConsumerGroups to fail, but got result: %v, err: %v",
-			listres, err)
-	}
-	if ctx.Err() != context.DeadlineExceeded {
-		t.Fatalf("Expected DeadlineExceeded, not %v", ctx.Err())
-	}
-
 	unknownGroupType := ConsumerGroupTypeFromString("Unknown")
 	if unknownGroupType != ConsumerGroupTypeUnknown {
 		t.Fatalf("%s: Expected ConsumerGroupTypeFromString to work for Unknown type", what)
@@ -467,10 +454,11 @@ func testAdminAPIsListConsumerGroups(
 		t.Fatalf("%s: Expected ConsumerGroupTypeFromString to work for Classic type", what)
 	}
 
-	ctx, cancel = context.WithTimeout(context.Background(), expDuration)
+	// Test with unknown group type
+	ctx, cancel := context.WithTimeout(context.Background(), expDuration)
 	defer cancel()
 
-	listres, err = a.ListConsumerGroups(
+	listres, err := a.ListConsumerGroups(
 		ctx, SetAdminRequestTimeout(time.Second),
 		SetAdminMatchConsumerGroupTypes([]ConsumerGroupType{unknownGroupType}))
 
@@ -479,6 +467,7 @@ func testAdminAPIsListConsumerGroups(
 			what, listres, err)
 	}
 
+	// Test with duplicate group type
 	ctx, cancel = context.WithTimeout(context.Background(), expDuration)
 	defer cancel()
 
@@ -489,6 +478,21 @@ func testAdminAPIsListConsumerGroups(
 	if err.(Error).Code() != ErrInvalidArg {
 		t.Fatalf("%s: Expected ListConsumerGroups to fail when duplicate group type options are set, but got result: %v, err: %v",
 			what, listres, err)
+	}
+
+	// Successful call
+	ctx, cancel = context.WithTimeout(context.Background(), expDuration)
+	defer cancel()
+	listres, err = a.ListConsumerGroups(
+		ctx, SetAdminRequestTimeout(time.Second),
+		SetAdminMatchConsumerGroupStates([]ConsumerGroupState{state}),
+		SetAdminMatchConsumerGroupTypes([]ConsumerGroupType{classicGroupType}))
+	if err == nil {
+		t.Fatalf("Expected ListConsumerGroups to fail, but got result: %v, err: %v",
+			listres, err)
+	}
+	if ctx.Err() != context.DeadlineExceeded {
+		t.Fatalf("Expected DeadlineExceeded, not %v", ctx.Err())
 	}
 }
 
--- a/kafka/adminapi_test.go
+++ b/kafka/adminapi_test.go
@@ -456,6 +456,40 @@ func testAdminAPIsListConsumerGroups(
 	if ctx.Err() != context.DeadlineExceeded {
 		t.Fatalf("Expected DeadlineExceeded, not %v", ctx.Err())
 	}
+
+	unknownGroupType := ConsumerGroupTypeFromString("Unknown")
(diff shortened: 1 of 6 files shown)


kafka/adminapi_test.go
```diff
@@ -456,6 +458,38 @@ func testAdminAPIsListConsumerGroups(
 	if ctx.Err() != context.DeadlineExceeded {
 		t.Fatalf("Expected DeadlineExceeded, not %v", ctx.Err())
 	}
+
+	unknownGroupType, err := ConsumerGroupTypeFromString("Unknown")
+	if err != nil || unknownGroupType != ConsumerGroupTypeUnknown {
+		t.Fatalf("Expected ConsumerGroupTypeFromString to work for Unknown type")
+	}
+
+	listres, err := a.ListConsumerGroups(
```
milindl, line 467: Recreate the context before each call to ListConsumerGroups as the context has already expired in the previous call.

kafka/adminapi_test.go
```diff
@@ -456,6 +458,38 @@ func testAdminAPIsListConsumerGroups(
...
+	}
+
+	classicGroupType, err := ConsumerGroupTypeFromString("Classic")
+	if err != nil || unknownGroupType != ConsumerGroupTypeUnknown {
+		t.Fatalf("Expected ConsumerGroupTypeFromString to work for Classic type")
+	}
+
+	listres, err := a.ListConsumerGroups(
+		ctx, SetAdminRequestTimeout(time.Second),
+		SetAdminMatchConsumerGroupTypes([]ConsumerGroupType{classicGroupType, classicGroupType}))
```
milindl, line 485:
```suggestion
		SetAdminMatchConsumerGroupTypes([]ConsumerGroupType{classicGroupType, classicGroupType}))
```
Why is the same group type included twice?

kafka/build_darwin_amd64.go
```diff
@@ -1,6 +1,6 @@
+//go:build !dynamic
```
milindl, line 1: Don't modify this file. I know go fmt modifies this automatically but since our commitment is to support as old a version of go as we possibly can, we avoid it. Same for the other build_ files.
Eventually, one day, we will make the change but not today.

kafka/integration_test.go
```diff
@@ -828,21 +828,135 @@ func (its *IntegrationTestSuite) TestAdminClient_DeleteConsumerGroups() {
...
 		return
 	}
+	return
 }
 
 // TestAdminClient_ListAndDescribeConsumerGroups validates the working of the
 // list consumer groups and describe consumer group APIs of the admin client.
 //
-// We test the following situations:
+// We test the following situations[Classic Protocol]:
```
milindl, line 837: Question: shouldn't we test these three situations both with classic and consumer protocol?
Listing a group with one client, two client, and an empty consumer group all seem to be legitimate operations even with the Consumer Protocol?

kafka/integration_test.go
```diff
@@ -828,21 +828,135 @@ func (its *IntegrationTestSuite) TestAdminClient_DeleteConsumerGroups() {
...
 //
 // 1. One consumer group with one client.
 // 2. One consumer group with two clients.
 // 3. Empty consumer group.
+//
+// We test the following situations[Consumer Protocol]:
+// 1. No Consumer Group should be listed with Classic Group Type Option
+// 2. Our Consumer Group should be listed with Consumer Group Type Option
+// 3. All Consumer Groups should be listed with NULL Group Type Option
```
milindl, line 846:
```suggestion
// 3. All Consumer Groups should be listed with nil Group Type Option
```

kafka/integration_test.go
```diff
@@ -828,21 +828,135 @@ func (its *IntegrationTestSuite) TestAdminClient_DeleteConsumerGroups() {
...
 func (its *IntegrationTestSuite) TestAdminClient_ListAndDescribeConsumerGroups() {
 	t := its.T()
 	if !testConsumerGroupProtocolClassic() {
-		t.Skipf("KIP 848 Admin operations changes still aren't " +
-			"available")
+		// Generating a new topic/groupID to ensure a fresh group/topic is created.
+		rand.Seed(time.Now().Unix())
+		groupID := fmt.Sprintf("%s-%d", testconf.GroupID, rand.Int())
+		topic := fmt.Sprintf("%s-%d", testconf.TopicName, rand.Int())
+		nonExistentGroupID := fmt.Sprintf("%s-nonexistent-%d", testconf.GroupID, rand.Int())
```
milindl, line 854: Move above the "if" block, and remove from below too, so they're not duplicated

milindl: Same for creation of adminclient, topic creation, etc.

milindl (commented):
Thanks for the changes! Looks mostly okay.

Remember to add it to CHANGELOG.md also

kafka/adminapi_test.go
mahajanadhitya: It is to test the INVALID_ARGS error code with duplicate group types.

kafka/integration_test.go
mahajanadhitya: Those have already been taken care by from the previous tests, we are just adding an OPTION for request and response to convey for the group type. I do not see any need of having different clients.

kafka/integration_test.go
mahajanadhitya: sure.

kafka/integration_test.go
mahajanadhitya: Actually this is done voluntarily because currently in semaphore we run the whole tests with broker version < 3.8.0.0 and then when we run with higher broker version on semaphore to test the consumer group protocol we only check the case of consumer group that is why we return at the end.

kafka/build_darwin_amd64.go
mahajanadhitya: sure I will checkout all the build files from the base branch commit.

kafka/adminapi_test.go
mahajanadhitya: sure.

kafka/adminapi_test.go
```diff
@@ -456,6 +461,44 @@ func testAdminAPIsListConsumerGroups(
...
+
+	unknownGroupType, err := ConsumerGroupTypeFromString("Unknown")
+	if err != nil || unknownGroupType != ConsumerGroupTypeUnknown {
+		t.Fatalf("Expected ConsumerGroupTypeFromString to work for Unknown type")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), expDuration)
+	defer cancel()
+ 
+  listres, err := a.ListConsumerGroups(
```
milindl, line 473: gofmt needs to be run

kafka/adminapi_test.go
```diff
@@ -464,7 +467,10 @@ func testAdminAPIsListConsumerGroups(
 		t.Fatalf("Expected ConsumerGroupTypeFromString to work for Unknown type")
 	}
 
-	listres, err := a.ListConsumerGroups(
+	ctx, cancel := context.WithTimeout(context.Background(), expDuration)
```
milindl, line 470: I think this code doesn't run. Are the tests running when you are dynamically linking librdkafka? This is a compilation error because := and not =

kafka/adminapi_test.go
milindl: But your test is not checking for invalid args error code below, so change that. You're changing for deadline exceeded still.

kafka/integration_test.go
milindl: Okay, so since there is nothing common between the tests, why don't you split them? I don't see why we should have an entirely different test in an if block if there's any shared code between them, it just makes it harder to read.

TestAdminClient_ListAndDescribeConsumerGroupsClassic and TestAdminClient_ListConsumerGroupsConsumer

milindl (dismissed):
Some more comments, please run the tests

emasab:
Converting to draft until we merge v2.6.0-RC2

pranavrth (approved):
All doubts cleared over a call.
LGTM!.

PR #1270, opened 2024-08-14
Pinned cp-kafka version

emasab:
as 7.7.0 has bug https://issues.apache.org/jira/browse/KAFKA-16310 that makes LisOffsets test fail


rayokota (approved):
LGTM

Issue #26, opened 2017-01-12
ConfigMap missing Get(), GetKey() methods


mcandre:

edenhill:
.Get() was added on master

PR #1297, opened 2024-09-19
Remove redundant rest_service.go (#26)

rayokota:


zhaochun-ma (approved):

oleksandrkorzhenevskyi:
@rayokota RestError is in the internal package now, so we cannot access its properties when use the schema registry client.
```
// RestError represents a Schema Registry HTTP Error response
type RestError struct {
	Code    int    `json:"error_code"`
	Message string `json:"message"`
}
```
Is there another way to check Code of this struct without accessing it directly (which is impossible after this PR)?

PR #1323, opened 2024-10-27
Add go.mod in kafka directory

acgtun:
Add modular in kafka directory, so that we could import only kafka modular without the other modulars.


milindl:
Hi @acgtun , I think this idea is worth looking into in more details and separate other modules (schemaregistry) but at the current time, I think this will break backwards compatibility by requiring every user to change their import statements (v2/kafka changes to kafka/v2). So we can't merge this immediately. 
I'll discuss this internally to see if this is something we do with our next major version upgrade whenever it is planned.

acgtun:
Thanks @milindl for looking into it.

acgtun:
Hi @milindl, for us, the main reason we want to create a kafka modular is to reduce some of the dependency requirements.
for example, github.com/docker/docker , k8s.io/cri, etc, these that are not used in kafka modular.

jdemeyer:
> so that we could import only kafka modular without the other modulars

Can you explain what you mean with this? Imports in Go work on the package level (where a package is a directory), so adding a `go.mod` won't change anything regarding imports. What is the actual problem that you're trying to solve here?

PR #1324, opened 2024-10-28
Support transforming maps in Data Contract rules

rayokota:
Support transforming maps in Data Contract rules.  Previously only transforming fields in structs were supported.  This PR adds ability to transform entries in maps.

### DIFF
--- a/schemaregistry/serde/jsonschema/json_schema_util.go
+++ b/schemaregistry/serde/jsonschema/json_schema_util.go
@@ -78,18 +78,31 @@ func transform(ctx serde.RuleContext, schema *jsonschema2.Schema, path string, m
 	switch typ {
 	case serde.TypeRecord:
 		val := deref(msg)
-		fieldByNames := fieldByNames(val)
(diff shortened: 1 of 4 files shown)


schemaregistry/serde/jsonschema/json_schema_util.go
```diff
@@ -238,6 +255,9 @@ func disjoint(slice1 []string, map1 map[string]bool) bool {
 }
 
 func getField(msg *reflect.Value, name string) (*reflect.Value, error) {
+	if msg.Kind() != reflect.Struct {
+		return nil, fmt.Errorf("message is not a struct")
```
kcorman0, line 259: why not do the same thing as above
`val.MapIndex(reflect.ValueOf(propName)`

rayokota: Good question.  This method is only for operating on Structs

kcorman0 (commented):
LGTM assuming build pass

kcorman0 (approved):

PR #1336, opened 2024-11-16
librdkafka v2.6.1-RC2

emasab:


pranavrth:
I think we should increase https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/00version.go#L32 as well. We are fix few bug and one of them is significant.

pranavrth (approved):
LGTM!. Thanks.

PR #1337, opened 2024-11-18
Add retry logic to RestService

rayokota:
Add retry logic to RestService.  We implement exponential backoff with full jitter per https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/


kcorman0 (approved):
LGTM thanks

PR #1340, opened 2024-11-19
Add deleted query param when looking up schema under subject

rayokota:
Add deleted query param when looking up schema under subject, which should be false during serialization and true during deserialization.


Claimundefine (approved):
LGTM

PR #1344, opened 2024-11-21
KIP 848:Extend DescribeConfigs and IncrementalAlterConfigs to support GROUP Config

PratRanj07:
This PR intends to add support for Group Resource type for IncrementalALterConfigs API as specified in KIP 848 and extended the DescribeConfigs support till API version 3.
Additional changed the integration test to check for Group Resource Type.
Librdkafka PR: https://github.com/confluentinc/librdkafka/pull/4939

### DIFF
--- a/kafka/integration_test.go
+++ b/kafka/integration_test.go
@@ -2016,6 +2016,8 @@ func (its *IntegrationTestSuite) TestAdminConfig() {
 		t.Fatalf("Failed to delete topic %s: %s", topic, topicResult[0].Error)
 	}
 
+	// TODO: enable this test for the classic run too, when
+	// Confluent Platform test cluster is upgraded to 8.0.0
 	if !testConsumerGroupProtocolClassic() {
 		// Test configs that are configured for a specific group.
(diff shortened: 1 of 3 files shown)


kafka/integration_test.go
```diff
@@ -2007,6 +2007,63 @@ func (its *IntegrationTestSuite) TestAdminConfig() {
 	if topicResult[0].Error.Code() != ErrNoError {
 		t.Fatalf("Failed to delete topic %s: %s", topic, topicResult[0].Error)
 	}
+
+	if !testConsumerGroupProtocolClassic() {
+		// New test case for ResourceType.Group (ResourceGroup)
```
milindl, line 2012: Don't add this comment, it's only "new" while it's in this PR. 
Just add a comment as "Test configs that are configured for a specific group."

milindl (commented):
Add a CHANGELOG.md entry for this. 

I will finally approve only after doing the RC but otherwise looks good.

emasab (approved):

PR #1347, opened 2024-11-26
For Protobuf, copy the result into the target if necessary

rayokota:
The copying is only needed for the DeserializeInto method for Protobuf.


Claimundefine (approved):
LGTM

PR #1369, opened 2024-12-20
fix(conf): improve type conversion

ekazakas:
What
----
Slightly improve type conversion in ConfigMap SetKey. If the type is not what we expect, it should return an error.

Checklist
------------------
No breaking changes

References
----------
JIRA: 
N/A

Test & Review
------------
Tested using available unit tests and added new unit tests scenarios

Open questions / Follow-ups
--------------------------
N/A

<!--
Review stakeholders
------------------
N/A


ekazakas:
Greetings @rayokota would it be possible for you or someone else to review this PR?

PR #1407, opened 2025-03-31
Add support for RD_KAFKA_RESOURCE_TRANSACTIONAL_ID in adminapi

jzvelc:
We need to set transactional.id ACL permissions with adminapi. ResourceTransactionalID is currently missing in golang implementation despite C.RD_KAFKA_RESOURCE_TRANSACTIONAL_ID is available. In scope of this PR I am adding support for ResourceTransactionalID.

### DIFF
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -498,6 +498,8 @@ const (
 	ResourceUnknown ResourceType = C.RD_KAFKA_RESOURCE_UNKNOWN
 	// ResourceAny - match any resource type (DescribeConfigs)
 	ResourceAny ResourceType = C.RD_KAFKA_RESOURCE_ANY
+	// ResourceTransactionalID - Transactional ID
+	ResourceTransactionalID ResourceType = C.RD_KAFKA_RESOURCE_TRANSACTIONAL_ID
 	// ResourceTopic - Topic
 	ResourceTopic ResourceType = C.RD_KAFKA_RESOURCE_TOPIC
 	// ResourceGroup - Group
@@ -517,6 +519,8 @@ func ResourceTypeFromString(typeString string) (ResourceType, error) {
 	switch strings.ToUpper(typeString) {
 	case "ANY":
 		return ResourceAny, nil
+	case "TRANSACTIONAL_ID":
+		return ResourceTransactionalID, nil
 	case "TOPIC":
 		return ResourceTopic, nil
 	case "GROUP":
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -498,14 +498,14 @@ const (
 	ResourceUnknown ResourceType = C.RD_KAFKA_RESOURCE_UNKNOWN
 	// ResourceAny - match any resource type (DescribeConfigs)
 	ResourceAny ResourceType = C.RD_KAFKA_RESOURCE_ANY
-	// ResourceTransactionalID - Transactional ID
-	ResourceTransactionalID ResourceType = C.RD_KAFKA_RESOURCE_TRANSACTIONAL_ID
 	// ResourceTopic - Topic
 	ResourceTopic ResourceType = C.RD_KAFKA_RESOURCE_TOPIC
 	// ResourceGroup - Group
 	ResourceGroup ResourceType = C.RD_KAFKA_RESOURCE_GROUP
 	// ResourceBroker - Broker
 	ResourceBroker ResourceType = C.RD_KAFKA_RESOURCE_BROKER
+	// ResourceTransactionalID - Transactional ID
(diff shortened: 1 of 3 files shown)


kafka/adminapi.go
```diff
@@ -493,6 +493,8 @@ const (
 	ResourceUnknown ResourceType = C.RD_KAFKA_RESOURCE_UNKNOWN
 	// ResourceAny - match any resource type (DescribeConfigs)
 	ResourceAny ResourceType = C.RD_KAFKA_RESOURCE_ANY
+	// ResourceTransactionalID - Transactional ID
+	ResourceTransactionalID ResourceType = C.RD_KAFKA_RESOURCE_TRANSACTIONAL_ID
```
Devarsh010: Nit (non-blocking): librdkafka declares these in numeric order (`UNKNOWN=0, ANY=1, TOPIC=2, GROUP=3, BROKER=4, TRANSACTIONAL_ID=5`). Could you move this line to after `ResourceBroker` so the Go const block mirrors the C header? Easier to keep in sync if more types are ever added.

kafka/adminapi.go
```diff
@@ -512,6 +514,8 @@ func ResourceTypeFromString(typeString string) (ResourceType, error) {
 	switch strings.ToUpper(typeString) {
 	case "ANY":
 		return ResourceAny, nil
+	case "TRANSACTIONAL_ID":
```
Devarsh010: Same nit — move this `case` to after `case "BROKER":` for consistency with the const ordering above.

kafka/integration_test.go
```diff
@@ -2143,6 +2143,15 @@ func (its *IntegrationTestSuite) TestAdminACLs() {
...
 		},
+		{
+			Type:                ResourceTransactionalID,
+			Name:                group,
+			ResourcePatternType: ResourcePatternTypePrefixed,
+			Principal:           "User:test-user-2",
+			Host:                "*",
+			Operation:           ACLOperationAll,
+			PermissionType:      ACLPermissionTypeAllow,
+		},
```
Devarsh010: Suggestion: move this binding to the **end** of `newACLs` rather than inserting it after the first Topic binding.

`ACLBindings.Less()` (`adminapi.go:919`) sorts by `ResourceType` first, and `ResourceTransactionalID = 5` is greater than `Topic = 2` and `Group = 3`. With the binding inserted in the middle, the declaration order no longer matches the sort order — which breaks the two `checkExpectedResult` assertions further down (see comments below). Putting it last keeps declaration order = sort order and avoids needing to sort `expected` everywhere.

kafka/integration_test.go
Devarsh010, line 2323: With the new binding inserted in the middle, `newACLs` is no longer in sort order, but `resultDescribeACLs.ACLBindings` is sorted via `sort.Sort` on the next line. `reflect.DeepEqual` will fail.

If you move the new binding to the end of `newACLs` (per the suggestion above), this line stays correct as-is.

Devarsh010: Bug: with the new ACL added, there are now **three** prefixed bindings (TransactionalID, Topic-Prefixed, Group-Prefixed), so this slice is missing one entry. The `Any/Prefixed` filter at `aclBindingFilters[1]` will delete all three.

If you move the new binding to the end of `newACLs` (so the order becomes `[Topic-Literal, Topic-Prefixed, Group-Prefixed, TransactionalID-Prefixed]`), this should become `newACLs[1:4]`.

Devarsh010 (dismissed):
Thanks for the patch, @jzvelc and apologies for the delay in getting eyes on this. The gap is real: `RD_KAFKA_RESOURCE_TRANSACTIONAL_ID` has been a first-class ACL resource since KIP-98, librdkafka has exposed it for years (`rdkafka.h`), and the Python client surfaces it. This is a clean parity fix.

The production change in `kafka/adminapi.go` is correct:
- The `C.RD_KAFKA_RESOURCE_TRANSACTIONAL_ID` constant resolves to `5`, matching the wire protocol.
- `"TRANSACTIONAL_ID"` is symmetric with what `rd_kafka_ResourceType_name()` returns (`src/rdkafka_admin.c`), and matches Java/Python casing — important since ACL configs often get shared across clients.

The integration test changes have a few bugs that would fail under `-docker.needed`, though — flagged inline. The simplest fix is to move the new binding to the end of `newACLs` so declaration order matches the sort order produced by `ACLBindings.Less()` (which sorts by `ResourceType` first, and `TransactionalID = 5` is the largest value). That naturally fixes both the `DescribeACLs` deep-equal and the prefixed-`DeleteACLs` slice bounds.

Before merge, could you also:
1. Rebase onto current `master` 
2. Add a one-line entry under `Enhancements` in `CHANGELOG.md` for the next release.

Non-blocking nit on constant ordering also flagged inline.

emasab (approved):

PR #1413, opened 2025-04-10
Add Go OAuth

Claimundefine:
What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


schemaregistry/internal/authentication_header_provider.go
```diff
@@ -0,0 +1,130 @@
...
+	"time"
+
+	"golang.org/x/oauth2"
+)
+
+const tokenExpiryThreshold = 0.8
+
+// AuthenticationHeaderProvider is an interface that provides a method to set authentication headers.
+type AuthenticationHeaderProvider interface {
+	SetAuthenticationHeaders(header *http.Header) error
```
rayokota, line 34: Change to 
```
GetAuthenticationHeader() (string, error)
```
The method `SetAuthenticationHeaders` allows any header to be set, whereas `GetAuthenticationHeader()` has a narrower focus.

schemaregistry/internal/rest_service.go
```diff
@@ -155,14 +157,73 @@ func NewRestService(conf *ClientConfig) (*RestService, error) {
...
 	return &RestService{
-		urls:             urls,
-		headers:          headers,
-		maxRetries:       conf.MaxRetries,
-		retriesWaitMs:    conf.RetriesWaitMs,
-		retriesMaxWaitMs: conf.RetriesMaxWaitMs,
-		ceilingRetries:   int(math.Log2(float64(conf.RetriesMaxWaitMs) / float64(conf.RetriesWaitMs))),
-		Client:           conf.HTTPClient,
+		urls:                         urls,
+		headers:                      headers,
```
rayokota, line 220: It's a bit messy that the auth headers are already set in the case of basic auth, whereas for OAuth we use the `authenticationHeaderProvider`.  It would be cleaner if the only headers that were set were possibly the `target-sr-cluster` and the `Confluent-Identity-Pool-Id`, and then the `authenticationHeaderProvider` would be one that could set the auth header for both basic auth and OAuth.  WDYT?

rayokota (commented):
Thanks @Claimundefine , left a couple of initial comments

rayokota (approved):
Thanks @Claimundefine , LGTM

Issue #1467, opened 2025-08-21
DeserializeInto panic when payload is empty


felixputera:
Follow up from #1382.

Currently, DeserializeInto can still panic even after the fix that closed that issue: 

https://github.com/confluentinc/confluent-kafka-go/blob/4c815e5de09bfc1813ada620ad46bfd6433bb247/schemaregistry/serde/protobuf/protobuf.go#L528-L551

See how `len(payload) == 0` will make `deserialize` return `nil, nil` (both result & error nil). Then the code still won't return early in this case and `rv.Set(reflect.ValueOf(result).Elem()) ` with nil result that was referenced in the earlier issue will still cause a panic.

PR #1468, opened 2025-08-21
Fix JSON validation during deserialization

rayokota:
What
----

Checklist
------------------
- [x] Contains customer facing changes? Including API/behavior changes 

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/schemaregistry/serde/jsonschema/json_schema.go
+++ b/schemaregistry/serde/jsonschema/json_schema.go
@@ -275,11 +275,21 @@ func (s *Deserializer) deserialize(topic string, headers []kafka.Header, payload
 		return nil, err
 	}
 	if s.validate {
+		raw, err := json.Marshal(msg)
+		if err != nil {
+			return nil, err
+		}
+		// Need to unmarshal to pure interface
+		var obj interface{}
+		err = json.Unmarshal(raw, &obj)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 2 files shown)


schemaregistry/serde/protobuf/protobuf.go
```diff
@@ -533,7 +533,7 @@ func (s *Deserializer) DeserializeInto(topic string, payload []byte, msg interfa
 // DeserializeWithHeadersInto implements deserialization of Protobuf data to the given object
 func (s *Deserializer) DeserializeWithHeadersInto(topic string, headers []kafka.Header, payload []byte, msg interface{}) error {
 	result, err := s.deserialize(topic, headers, payload, msg)
-	if err != nil {
+	if result == nil || err != nil {
```
Claimundefine, line 536: Should an error be returned from deserialize if result is nil instead?

Claimundefine (approved):

schemaregistry/serde/protobuf/protobuf.go
felixputera: I raised this issue on #1467. There's a case when both can be `nil`:

```
func (s *Deserializer) deserialize(topic string, headers []kafka.Header, payload []byte, result interface{}) (interface{}, error) {
	if len(payload) == 0 {
		return nil, nil
	}
```

Issue #1421, opened 2025-04-22
Incorrect documentation on handleRebalanceEvent


fgm:
The end of the method comment appears to be wrong.  It says:

```
// In the polling case (not channel based consumer) the rebalance event
// is returned in retval, else nil is returned.
```

While the code actually returns a non-`nil` event only in the channel case, which is the exact opposite of the comment. 
- lines 1022-1025: creates `ev` if there is a `rebalanceCb`  or `appRebalanceEnable`
- lines 1029-1034 : returns it if `channel != nil && c.appRebalanceEnable && c.rebalanceCb == nil` which is the opposite of the documentation
- line 1046: return `nil` if there is a `rebalanceCb` and it called `Assign/Unassign`
- line 1088: return nil unconditionally

It seems that:

-  either the doc should read like this instead:

        // In the polling case (not channel based consumer) the rebalance event
        // is passed to the `rebalanceCb` if it is set, and never returned to `eventPoll`
- or the function logic should be amended to match the comment and allow the method to return the event to `eventPoll`. 

The latter would provide most consistent handling, but would make it easier for callers to introduce bugs in their code because then the protection logic built when calling `rebalanceCb` would have no equivalent for the processing in `eventPoll`.

Do you want a PR either way ?

(tangentially related to #624 / #625)

PR #1472, opened 2025-09-04
consumer: return rebalance event via Poll() when appRebalanceEnable && no RebalanceCb

### LINKED ISSUE #1457: Bug: When appRebalanceEnable == true and  rebalanceCb == nil  and channel == nil,  it shound return ev
When `appRebalanceEnable == true and  rebalanceCb == nil  and channel == nil`, we want handle the event by `poll` , so the rebalance event should return by `poll` but not nil.

link: <https://github.com/confluentinc/confluent-kafka-go/blob/1b46fb97a489f12abc859700d6037660e88bd537/kafka/consumer.go#L1029>

fix example:
```golang
	if (channel != nil || c.rebalanceCb == nil) && c.appRebalanceEnable {
		// Channel-based consumer with rebalancing enabled,
		// return the rebalance event and rely on the  application
		// to call *Assign() / *Unassign().
		return ev
	}
```

aryanrahar:
What

Return the rebalance event from `handleRebalanceEvent` when
`go.application.rebalance.enable == true` **and** no `RebalanceCb` is set,
regardless of whether an events channel is used.

Why: In this configuration, Poll()-based consumers did not receive the
rebalance event and could hang until a callback was wired. This change makes
Poll()-based apps receive `AssignedPartitions` / `RevokedPartitions` so they
can call `Assign()` / `Unassign()` to ack, aligning behavior with the config’s
intent. No API changes; callback path unchanged.

Checklist

- [x] Contains customer facing changes? Including API/behavior changes  
      (Behavioral bug fix only; no API change; backward-compatible.)
      (No new unit test in this PR; integration test typically requires a broker.
       See “Test & Review” for manual steps. I can follow up with a tagged
       integration test if desired.)

References

Fixes #1457

Test & Review

Local sanity:
- `go build ./...` ✅
- `go test ./...` ✅ (unit tests; no broker required)

Manual verification (requires reachable Kafka):
1. Create a consumer with:
"bootstrap.servers": "<brokers>",
"group.id": "rebalance-poll-it",
"go.application.rebalance.enable": true
(No `RebalanceCb`; use `Poll()` — no events channel.)

2. `Subscribe()` to a topic and call `Poll()` in a loop.

3. **Before** this change: no rebalance event arrives via `Poll()`.  
**After** this change: an `AssignedPartitions`/`RevokedPartitions` event is
returned by `Poll()`.

4. Call `Assign()` / `Unassign()` accordingly; consumer proceeds and closes cleanly.

Open questions / Follow-ups

- I can send a tiny follow-up PR to update the comment above
`handleRebalanceEvent` to explicitly document this behavior (refs #1421).
- If maintainers prefer, I can also add a small integration test behind a build
tag (requires `kafka/testconf.json` and a broker).

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -1026,7 +1026,7 @@ func (c *Consumer) handleRebalanceEvent(channel chan Event, rkev *C.rd_kafka_eve
 
 	}
 
-	if channel != nil && c.appRebalanceEnable && c.rebalanceCb == nil {
+	if c.appRebalanceEnable && c.rebalanceCb == nil {
 		// Channel-based consumer with rebalancing enabled,
 		// return the rebalance event and rely on the application
 		// to call *Assign() / *Unassign().


aryanrahar:
?

PR #1485, opened 2025-10-13
SR Associations mock client impl

xhan-cflt:
What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/schemaregistry/schemaregistry_client_test.go
+++ b/schemaregistry/schemaregistry_client_test.go
@@ -396,7 +396,7 @@ func TestAssociations(t *testing.T) {
 		// Pre-create subjects used for testing
 		testValueSubject := "testValue"
 		schemaInfo := SchemaInfo{
-			Schema: "{\"namespace\":\"basicavro\",\"type\":\"struct\",\"name\":\"Payment\",\"fields\":[{\"type\":\"string\",\"name\":\"id\"}]}",
+			Schema: "{\"namespace\":\"basicavro\",\"type\":\"record\",\"name\":\"Payment\",\"fields\":[{\"type\":\"string\",\"name\":\"id\"}]}",
 		}
 		_, err = client.Register(testValueSubject, schemaInfo, true)
 		maybeFail("Register schema", err)
@@ -419,7 +419,7 @@ func TestAssociations(t *testing.T) {
 
 		// Make an association with an existing subject with new schema
 		updatedSchemaInfo := SchemaInfo{
-			Schema: "{\"namespace\":\"basicavro\",\"type\":\"struct\",\"name\":\"Payment\",\"fields\":[{\"type\":\"string\",\"name\":\"id\"}, {\"type\":\"string\",\"name\":\"id2\"}]}",
+			Schema: "{\"namespace\":\"basicavro\",\"type\":\"record\",\"name\":\"Payment\",\"fields\":[{\"type\":\"string\",\"name\":\"id\"}, {\"type\":\"string\",\"name\":\"id2\"}]}",
 		}
 		createRequest = AssociationCreateRequest{ResourceName: "test", ResourceNamespace: "lkc1", ResourceID: "test-id",
 			Associations: []AssociationCreateInfo{AssociationCreateInfo{Subject: testValueSubject, Schema: &updatedSchemaInfo}}}
@@ -449,7 +449,7 @@ func TestAssociations(t *testing.T) {
 		keySubject, valueSubject := "test1Key", "test1Value"
 		resourceName, resourceID := "test1", "test1-id"
 		schemaInfo := SchemaInfo{
-			Schema: "{\"namespace\":\"basicavro\",\"type\":\"struct\",\"name\":\"Payment\",\"fields\":[{\"type\":\"string\",\"name\":\"id\"}]}",
+			Schema: "{\"namespace\":\"basicavro\",\"type\":\"record\",\"name\":\"Payment\",\"fields\":[{\"type\":\"string\",\"name\":\"id\"}]}",
 		}
 		// Pre-create subjects
 		_, err = client.Register(keySubject, schemaInfo, true)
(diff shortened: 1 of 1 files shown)


schemaregistry/mock_schemaregistry_client.go
```diff
@@ -59,22 +67,34 @@ type metadataCacheEntry struct {
 	softDeleted bool
 }
 
+type resourceIdentityAndAssociationType struct {
```
rayokota:
```suggestion
type resourceIDAndAssocType struct {
```

schemaregistry/mock_schemaregistry_client.go
```diff
@@ -59,22 +67,34 @@ type metadataCacheEntry struct {
...
+	idToSchemaCacheLock                         sync.RWMutex
+	guidToSchemaCache                           map[string]infoCacheEntry
+	guidToSchemaCacheLock                       sync.RWMutex
+	schemaToVersionCache                        map[subjectJSON]versionCacheEntry
+	schemaToVersionCacheLock                    sync.RWMutex
+	configCache                                 map[string]ServerConfig
+	configCacheLock                             sync.RWMutex
+	subjectToAssocCache                         map[string][]*Association
+	subjectToAssocCacheLock                     sync.RWMutex
+	resourceIdentityAndAssociationTypeCache     map[resourceIdentityAndAssociationType]*Association
```
rayokota:
```suggestion
	resourceIDAndAssocTypeCache     map[resourceIDAndAssocType]*Association
```

schemaregistry/mock_schemaregistry_client.go
```diff
@@ -19,17 +19,25 @@ package schemaregistry
...
+
 	"github.com/confluentinc/confluent-kafka-go/v2/schemaregistry/internal"
 )
 
 const noSubject = ""
+const defaultResourceType = "topic"
+const defaultAssociationType = "value"
+const defaultLifecyclePolicy = STRONG
+
+var validResourceTypesAndAssociationTypesMap = map[string][]string{
```
rayokota:
```suggestion
var validResourceTypesAndAssocTypesMap = map[string][]string{
```

schemaregistry/mock_schemaregistry_client.go
```diff
@@ -59,22 +67,34 @@ type metadataCacheEntry struct {
...
+	guidToSchemaCache                           map[string]infoCacheEntry
+	guidToSchemaCacheLock                       sync.RWMutex
+	schemaToVersionCache                        map[subjectJSON]versionCacheEntry
+	schemaToVersionCacheLock                    sync.RWMutex
+	configCache                                 map[string]ServerConfig
+	configCacheLock                             sync.RWMutex
+	subjectToAssocCache                         map[string][]*Association
+	subjectToAssocCacheLock                     sync.RWMutex
+	resourceIdentityAndAssociationTypeCache     map[resourceIdentityAndAssociationType]*Association
+	resourceIdentityAndAssociationTypeCacheLock sync.RWMutex
```
rayokota:
```suggestion
	resourceIDAndAssocTypeCacheLock sync.RWMutex
```

schemaregistry/schemaregistry_client.go
```diff
@@ -1029,6 +1039,293 @@ func (c *Compatibility) ParseString(val string) error {
 	return fmt.Errorf("failed to unmarshal Compatibility")
 }
 
+// LifecyclePolicy represents the lifecycle policy for an association
+type LifecyclePolicy string
+
+const (
+	// STRONG LifecyclePolicy
+	STRONG LifecyclePolicy = "strong"
```
rayokota:
```suggestion
	STRONG LifecyclePolicy = "STRONG"
```

schemaregistry/schemaregistry_client.go
```diff
@@ -1029,6 +1039,293 @@ func (c *Compatibility) ParseString(val string) error {
...
+	// WEAK LifecyclePolicy
+	WEAK LifecyclePolicy = "weak"
```
rayokota:
```suggestion
	WEAK LifecyclePolicy = "WEAK"
```

schemaregistry/mock_schemaregistry_client.go
xhan-cflt: For this struct naming, I was originally thinking about using "resourceIDAndAssocType". However because a resource is uniquely identified by "resourceID + resourceType", the struct has both fields. If we use the name "resourceIDAndAssocType", that's kind of implying that the map only contains "resourceID". What do you think?

schemaregistry/schemaregistry_client.go
xhan-cflt: For LifecyclePolicy, we discussed before that we want users to provide lower case "strong" and "weak", so I set the const to be lower case for easy comparison.
If we want to use the capital const, I can convert the input to upper case first before comparison. In this case, we don't check whether it is lower case or upper case in the request. Does that sound good to you?

schemaregistry/mock_schemaregistry_client.go
rayokota: maybe `resourceAndAssocType` then?  Shorter is better than super-accurate, esp since this is internal

schemaregistry/schemaregistry_client.go
rayokota: I think I didn't state things clearly in our last discussion.  Right now the backend expects uppercase for enums.  Also see `compatibilityEnum` in the same file, which are all uppercase

xhan-cflt: Sounds good. So we expect users to provide upper case for lifecycle policy, and if they provide lower case, the request will fail right?

rayokota: Yes

PR #1495, opened 2025-12-04
Add Confluent-Client-Version header to requests to SR, Golang client

fangnx:
What
----

Checklist
------------------

References
----------
JIRA: https://confluentinc.atlassian.net/browse/DGS-22770

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/schemaregistry/internal/rest_service.go
+++ b/schemaregistry/internal/rest_service.go
@@ -36,6 +36,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
 	"github.com/confluentinc/confluent-kafka-go/v2/schemaregistry/rest"
 	"golang.org/x/oauth2/clientcredentials"
 )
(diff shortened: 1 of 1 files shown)


schemaregistry/internal/rest_service.go
```diff
@@ -36,6 +36,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
```
rayokota, line 39: Can we avoid the dependency of the SR client on Kafka?

naxinf: Go client version is currently tied to librdkafka version, so we need to import Kafka to use `C.rd_kafka_version()`   to get the version from librdkafka C binding. We can either:
1. Use a hardcoded constant to record client version in SR module, and add notes to manually bump it
2. Keep it as it is, as SR client already has dependency on Kafka: https://github.com/confluentinc/confluent-kafka-go/blob/bb44f4b44ab8bc1d772b6167c432800cea7a9eda/schemaregistry/serde/avro/avro_generic.go#L21

Personally I prefer 2. Manual bumping is hard to maintain and it's easily overlooked

rayokota (approved):
Thanks @fangnx , LGTM

PR #1512, opened 2026-02-12
schemaregistry/rules/encryption: fix dropped errors

alrs:
This fixes three dropped errors in `schemaregistry/rules/encryption`.

`go test -v ./...` continues to pass, though test coverage in this module is sparse.

The re-ordering of imports is a feature of my editor. No imports were added or removed.


rayokota (approved):
Thanks @alrs , LGTM

PR #1552, opened 2026-05-18
fix(jsonschema): validate deserialized payload against reader schema instead of writer schema

maldimirov:
Description
===========
Validate the deserialized JSON payload against the reader JSON schema instead of the writer JSON schema.

Deserialization is on the "reader side", so validating against the writer schema is just plain wrong.
More info in the ticket: https://github.com/confluentinc/confluent-kafka-go/issues/1553

Checklist
------------------
- [x] Contains customer facing changes? Including API/behavior changes: contains behavior changes, but no API changes
  - Added a new unit test that covers a schema evolution scenario that fails before the proposed fix an passes after it.

References
----------
Github ticket: https://github.com/confluentinc/confluent-kafka-go/issues/1553

Test & Review
------------
Repro/manual testing is described in the ticket: https://github.com/confluentinc/confluent-kafka-go/issues/1553

### DIFF
--- a/schemaregistry/serde/jsonschema/json_schema.go
+++ b/schemaregistry/serde/jsonschema/json_schema.go
@@ -334,7 +334,7 @@ func (s *Deserializer) deserialize(topic string, headers []kafka.Header, payload
 		if err != nil {
 			return nil, err
 		}
-		jschema, err := s.toJSONSchema(s.Client, info)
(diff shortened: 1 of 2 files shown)


rayokota (approved):
Thanks @maldimirov , LGTM

PR #1582, opened 2026-07-22
Fix inconsistent Go version in README and remove dormant base.yml workflow 

### LINKED ISSUE #1580: Go version is inconsistent across README (1.17+), base.yml workflow (1.14), and go.mod (1.25); base.yml is dormant (0 runs, real CI is Semaphore)
Hi, and thank you for confluent-kafka-go.

Three related Go-version inconsistencies I noticed at HEAD (`a6afd32`), grouped as one issue since they share a root cause and the right fix is your call:

1. **README vs go.mod.** `README.md:140` says "Supports Go 1.17+", but `go.mod` declares `go 1.25.0`. A `go 1.25.0` directive can't be built by toolchains older than 1.21 (no auto-toolchain-download), so "1.17+" isn't achievable on master. (Separately, line 266 mentions Go 1.24.3+ for FIPS — also inconsistent with "1.17+".)

2. **`.github/workflows/base.yml` is dormant and internally contradictory.** It pins `setup-go@v3` with `go-version: '=1.14'` (comment: "the lowest supported version") and `setup-python@v4` with Python `'3.8'` (EOL). A Go 1.14 toolchain cannot build a `go 1.25.0` module, so if this workflow ran it would fail at `go build`. It also has **0 retained runs** — the real CI is Semaphore (`.semaphore/`).

3. So the effective supported-Go floor is ambiguous across three sources (README 1.17, base.yml 1.14, go.mod 1.25).

**Why an issue, not a PR:** the fix depends on decisions only you can make — is `base.yml` meant to still run (update it) or is it vestigial now that CI is Semaphore (delete it)? and what Go floor do you actually want to claim in the README? Tell me the intended answers and I'll send whichever PR matches (update-or-remove the workflow, correct the README line).

For transparency: I used AI assistance in preparing this; I verified go.mod, base.yml, its 0-run history, and the README lines myself.

Ankith-Confluent:
Resolves #1580.

- **README.md**: "Go 1.17+" → "Go 1.25+" to match the `go 1.25.0` directive in `go.mod`.
- **Remove `.github/workflows/base.yml`**: dormant (0 runs), pinned Go 1.14 (can't build a `go 1.25.0` module); CI runs on Semaphore.


tobiogunbi (approved):
thanks for opening @Ankith-Confluent

ojasvajain:
@Ankith-Confluent There is one more mention in kafka/README.md. Let's fix that as well.

ojasvajain (approved):
Thanks, approved.

PR #1583, opened 2026-07-23
Integrate Schema Registry serdes into librdkafka producer/consumer (Go)

emasab:
## Summary

Integrates the Schema Registry (SR) serializer/deserializer clients directly into the librdkafka-based `Producer`/`Consumer`, so serialization and deserialization happen transparently inside the client instead of being wired up by the application. Keys and values are (de)serialized automatically as part of produce/poll, using a fluent, interface-based builder API.

## What's new

- **Generic client wrappers** — `SerializingProducer[K, V]` and `DeserializingConsumer[K, V]` wrap `*Producer`/`*Consumer`, delegate the full public surface, and add transparent (de)serialization.
- **Serde + builder interfaces** — `Serializer`/`Deserializer` and `SerializerBuilder`/`DeserializerBuilder` interfaces (`Build(conf, isKey)`), so any builder can be passed to the wrapper constructors.
- **Typed message types** — `SerializableMessage[K, V]` (produce + delivery report) and `DeserializedMessage[K, V]` (poll result).
- **Typed deserialization errors** — `KeyDeserializationError` / `ValueDeserializationError` events returned from `Poll` so the app can seek past bad records.
- **Unified SR config** — `schemaregistry.NewConfigFromKafkaConfigMap(srConf, conf)` derives an SR `Config` from the Kafka `ConfigMap` (pass `nil` to derive from `schema.registry.url`, or a pre-built config for auth/TLS) and returns the ConfigMap with SR-only keys filtered out.
- **Fluent serde builders** — `KafkaSerializerBuilder`/`KafkaDeserializerBuilder` in `serde/avrov2`, `serde/avrov3`, `serde/jsonschema`, `serde/protobuf`, with chainable `SetSchemaRegistryConfig` / `SetSerializerConfig`(`SetDeserializerConfig`) / `SetSerializerInit`(`SetDeserializerInit`), `Build`, and no-arg `NewKafka…Builder()` constructors.
- **Reusable SR client** — `SetSchemaRegistryClient()` on every builder, so an application that already has a `schemaregistry.Client` (or wants to share one across serdes) can hand it over instead of having each builder create its own. When a client is supplied, `Build()` uses it as-is and passes the Kafka `ConfigMap` through unfiltered.
- **Cluster ID propagation** — `NeedsClusterID`/`SetClusterID` are now part of both serde interfaces, implemented once on `serde.Serde`. `NeedsClusterID` reports true only for the associated name strategy — the one strategy that uses the cluster ID as its resource namespace — so `NewSerializingProducer`/`NewDeserializingConsumer` skip the metadata round-trip entirely unless a configured serde asks for it. When they do, the ID comes from new unexported `Producer.getClusterID`/`Consumer.getClusterID` helpers wrapping librdkafka's `rd_kafka_clusterid`, with a 60s timeout (matching Java's default `max.block.ms`).

## Fixes

Found while writing the tests, all in code added by this PR except the last:

- `SerializingProducer.Produce()` dereferenced the message topic before reaching the producer, so a message without one panicked instead of being rejected. It is now validated the way `Producer.produce()` validates it, before anything is serialized.
- Neither `SerializingProducer.Close()` nor `DeserializingConsumer.Close()` closed the serdes they had built. They do now — the consumer joins their errors with the consumer's, the producer discards them since `Producer.Close()` reports none.
- A delivery report whose opaque is not the produced `SerializableMessage` was silently dropped. `Produce()` is the only path producing on the wrapped producer and always sets that opaque, so this is now a panic rather than a lost delivery report.
- `Build()` silently discarded the error from `schemaregistry.NewClient()` — it was overwritten by the subsequent `NewSerializer`/`NewDeserializer` call.
- `needsClusterID` compared `kafkaClusterID` against `""`, but `newAssociatedNameStrategy` defaults that field to `NamespaceWildcard` (`"-"`) when `subject.name.strategy.kafka.cluster.id` is absent, so the check never fired and the broker-supplied cluster ID was never fetched or applied. Tracked explicitly with a `kafkaClusterIDSet` flag instead of overloading the wildcard as a sentinel.

## Examples

Migrated the avrov3, JSON Schema and Protobuf producer/consumer examples (plain, encryption and migration variants) to the typed serde APIs and fluent builders. There are no avrov2 examples.

## Testing

**Unit tests** (no broker, against the mock SR client) cover the wrappers, the typed messages, the delivery report hook, `deserializeMessage`, the deserialization errors, `NewConfigFromKafkaConfigMap`, the subject name strategy's cluster ID methods, and the fluent builders of all four formats. To make the poll-path message conversion testable without a broker, it was extracted into `deserializeMessage()`, unchanged.

**Integration tests** cover what the unit tests structurally cannot — the success path of `getClusterID()` and a real `loadAssociatedSubjectName()` lookup:

- `schemaregistry/integration/` is a **new sub-module** with its own `go.mod` and its own compose stack (`apache/kafka:4.0.0` + `confluentinc/cp-schema-registry:8.3.1`), following the `kafka/integration/` pattern: a `testconf` struct, the same `-docker.needed` / `-docker.exists` flags, and a single `TestIntegration` entry point. The sub-module keeps the testcontainers/testify surface out of the main module. It adds `TestAssociatedClusterIDJSON` and `TestAssociatedClusterIDIsScopedJSON`, which assert the association actually drove the subject choice — registering under a subject the topic-name fallback could never produce, since `loadAssociatedSubjectName` swallows a 404 and falls back silently.
- `kafka/integration/` gains `TestClusterIDConsistency` and `TestClusterIDPropagation` (broker only, no SR).
- Both are wired into `.semaphore/semaphore_integration_commands.sh`. Note that the two integration stacks bind the same host ports and cannot run at once.

`go build ./...`, `go vet ./schemaregistry/...`, and the unit tests for `kafka` and `schemaregistry` are green; the examples module builds cleanly. The avrov3 / JSON Schema / Protobuf plain examples were verified to round-trip end to end against a live broker + Schema Registry. Encryption/migration rule variants require a data-contract-capable Schema Registry (Confluent Cloud / CP with data contracts); the local `cp-schema-registry` used for testing does not persist ruleSets.

## Dependencies

The branch carries a commit restoring `github.com/google/cel-go` to v0.31.0 and the versions that move with it, because master did not compile at the time: `3a040447` ("Merge release in master", #1604) restored the inline-validation sources from `b5b53fd6` — including `ext.ParseStructField`, which exists only in cel-go v0.31.0 — without the matching `go.mod` bump that the earlier revert `585e624b` had removed along with the code. That has since been fixed on master independently by #1606, so this commit is now redundant and drops out on the next rebase; `go.mod`/`go.sum` here are already byte-identical to master's.


kafka/serializing_producer.go
```diff
@@ -0,0 +1,245 @@
...
+		case nil:
+		case string:
+			if any(serializableMessage.Value).(string) == "" {
+				break
+			}
+			return NewError(ErrInvalidArg, "Value serializer is not set, but Value is not empty", false)
+		default:
+			return NewError(ErrInvalidArg, "Value serializer is not set, but Value is not empty", false)
+		}
+	}
```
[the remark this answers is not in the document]
emasab: this is by design, a serializer is needed in every case the key or value is not nil or empty string

kafka/serializing_producer.go
```diff
@@ -0,0 +1,245 @@
...
+		case nil:
+		case string:
+			if any(serializableMessage.Key).(string) == "" {
+				break
+			}
+			return NewError(ErrInvalidArg, "Key serializer is not set, but Key is not empty", false)
+		default:
+			return NewError(ErrInvalidArg, "Key serializer is not set, but Key is not empty", false)
+		}
+	}
```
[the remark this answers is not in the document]
emasab: this is by design, a serializer is needed in every case the key or value is not nil or empty string

kafka/serializing_producer.go
```diff
@@ -0,0 +1,245 @@
...
+
+	p, err := NewProducer(filteredConf)
+	if err != nil {
+		return nil, err
+	}
+
+	clusterID, err := p.getClusterID(5000)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
emasab: In any case it wouldn't be possible to serialize the message correctly without cluster ID

kafka/deserializing_consumer.go
```diff
@@ -0,0 +1,333 @@
...
+
+	c, err := NewConsumer(filteredConf)
+	if err != nil {
+		return nil, err
+	}
+
+	clusterID, err := c.getClusterID(5000)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
emasab: In any case it wouldn't be possible to serialize the message correctly without cluster ID

kafka/serializing_producer.go
```diff
@@ -0,0 +1,283 @@
...
+//
+// A Serializer that resolves part of its configuration from the Kafka cluster
+// ID reports that by returning true from NeedsClusterID; the cluster ID is then
+// fetched once while the [SerializingProducer] is built and handed over through
+// SetClusterID before the first message is serialized.
+type Serializer interface {
+	Serialize(topic string, msg interface{}) ([]byte, error)
+	SerializeWithHeaders(topic string, msg interface{}) ([]Header, []byte, error)
+	NeedsClusterID() bool
+	SetClusterID(clusterID string)
```
ojasvajain: The only serializer that will use cluster ID is `AssociatedNameStrategy`. For any other serializer, cluster ID will not be used. Would it better we have another interface for clusterID related methods? Here is what Claude suggested me:


```
The idiomatic fix would be interface segregation — pull it out into its own small optional interface and have NewSerializingProducer probe for it via a type assertion, the same pattern Go's stdlib uses for
  optional capabilities (io.ReaderFrom, http.Flusher, http.Hijacker):

  type Serializer interface {
        Serialize(topic string, msg interface{}) ([]byte, error)
        SerializeWithHeaders(topic string, msg interface{}) ([]Header, []byte, error)
        Close() error
  }

  type ClusterIDAware interface {
        NeedsClusterID() bool
        SetClusterID(clusterID string)
  }

  if aware, ok := keySerializer.(ClusterIDAware); ok && aware.NeedsClusterID() {
        ...
  }
```

kafka/serializing_producer.go
```diff
@@ -0,0 +1,283 @@
...
+
+// Serializer turns a typed key or value into the bytes produced to Kafka.
+//
+// A Serializer that resolves part of its configuration from the Kafka cluster
+// ID reports that by returning true from NeedsClusterID; the cluster ID is then
+// fetched once while the [SerializingProducer] is built and handed over through
+// SetClusterID before the first message is serialized.
+type Serializer interface {
+	Serialize(topic string, msg interface{}) ([]byte, error)
+	SerializeWithHeaders(topic string, msg interface{}) ([]Header, []byte, error)
```
ojasvajain: This is not called by `SerializingProducer.Produce()`. It only calls plain `Serialize()`. Do we need it?

kafka/serializing_producer.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// NewSerializingProducer is the same as [NewProducer], returning a
+// [SerializingProducer] wrapping the created [Producer].
+func NewSerializingProducer[K, V any](conf *ConfigMap,
+	keySerializerBuilder SerializerBuilder,
+	valueSerializerBuilder SerializerBuilder) (*SerializingProducer[K, V], error) {
+
+	var keySerializer, valueSerializer Serializer
+	var filteredConf *ConfigMap = conf
```
ojasvajain: nit: *ConfigMap can be omitted.

kafka/serializing_producer.go
```diff
@@ -0,0 +1,283 @@
...
+	var err error
+	if keySerializerBuilder != nil {
+		keySerializer, filteredConf, err = keySerializerBuilder.Build(conf, true)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if valueSerializerBuilder != nil {
+		valueSerializer, filteredConf, err = valueSerializerBuilder.Build(conf, false)
```
ojasvajain: On line 75, we are passing conf. Should it not be filteredConf? Seems like we are ignoring the filtering done by keySerializerBuilder.Build().

kafka/serializing_producer.go
```diff
@@ -0,0 +1,283 @@
...
+
+// SerializerBuilder creates the [Serializer] a [SerializingProducer] uses for
+// its keys or its values.
+//
+// Build is given the producer's [ConfigMap] and whether it is building the key
+// serializer, and returns the serializer together with the ConfigMap to carry
+// on with: any property the serializer consumed itself is filtered out, so that
+// what reaches [NewProducer] holds Kafka properties only.
+type SerializerBuilder interface {
+	Build(conf *ConfigMap, isKey bool) (Serializer, *ConfigMap, error)
```
ojasvajain, line 73: Instead of isKey bool, can we create a named type? Maybe
```
type KeyOrValue int
```
I think it would improve readability.

kafka/serializing_producer.go
```diff
@@ -0,0 +1,283 @@
...
+// String is the same as [Producer.String].
+func (sp *SerializingProducer[K, V]) String() string {
+	return sp.producer.String()
+}
+
+// Produce is the same as [Producer.Produce].
+func (sp *SerializingProducer[K, V]) Produce(serializableMessage *SerializableMessage[K, V], deliveryChan chan Event) error {
+	if serializableMessage == nil || serializableMessage.TopicPartition.Topic == nil ||
+		len(*serializableMessage.TopicPartition.Topic) == 0 {
+		return newErrorFromString(ErrInvalidArg, "")
```
ojasvajain: Should we also add an error string?

kafka/serializing_producer.go
```diff
@@ -0,0 +1,283 @@
...
+ */
+
+package kafka
+
+import (
+	"context"
+	"fmt"
+)
+
+// SerializingProducer wraps a [Producer] and exposes all of its public
```
ojasvajain, line 29: Anytime we add a new method in the plain producer, we will need to define it here. To ensure this is not missed, can we add a small test case using reflection to ensure `SerializingProducer` exposes all public methods of Producer?

kafka/serializing_producer.go
```diff
@@ -0,0 +1,283 @@
...
+				break
+			}
+			return NewError(ErrInvalidArg, "Key serializer is not set, but Key is not empty", false)
+		default:
+			return NewError(ErrInvalidArg, "Key serializer is not set, but Key is not empty", false)
+		}
+	}
+
+	if sp.valueSerializer != nil {
+		serializableMessage.valueBytes, err = sp.valueSerializer.Serialize(*serializableMessage.TopicPartition.Topic, serializableMessage.Value)
```
ojasvajain: I think we missed calling .SerializeWithHeaders

kafka/serializing_producer.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// String is the same as [Producer.String].
+func (sp *SerializingProducer[K, V]) String() string {
+	return sp.producer.String()
+}
+
+// Produce is the same as [Producer.Produce].
+func (sp *SerializingProducer[K, V]) Produce(serializableMessage *SerializableMessage[K, V], deliveryChan chan Event) error {
+	if serializableMessage == nil || serializableMessage.TopicPartition.Topic == nil ||
```
ojasvajain: Can we also check if serializing producer is already closed and fail fast? This only gets checked when we call `sp.producer.Produce` which gets called much later after key and value serializers have been called.

kafka/deserializing_consumer.go
```diff
@@ -0,0 +1,395 @@
...
+// Deserializer turns the bytes consumed from Kafka into a typed key or value.
+//
+// A Deserializer that resolves part of its configuration from the Kafka cluster
+// ID reports that by returning true from NeedsClusterID; the cluster ID is then
+// fetched once while the [DeserializingConsumer] is built and handed over
+// through SetClusterID before the first message is deserialized.
+type Deserializer interface {
+	DeserializeWithHeaders(topic string, headers []Header, payload []byte) (interface{}, error)
+	NeedsClusterID() bool
+	SetClusterID(clusterID string)
```
ojasvajain: Same suggestion for having them in a separate interface.

kafka/serializing_producer.go
ojasvajain, line 40: I don't see it getting used.

kafka/deserializing_consumer.go
```diff
@@ -0,0 +1,395 @@
...
+ */
+
+package kafka
+
+import (
+	"errors"
+	"fmt"
+)
+
+// DeserializingConsumer wraps a [Consumer] and exposes all of its public
```
ojasvajain, line 26: https://github.com/confluentinc/confluent-kafka-go/pull/1583/changes#r4049513612

kafka/deserializing_consumer.go
```diff
@@ -0,0 +1,395 @@
...
+
+// DeserializerBuilder creates the [Deserializer] a [DeserializingConsumer] uses
+// for its keys or its values.
+//
+// Build is given the consumer's [ConfigMap] and whether it is building the key
+// deserializer, and returns the deserializer together with the ConfigMap to
+// carry on with: any property the deserializer consumed itself is filtered out,
+// so that what reaches [NewConsumer] holds Kafka properties only.
+type DeserializerBuilder interface {
+	Build(conf *ConfigMap, isKey bool) (Deserializer, *ConfigMap, error)
```
ojasvajain, line 144: Same suggestion as SerializerBuilder. Can we use a typedef instead of bool?

kafka/deserializing_consumer.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+// NewDeserializingConsumer is the same as [NewConsumer], returning a
+// [DeserializingConsumer] wrapping the created [Consumer].
+func NewDeserializingConsumer[K, V any](conf *ConfigMap,
+	keyDeserializerBuilder DeserializerBuilder,
+	valueDeserializerBuilder DeserializerBuilder) (*DeserializingConsumer[K, V], error) {
+
+	var keyDeserializer, valueDeserializer Deserializer
+	var filteredConf *ConfigMap = conf
```
ojasvajain: `*ConfigMap` can be omitted.

kafka/deserializing_consumer.go
```diff
@@ -0,0 +1,395 @@
...
+	var err error
+	if keyDeserializerBuilder != nil {
+		keyDeserializer, filteredConf, err = keyDeserializerBuilder.Build(conf, true)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if valueDeserializerBuilder != nil {
+		valueDeserializer, filteredConf, err = valueDeserializerBuilder.Build(conf, false)
```
ojasvajain: We should pass filteredConf instead of conf

kafka/deserializing_consumer.go
```diff
@@ -0,0 +1,395 @@
...
+
+// deserializeMessage deserializes the key and the value of msg and returns the
+// resulting [*DeserializedMessage], or a deserialization error event.
+func (dc *DeserializingConsumer[K, V]) deserializeMessage(msg *Message) Event {
+	var deserializedKey K
+	var deserializedValue V
+	var ok bool
+
+	if msg.TopicPartition.Topic == nil {
+		return msg
```
ojasvajain: We have documented that 
```
// A [*Message] is returned as a [*DeserializedMessage] with its key and value
// deserialized. If deserialization fails, a [KeyDeserializationError] or a
// [ValueDeserializationError] is returned instead. Any other event is returned
// as-is.
``` 
but here we are returning a Message type. Should we return an error instead? Something like:

```
 if msg.TopicPartition.Topic == nil {
        if msg.TopicPartition.Error != nil {
                return newErrorFromString(...)  // or however Error is normally constructed from msg.TopicPartition.Error
        }
        return msg
  }
```

kafka/deserializing_consumer.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	if msg.Key != nil && dc.keyDeserializer != nil {
```
ojasvajain, line 353: There can be a case where msg.Key is set but dc.keyDeserializer is nil. In that case, we are not doing any deserialization at all and the key value in deserialized message will be empty. Is this behaviour expected?

kafka/deserializing_consumer.go
```diff
@@ -0,0 +1,395 @@
...
+			return NewKeyDeserializationError(msg.TopicPartition, err)
+		}
+
+		deserializedKey, ok = deserializedKeyInterface.(K)
+		if !ok {
+			return NewKeyDeserializationError(msg.TopicPartition,
+				fmt.Errorf("Wrong deserialized key type: %T", deserializedKeyInterface))
+		}
+	}
+	if msg.Value != nil && dc.valueDeserializer != nil {
```
ojasvajain, line 368: Same as above

kafka/deserialized_message.go
```diff
@@ -0,0 +1,81 @@
...
+		TopicPartition: msg.TopicPartition,
+		Key:            key,
+		Value:          value,
+		keyBytes:       msg.Key,
+		valueBytes:     msg.Value,
+		Timestamp:      msg.Timestamp,
+		TimestampType:  msg.TimestampType,
+		Opaque:         msg.Opaque,
+		Headers:        msg.Headers,
+		message:        msg,
```
ojasvajain: I am not sure if we need to store message within the `DeserializedMessage` object. One issue that I see is: now there are two copies of TopicParition info, one inside `DeserializedMessage` and one in DeserializedMessage.message. If I change the first one and then do a commit, commit will still the read the topic parititon from DeserializedMessage.message.


I would suggest that for commit message and commit offsets functions (which are the only two places where we use DeserializedMessage.message, we create a temporary Message object. 

```
 func (dc *DeserializingConsumer[K, V]) CommitMessage(m *DeserializedMessage[K, V]) ([]TopicPartition, error) {
        return dc.consumer.CommitMessage(&Message{TopicPartition: m.TopicPartition})
  }

  func (dc *DeserializingConsumer[K, V]) StoreMessage(m *DeserializedMessage[K, V]) (storedOffsets []TopicPartition, err error) {
        return dc.consumer.StoreMessage(&Message{TopicPartition: m.TopicPartition})
  }
```

kafka/handle.go
```diff
@@ -130,6 +132,9 @@ type handle struct {
 
 	// WaitGroup to wait for spawned go-routines to finish.
 	waitGroup sync.WaitGroup
+
+	// sendMessageToChannel is a function that sends a message to a delivery channel. It can be overridden for additional transformation or processing of the message before sending it to the channel. If not set, the default behavior is to send the message directly to the channel.
+	sendMessageToChannel sendMessageToChannelFunc
```
ojasvajain: I can see that only producer can set a `sendMessageToChannel` func. Can we move it to producer struct if it is specific only to producer?

kafka/producer.go
```diff
@@ -172,6 +172,16 @@ func (p *Producer) gethandle() *handle {
 	return &p.handle
 }
 
+func (p *Producer) getClusterID(timeout int) (string, error) {
```
ojasvajain:
```suggestion
func (p *Producer) getClusterID(timeoutMs int) (string, error) {
```

ojasvajain: We are defining it for both producer and consumer and the definition is exactly same. Can we define it on the handle instead?

schemaregistry/serde/avrov2/avro.go
```diff
@@ -92,6 +92,104 @@ func NewSerializer(client schemaregistry.Client, serdeType serde.Type, conf *Ser
...
+		serdeType = serde.KeySerde
+	} else {
+		serdeType = serde.ValueSerde
+	}
+	if serializerConf == nil {
+		serializerConf = NewSerializerConfig()
+	}
+	s, err := NewSerializer(client, serdeType, serializerConf)
+	if err != nil {
+		return nil, nil, err
```
ojasvajain, line 170: we should also close the SR client before returning, maybe using defer would help.

kafka/serializing_producer.go
emasab: This approach is taken on .NET client, but there the kafka Serializer is already published, so it's an extension method on the interface. Here it's a new interface so we can have it in the same interface definition.
This was changed to `SetClusterIDResolver` for the problems with the OAuth custom callback on client creation.

kafka/serializing_producer.go
emasab: It should be conf as it contains the additional SR properties, that we're now using at the moment. The SR properties must be filtered in both cases so it's ok to overwrite the value.

kafka/serializing_producer.go
emasab: Here it matches the Java build method for a serializer:

https://kafka.apache.org/42/javadoc/org/apache/kafka/common/serialization/Serializer.html#configure(java.util.Map,boolean)

kafka/serializing_producer.go
emasab: Now it's used, I think I can remove the Serialize as it's the legacy one and now it calls SerializeWithHeaders

kafka/deserializing_consumer.go
emasab: [See ](https://github.com/confluentinc/confluent-kafka-go/pull/1583#discussion_r4060917892)

kafka/deserializing_consumer.go
emasab: [See](https://github.com/confluentinc/confluent-kafka-go/pull/1583#discussion_r4060965216)

kafka/deserializing_consumer.go
emasab: [See](https://github.com/confluentinc/confluent-kafka-go/pull/1583#discussion_r4060947649)

kafka/deserializing_consumer.go
emasab: yes, this shouldn't happen but now I'm returning it as KeyDeserializationError or ValueDeserializationError, whatever triggers first

kafka/deserializing_consumer.go
emasab: Yes, because in case the consumer doesn't set a key deserializer, and the producer starts producing with keys, this way it doesn't break the consumer and you can upgrade them later.

schemaregistry/serde/avrov2/avro.go
emasab: I added in the new commit ownership about the client in the serde. In case it's owned it'll be closed when closing the kafka client.

kafka/handle.go
emasab: no, because it's used in `event.go` `eventPoll` that is generic and uses the handle only.

kafka/deserializing_consumer.go
emasab: Same reasoning about the key deserializer just here it's a rarer case: you usually set null values as a tombstone marker for compacted topics. Even in that case it's difficult you don't have a `valueDeserializer` because you're not only processing tombstone markers. In some other cases you could want to receive only keys without values, for example for a topic holding a stream of id updates.

kafka/producer.go
emasab: Yes, moved it to the handle and made the GetClusterID public as in other languages

kafka/serializing_producer.go
emasab: done

kafka/serializing_producer.go
ojasvajain: But we are assuming that both Build() calls filter conf the same way. But that's not guaranteed. For example, `SetSchemaRegistryClient` skips filtering entirely and returns conf unfiltered. So if key builder has no explict client and value builder has one via (SetSchemaRegistryClient), the unfiltered conf from `valueSerializerBuilder.Build` will overwrite the filtered conf from `keySerializerBuilder.Build`

kafka/producer.go
ojasvajain: Nice, thanks. This was also a community request.

kafka/handle.go
ojasvajain: eventPoll already reaches into type-specific state this way via h.c.handleRebalanceEvent for RD_KAFKA_EVENT_REBALANCE. `h.p.sendMessageToChannel` for RD_KAFKA_EVENT_DR can follow the same existing pattern.

kafka/serializing_producer.go
emasab: Even in future, if will we implemented this way, the user must not set the SR configuration in conf and then set the SR client.
It's possible that the two implementations will be removing different values, I'll scan the second filtered configuration and remove the values that aren't in the first one.

kafka/handle.go
emasab: ok, moved it to the producer structure

ojasvajain (approved):

Issue #1569, opened 2026-06-30
Calling Consumer.Close is not threadsafe.


oliverbestmann-bauer:
rdkafka is supposed to be threadsafe. I was under the impression, that the go wrapper is threadsafe too.

```go
package testx

import (
	"sync"
	"testing"
	"time"

	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
	"github.com/stretchr/testify/require"
)

func TestConcurrency(t *testing.T) {
	// create a new mock cluster
	cluster, err := kafka.NewMockCluster(1)
	require.NoError(t, err, "create mock cluster")

	_ = cluster.CreateTopic("test", 16, 1)

	for idx := range 1024 {
		t.Logf("Iteration %d", idx)
		// no crash
		// var lock sync.Locker = new(sync.Mutex)

		var lock sync.Locker = new(noLock)

		consumer, err := kafka.NewConsumer(&kafka.ConfigMap{
			"group.id":          "test",
			"bootstrap.servers": cluster.BootstrapServers(),
			"auto.offset.reset": "earliest",
		})
		require.NoError(t, err)

		go func() {
			lock.Lock()
			err := consumer.Subscribe("test", nil)
			lock.Unlock()

			require.NoError(t, err)

			for !consumer.IsClosed() {
				lock.Lock()
				_, _ = consumer.ReadMessage(50 * time.Millisecond)
				lock.Unlock()
			}
		}()

		time.Sleep(50 * time.Millisecond)

		lock.Lock()
		_ = consumer.Close()
		lock.Unlock()
	}
}

type noLock struct{}

func (n noLock) Lock() {
}

func (n noLock) Unlock() {
}
```

After only a few iterations, I get a panic:
```
malloc(): unaligned tcache chunk detected
SIGABRT: abort
PC=0x767c0f49fdcc m=3 sigcode=18446744073709551610
signal arrived during cgo execution

goroutine 9 gp=0x1480cbfd81e0 m=3 mp=0x1480cbe51008 [syscall]:
runtime.cgocall(0xd7d440, 0x1480cbe5fac8)
	/home/oliver/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.2.linux-amd64/src/runtime/cgocall.go:167 +0x4b fp=0x1480cbe5faa0 sp=0x1480cbe5fa68 pc=0x49104b
[...]

PR #1585, opened 2026-07-27
Fix use-after-free race between Close and in-flight C calls

oliverbestmann-bauer:
This fixes issue #1569.

Produced mostly by AI for fixing an issue that happends at $WORK.

## Why

`Consumer`, `Producer`, and `AdminClient` each wrap a librdkafka handle (`rd_kafka_t`) and, for consumers, a queue (`rd_kafka_queue_t`). Every public method follows the same pattern:

```go
err := c.verifyClient()
if err != nil {
    return err
}
// ... use c.handle.rk / c.handle.rkq in a C call ...
```

`verifyClient()` just checks an atomic `isClosed` flag. That's a classic check-then-use (TOCTOU) race: a goroutine can pass the closed check while `isClosed == 0`, then enter a **blocking** C call (e.g. `rd_kafka_queue_poll` inside `ReadMessage`/`Poll`, which can block for the full poll timeout). Meanwhile, `Close()` sets `isClosed = 1` and immediately calls `rd_kafka_queue_destroy()` / `rd_kafka_destroy()`, freeing the very memory the other goroutine is still operating on inside its blocking cgo call.

The result is a heap use-after-free. In practice this doesn't crash where the corruption happens — it typically shows up later, e.g. as a `SIGSEGV` inside an unrelated `rd_kafka_new()` call in a subsequent test iteration, which made it look scarier and harder to diagnose than it actually was.

This is not a librdkafka thread-safety issue. librdkafka *is* thread-safe for concurrent operations on a live handle, but it (like any C library with explicit `destroy`/`free` calls) has no way to make `rd_kafka_destroy()` safe against a concurrent in-flight caller — freeing memory someone is actively dereferencing is inherently unsafe, in any language. Establishing quiescence (no in-flight calls) before destroying is the caller's responsibility, and the Go wrapper wasn't doing that.

The included regression test (`kafka/race_test.go`, `TestConcurrency`) reproduces this reliably: it spins up many short-lived consumers, each with a background goroutine calling `ReadMessage` in a loop, and closes the consumer out from under it. Before this fix, this reliably `SIGSEGV`s within a few hundred iterations against a mock cluster.

## What changed

- `handle` (shared by `Consumer`, `Producer`, `AdminClient`) gained a `pollLock sync.RWMutex`, plus `rlock()`/`runlock()` helpers. `rlock()` takes the read lock and returns `ErrState` if the handle has already been closed (`rk == nil`), checked atomically under the lock.
- Every method that touches the C handle/queue now holds `pollLock.RLock()` for the duration of its C call — across `Consumer` (`Subscribe*`, `Assign*`, `Commit*`, `Seek*`, `Position`, `Pause`/`Resume`, `GetMetadata`, transactions-adjacent getters, etc.), `Producer` (`produce`/`produceBatch`, `Flush`'s background flush goroutine, `Purge`, metadata/watermark queries, all transactional APIs, SASL/OAuth setters), and `AdminClient` (all ~26 methods that previously used `verifyClient`, e.g. `CreateTopics`, `DeleteTopics`, `DescribeConfigs`, ACL and consumer-group APIs, etc.).
- `eventPoll` (the shared implementation behind `Poll`/`ReadMessage`) takes the read lock for its entire blocking C poll, and bails out cleanly if the queue has already been destroyed.
- `Close()` on `Consumer`, `Producer`, and `AdminClient` now takes `pollLock.Lock()` **before** destroying the queue/handle, and nils out `rk`/`rkq` under that same lock.

Note: to avoid nested read-locks (which can deadlock a `sync.RWMutex` if a writer is waiting in between), only top-level public entry points take the lock. Methods that merely delegate to another already-locked method (e.g. `Subscribe` → `SubscribeTopics`, `Commit`/`CommitMessage`/`CommitOffsets` → the private `commit`, `StoreMessage` → `StoreOffsets`) do not lock themselves.

## API behavior changes

- **No signature or semantic changes** - same methods, same return types, same errors on a closed client (`ErrState` via `getOperationNotAllowedErrorForClosedClient()`).
- **`Close()` now blocks until all in-flight calls on that client have returned**, instead of racing them. Concretely:
  - If another goroutine is inside `Poll`/`ReadMessage` when `Close()` is called, `Close()` waits for that call to return (up to its configured timeout) before proceeding with teardown.
  - Any blocking call with a long or infinite timeout (e.g. `Consumer.commit`'s internal indefinite wait, or `AdminClient` calls with a long `ctx` deadline) will similarly make a concurrent `Close()` wait for that call to finish.
  - Once `Close()` has started, any new call into the client blocks briefly on the write lock and then returns the existing "client is closed" error — no C code path can execute concurrently with teardown anymore.
- This trades a previously-undefined-behavior (crash) for well-defined, bounded blocking. Applications that already avoid overlapping `Close()` with in-flight calls (the documented/intended usage) will see no observable difference. Applications relying on the previous race window (unlikely, since it was UB) may see `Close()` take slightly longer to return when calls are in flight.

### DIFF
--- a/kafka/adminapi.go
+++ b/kafka/adminapi.go
@@ -157,9 +157,9 @@ import "C"
 
 // AdminClient is derived from an existing Producer or Consumer
 type AdminClient struct {
-	handle    *handle
-	isDerived bool   // Derived from existing client handle
-	isClosed  uint32 // to check if Admin Client is closed or not.
+	handle        *handle
+	isDerived     bool      // Derived from existing client handle
+	isClosed      uint32    // to check if Admin Client is closed or not.
 	adminTermChan chan bool // For log channel termination
 }
 
@@ -1686,10 +1686,11 @@ func cToDeletedRecordResult(
 //
 // Requires broker version >= 0.10.0.
 func (a *AdminClient) ClusterID(ctx context.Context) (clusterID string, err error) {
-	err = a.verifyClient()
+	err = a.handle.rlock()
 	if err != nil {
 		return "", err
 	}
+	defer a.handle.runlock()
 
 	responseChan := make(chan *C.char, 1)
 
@@ -1723,10 +1724,11 @@ func (a *AdminClient) ClusterID(ctx context.Context) (clusterID string, err erro
 //
 // Requires broker version >= 0.10.0.
 func (a *AdminClient) ControllerID(ctx context.Context) (controllerID int32, err error) {
-	err = a.verifyClient()
+	err = a.handle.rlock()
 	if err != nil {
 		return -1, err
 	}
+	defer a.handle.runlock()
 
 	responseChan := make(chan int32, 1)
 
@@ -1757,10 +1759,11 @@ func (a *AdminClient) ControllerID(ctx context.Context) (controllerID int32, err
 //
 // Note: TopicSpecification is analogous to NewTopic in the Java Topic Admin API.
 func (a *AdminClient) CreateTopics(ctx context.Context, topics []TopicSpecification, options ...CreateTopicsAdminOption) (result []TopicResult, err error) {
-	err = a.verifyClient()
+	err = a.handle.rlock()
 	if err != nil {
 		return nil, err
 	}
+	defer a.handle.runlock()
 
 	cTopics := make([]*C.rd_kafka_NewTopic_t, len(topics))
 
@@ -1876,10 +1879,11 @@ func (a *AdminClient) CreateTopics(ctx context.Context, topics []TopicSpecificat
 //
 // Requires broker version >= 0.10.1.0
 func (a *AdminClient) DeleteTopics(ctx context.Context, topics []string, options ...DeleteTopicsAdminOption) (result []TopicResult, err error) {
-	err = a.verifyClient()
+	err = a.handle.rlock()
 	if err != nil {
 		return nil, err
 	}
+	defer a.handle.runlock()
 
 	cTopics := make([]*C.rd_kafka_DeleteTopic_t, len(topics))
 
@@ -1939,10 +1943,11 @@ func (a *AdminClient) DeleteTopics(ctx context.Context, topics []string, options
 
 // CreatePartitions creates additional partitions for topics.
 func (a *AdminClient) CreatePartitions(ctx context.Context, partitions []PartitionsSpecification, options ...CreatePartitionsAdminOption) (result []TopicResult, err error) {
-	err = a.verifyClient()
+	err = a.handle.rlock()
 	if err != nil {
 		return nil, err
 	}
+	defer a.handle.runlock()
 
 	cParts := make([]*C.rd_kafka_NewPartitions_t, len(partitions))
 
@@ -2035,10 +2040,11 @@ func (a *AdminClient) CreatePartitions(ctx context.Context, partitions []Partiti
 // resource requests must be sent to the broker specified in the resource.
 // Deprecated: AlterConfigs is deprecated in favour of IncrementalAlterConfigs
 func (a *AdminClient) AlterConfigs(ctx context.Context, resources []ConfigResource, options ...AlterConfigsAdminOption) (result []ConfigResourceResult, err error) {
-	err = a.verifyClient()
+	err = a.handle.rlock()
 	if err != nil {
 		return nil, err
 	}
+	defer a.handle.runlock()
 
 	cRes := make([]*C.rd_kafka_ConfigResource_t, len(resources))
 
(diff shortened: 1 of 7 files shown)


kafka/event.go
```diff
@@ -160,6 +160,13 @@ func (h *handle) eventPoll(channel chan Event, timeoutMs int, maxEvents int, ter
 	if channel == nil {
 		maxEvents = 1
 	}
+
+	h.pollLock.RLock()
```
Ankith-Confluent, line 164: The read lock is held across the whole body, including handleRebalanceEvent at line 189, so it's held across application callback code. Those callbacks call `Assign()/Commit()`, which take `rlock()` again. RWMutex isn't reentrant, so a concurrent Close() deadlocks both goroutines. Its better to not hold the lock across the rebalance dispatch.

kafka/consumer.go
```diff
@@ -222,6 +228,10 @@ func (c *Consumer) IncrementalUnassign(partitions []TopicPartition) (err error)
 // is returned.
 // Should typically only be called during rebalancing.
 func (c *Consumer) GetRebalanceProtocol() string {
+	if err := c.handle.rlock(); err != nil {
```
Ankith-Confluent, line 237: This is also called from inside eventPoll's read lock, `handleRebalanceEvent → GetRebalanceProtocol()` at consumer.go:1095. That's a nested RLock, which deadlocks if Close() is waiting. Needs an unlocked internal variant. Note it fires for the default consumer with no rebalance callback.

kafka/consumer.go
```diff
@@ -553,6 +575,11 @@ func (c *Consumer) Close() (err error) {
 		c.Poll(100)
 	}
 
+	// Wait for any in-flight eventPoll to finish before tearing down the
+	// C handle and queue, then prevent new polls from starting.
+	c.handle.pollLock.Lock()
```
Ankith-Confluent, line 615: close(c.events) happens at line 569, before this lock, but Producer.Close() was deliberately reordered to close under it. handleRebalanceEvent sends to c.events at 1125/1127. When eventsChanEnable=false that channel is nil, so the send blocks forever holding the read lock and Close() deadlocks here.

kafka/handle.go
```diff
@@ -136,6 +140,31 @@ func (h *handle) String() string {
...
+// goroutine. It returns an error if the handle has already been closed.
+//
+// On success the caller becomes responsible for calling h.pollLock.RUnlock()
+// (typically via defer) once it is done using the C handle. On error the lock
+// is not held.
+//
+// Note: this must only be taken by top-level entry points. Helpers reachable
+// from an already-locked method must NOT call rlock() again, since sync.RWMutex
+// read locks are not reentrant and may deadlock if a writer (Close) is waiting.
+func (h *handle) rlock() error {
```
Ankith-Confluent, line 157: Close() waits on readers with no bound. ReadMessage(-1), commit() (consumer.go:282), Seek (consumer.go:424), and ClusterID with a deadline-less ctx can hold this forever. Consider rd_kafka_queue_yield(h.rkq) before taking the write lock.

kafka/producer.go
```diff
@@ -374,7 +389,11 @@ func (p *Producer) Flush(timeoutMs int) int {
 	// might do in case it is flushing.
 	go func() {
 		for flushInterval := range flushIntervalChan {
+			if err := p.handle.rlock(); err != nil {
```
Ankith-Confluent, line 392: When rlock() fails nothing paces the outer Flush loop (404–419), eventPoll also returns instantly, and Len() still counts buffered events because it reads them before the rlock() check (line 370). Flush(30000) after close spins hot for 30s.

kafka/producer.go
```diff
@@ -1031,9 +1071,10 @@ func (p *Producer) AbortTransaction(ctx context.Context) error {
 // existing broker connections that were established with the old credentials.
 // This method applies only to the SASL PLAIN and SCRAM mechanisms.
 func (p *Producer) SetSaslCredentials(username, password string) error {
```
Ankith-Confluent, line 1087: This one was converted, but Consumer.SetSaslCredentials (consumer.go:1139) was missed, its still verifyClient() + c.handle.rk, i.e. the original race. Credential rotation usually runs on its own goroutine, which is exactly the target case.

kafka/adminapi.go
```diff
@@ -157,9 +157,9 @@ import "C"
 
 // AdminClient is derived from an existing Producer or Consumer
 type AdminClient struct {
-	handle    *handle
-	isDerived bool   // Derived from existing client handle
-	isClosed  uint32 // to check if Admin Client is closed or not.
+	handle        *handle
```
Ankith-Confluent, line 160: verifyClient() at line 171 now has zero call sites in this file. Worth removing or noting.

Ankith-Confluent (changes requested):
Hey @oliverbestmann-bauer 
Thanks for the PR!
I've reproduced the bug locally and can confirm it.
These are my first round of review comments.

oliverbestmann-bauer:
Please have a another look, it should work now. I've also added tests for the deadlocks that now do not trigger anymore

martin-joehren-bm:
Hi, can we get a review here please?

PR #1588, opened 2026-07-30
fix: handle rd_kafka_consumer_close_queue() error to prevent Close() hang on fatally-errored consumers

nolansherman-spoton:
## Summary

Fix Close() hanging indefinitely when the consumer has a fatal error (e.g., `FENCED_INSTANCE_ID`).

### Root Cause

`rd_kafka_consumer_close_queue()` returns an `rd_kafka_error_t*` when it cannot start the close protocol (e.g., when a fatal error is set). The Go wrapper was **discarding this return value**, then entering a polling loop waiting for `rd_kafka_consumer_closed()` to return 1 — which never happens because the close was never started.

```go
// Before (buggy): return value discarded
C.rd_kafka_consumer_close_queue(c.handle.rk, c.handle.rkq)

for C.rd_kafka_consumer_closed(c.handle.rk) != 1 {  // never becomes 1
    c.Poll(100)
}
```

### Fix

Check the return value of `rd_kafka_consumer_close_queue()`. If it returns an error, the close was never started — there is nothing to wait for. Force-destroy the handle with `RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE` and return the error to the caller.

```go
// After (fixed): check return value
if cErr := C.rd_kafka_consumer_close_queue(c.handle.rk, c.handle.rkq); cErr != nil {
    closeErr := newErrorFromCErrorDestroy(cErr)
    atomic.StoreUint32(&c.isClosed, 1)
    C.rd_kafka_queue_destroy(c.handle.rkq)
    c.handle.rkq = nil
    c.handle.cleanup()
    C.rd_kafka_destroy_flags(c.handle.rk, C.RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE)
    return closeErr
}
```

### Why This Works By Default

Unlike the previous opt-in config approach (`go.consumer.close.skip.on.fatal`), this fix works for **all users** without any configuration. Any error from `rd_kafka_consumer_close_queue()` — not just fatal errors — is handled correctly.

### Behavioral Changes

- `Close()` now returns an error (instead of `nil`) when the consumer has a fatal error. This gives callers visibility that offsets were not committed.
- No new configuration options are needed.
- The `GetFatalError()` and `TestFatalError()` methods are retained on `Consumer` (mirrors the Producer API).

### Why This Is Safe

A fatally-errored consumer is already dead from the broker perspective:
- Its partitions are reassigned to the new consumer instance
- Its member ID is removed from the group
- Any commits will be rejected with `FENCED_INSTANCE_ID`
- LeaveGroup is a no-op (member already removed)

Credit to @Ankith-Confluent for identifying the simpler root cause and suggesting this approach.

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -51,6 +51,18 @@ type Consumer struct {
 
 	isClosed  uint32
 	isClosing uint32
+
+	// closeSkipOnFatal, when true, causes Close() to skip the consumer group
+	// close protocol (LeaveGroup, offset commits) if the consumer has a fatal
+	// error set (e.g., FENCED_INSTANCE_ID). Instead, it force-destroys the
+	// handle immediately using rd_kafka_destroy_flags(NO_CONSUMER_CLOSE).
+	//
+	// This prevents Close() from hanging indefinitely when the consumer is
+	// already dead and the broker is slow to respond to commits that will be
+	// rejected anyway.
+	//
+	// Configured via "go.consumer.close.skip.on.fatal" (bool, default false).
+	closeSkipOnFatal bool
 }
 
 // IsClosed returns boolean representing if client is closed or not
@@ -528,6 +540,11 @@ func (c *Consumer) ReadMessage(timeout time.Duration) (*Message, error) {
 
 // Close Consumer instance.
 // The object is no longer usable after this call.
+//
+// If go.consumer.close.skip.on.fatal is enabled and the consumer has a fatal
+// error (e.g., FENCED_INSTANCE_ID), Close() will skip the consumer group close
+// protocol entirely and force-destroy the handle. This prevents hanging on
+// offset commits that will be rejected by the broker anyway.
 func (c *Consumer) Close() (err error) {
 	// Check if the client is already closed.
 	err = c.verifyClient()
@@ -545,6 +562,24 @@ func (c *Consumer) Close() (err error) {
 	c.handle.waitGroup.Wait()
 	if c.eventsChanEnable {
 		close(c.events)
+	}
+
+	// If the consumer has a fatal error and skip-on-fatal is enabled,
+	// bypass the close protocol entirely. The consumer is already dead
+	// from the broker's perspective — attempting commits or LeaveGroup
+	// will either be rejected or hang if the broker is unresponsive.
+	if c.closeSkipOnFatal {
+		cErrstr := (*C.char)(C.malloc(C.size_t(512)))
+		fatalCode := C.rd_kafka_fatal_error(c.handle.rk, cErrstr, 512)
+		C.free(unsafe.Pointer(cErrstr))
+		if int(fatalCode) != 0 {
+			atomic.StoreUint32(&c.isClosed, 1)
+			C.rd_kafka_queue_destroy(c.handle.rkq)
+			c.handle.rkq = nil
+			c.handle.cleanup()
+			C.rd_kafka_destroy_flags(c.handle.rk, C.RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE)
+			return nil
+		}
 	}
 
 	C.rd_kafka_consumer_close_queue(c.handle.rk, c.handle.rkq)
@@ -581,6 +616,11 @@ func (c *Consumer) Close() (err error) {
 //	go.events.channel.size (int, 1000) - Events() channel size
 //	go.logs.channel.enable (bool, false) - Forward log to Logs() channel.
 //	go.logs.channel (chan kafka.LogEvent, nil) - Forward logs to application-provided channel instead of Logs(). Requires go.logs.channel.enable=true.
+//	go.consumer.close.skip.on.fatal (bool, false) - If true, Close() will skip the consumer group close protocol
+//	                                     (LeaveGroup, offset commits) when the consumer has a fatal error
+//	                                     (e.g., fenced by another instance). Instead it force-destroys the
+//	                                     handle immediately, preventing Close() from hanging on commits
+//	                                     that will be rejected by the broker.
 //
 // WARNING: Due to the buffering nature of channels (and queues in general) the
 // use of the events channel risks receiving outdated events and
@@ -627,6 +667,12 @@ func NewConsumer(conf *ConfigMap) (*Consumer, error) {
 		return nil, err
 	}
 	eventsChanSize := v.(int)
+
+	v, err = confCopy.extract("go.consumer.close.skip.on.fatal", false)
+	if err != nil {
+		return nil, err
+	}
+	c.closeSkipOnFatal = v.(bool)
 
 	logsChanEnable, logsChan, err := confCopy.extractLogConfig()
 	if err != nil {
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -70,6 +70,18 @@ func (c *Consumer) IsClosed() bool {
 	return atomic.LoadUint32(&c.isClosed) == 1
 }
 
+// GetFatalError returns an Error object if the client instance has raised a
+// fatal error, else nil.
+func (c *Consumer) GetFatalError() error {
+	return getFatalError(c)
+}
+
+// TestFatalError triggers a fatal error in the underlying client.
+// This is to be used strictly for testing purposes.
+func (c *Consumer) TestFatalError(code ErrorCode, str string) ErrorCode {
+	return testFatalError(c, code, str)
+}
+
 func (c *Consumer) verifyClient() error {
 	if c.IsClosed() {
 		return getOperationNotAllowedErrorForClosedClient()
@@ -568,18 +580,13 @@ func (c *Consumer) Close() (err error) {
 	// bypass the close protocol entirely. The consumer is already dead
 	// from the broker's perspective — attempting commits or LeaveGroup
 	// will either be rejected or hang if the broker is unresponsive.
-	if c.closeSkipOnFatal {
-		cErrstr := (*C.char)(C.malloc(C.size_t(512)))
-		fatalCode := C.rd_kafka_fatal_error(c.handle.rk, cErrstr, 512)
-		C.free(unsafe.Pointer(cErrstr))
-		if int(fatalCode) != 0 {
-			atomic.StoreUint32(&c.isClosed, 1)
-			C.rd_kafka_queue_destroy(c.handle.rkq)
-			c.handle.rkq = nil
-			c.handle.cleanup()
-			C.rd_kafka_destroy_flags(c.handle.rk, C.RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE)
-			return nil
-		}
+	if c.closeSkipOnFatal && getFatalError(c) != nil {
+		atomic.StoreUint32(&c.isClosed, 1)
+		C.rd_kafka_queue_destroy(c.handle.rkq)
+		c.handle.rkq = nil
+		c.handle.cleanup()
+		C.rd_kafka_destroy_flags(c.handle.rk, C.RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE)
+		return nil
 	}
 
 	C.rd_kafka_consumer_close_queue(c.handle.rk, c.handle.rkq)
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -51,35 +51,11 @@ type Consumer struct {
 
 	isClosed  uint32
 	isClosing uint32
-
-	// closeSkipOnFatal, when true, causes Close() to skip the consumer group
-	// close protocol (LeaveGroup, offset commits) if the consumer has a fatal
-	// error set (e.g., FENCED_INSTANCE_ID). Instead, it force-destroys the
-	// handle immediately using rd_kafka_destroy_flags(NO_CONSUMER_CLOSE).
-	//
-	// This prevents Close() from hanging indefinitely when the consumer is
-	// already dead and the broker is slow to respond to commits that will be
-	// rejected anyway.
-	//
-	// Configured via "go.consumer.close.skip.on.fatal" (bool, default false).
-	closeSkipOnFatal bool
 }
 
 // IsClosed returns boolean representing if client is closed or not
 func (c *Consumer) IsClosed() bool {
 	return atomic.LoadUint32(&c.isClosed) == 1
-}
-
-// GetFatalError returns an Error object if the client instance has raised a
-// fatal error, else nil.
-func (c *Consumer) GetFatalError() error {
-	return getFatalError(c)
-}
-
-// TestFatalError triggers a fatal error in the underlying client.
-// This is to be used strictly for testing purposes.
-func (c *Consumer) TestFatalError(code ErrorCode, str string) ErrorCode {
-	return testFatalError(c, code, str)
 }
 
 func (c *Consumer) verifyClient() error {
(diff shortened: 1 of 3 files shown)


kafka/consumer.go
```diff
@@ -628,6 +668,12 @@ func NewConsumer(conf *ConfigMap) (*Consumer, error) {
 	}
 	eventsChanSize := v.(int)
 
+	v, err = confCopy.extract("go.consumer.close.skip.on.fatal", false)
+	if err != nil {
+		return nil, err
+	}
+	c.closeSkipOnFatal = v.(bool)
```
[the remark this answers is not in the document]
nolansherman-spoton: tests added,

kafka/consumer.go
```diff
@@ -547,6 +564,24 @@ func (c *Consumer) Close() (err error) {
...
+		C.free(unsafe.Pointer(cErrstr))
+		if int(fatalCode) != 0 {
+			atomic.StoreUint32(&c.isClosed, 1)
+			C.rd_kafka_queue_destroy(c.handle.rkq)
+			c.handle.rkq = nil
+			c.handle.cleanup()
+			C.rd_kafka_destroy_flags(c.handle.rk, C.RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE)
+			return nil
+		}
+	}
```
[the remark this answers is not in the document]
nolansherman-spoton: updated to use getFatalError

c1moore (approved):

JacekPietrasSpotOn (approved):

nolansherman-spoton:
@Devarsh010 @PratRanj07 Hoping to get some recent contributor eyes on this.

nolansherman-spoton:
@edenhill Hope you could take some time to look at this PR, it resolves a leak we are experiencing with static single consumer groups.

Ankith-Confluent:
Hi @nolansherman-spoton 
Thanks for digging into this. The hang is real, I was able to reproduce it. But when I ran it with `debug=all` I landed somewhere different, and I think the fix is smaller than this.

Librdkafka won't close a consumer that has a fatal error set. `rd_kafka_consumer_close_queue()` hits the `rd_kafka_get_fatal_error()` guard in `rd_kafka_consumer_close_q()`, returns an error, and never starts the close at all. The problem is that we drop that return value.

```go
C.rd_kafka_consumer_close_queue(c.handle.rk, c.handle.rkq)   // error is discarded

for C.rd_kafka_consumer_closed(c.handle.rk) != 1 {           // never becomes 1
    c.Poll(100)
}
```

Since the close never started, `rd_kafka_consumer_closed()` stays 0 forever, so the loop never ends and we never reach `rd_kafka_destroy()`. That's what leaks the handle and its threads.

The debug log backs this up. From the hung consumer there isn't a single `Closing consumer`, `Terminating group` or `CGRPTERM` line, so the close genuinely never begins.

My suggestion would be to keep that return value. If it isn't nil then the close never started, so there's nothing to wait for. Free the handle with `rd_kafka_destroy_flags(..., RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE)` and return the error to the caller instead of nil, so the app knows its offsets weren't committed. Every error that function can return means the close didn't start, so the check doesn't need to be specific to fatal errors.

One upside of doing it that way is that it works by default. With a property you have to switch on, anyone who doesn't know about it still hangs.

kafka/consumer.go
```diff
@@ -547,6 +576,19 @@ func (c *Consumer) Close() (err error) {
...
+	if c.closeSkipOnFatal && getFatalError(c) != nil {
+		atomic.StoreUint32(&c.isClosed, 1)
+		C.rd_kafka_queue_destroy(c.handle.rkq)
+		c.handle.rkq = nil
+		c.handle.cleanup()
+		C.rd_kafka_destroy_flags(c.handle.rk, C.RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE)
+		return nil
+	}
+
 	C.rd_kafka_consumer_close_queue(c.handle.rk, c.handle.rkq)
```
Ankith-Confluent: Catching the error here would fix it without any new config, roughly:

```go
if cErr := C.rd_kafka_consumer_close_queue(c.handle.rk, c.handle.rkq); cErr != nil {
    closeErr := newErrorFromCErrorDestroy(cErr)
    atomic.StoreUint32(&c.isClosed, 1)
    C.rd_kafka_queue_destroy(c.handle.rkq)
    c.handle.rkq = nil
    c.handle.cleanup()
    C.rd_kafka_destroy_flags(c.handle.rk, C.RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE)
    return closeErr
}
```

Ankith-Confluent (changes requested):

nolansherman-spoton:
@Ankith-Confluent Great catch — you were right. The root cause is simpler than my original analysis: `rd_kafka_consumer_close_queue()` returns an error when a fatal error is set (the `rd_kafka_get_fatal_error()` guard in `rd_kafka_consumer_close_q()`), and we were discarding that return value. The close never starts, so `rd_kafka_consumer_closed()` never returns 1, and the loop spins forever.

I have implemented your suggested fix in the latest commit (`8b5ce82`). The changes:

1. Check the return value of `rd_kafka_consumer_close_queue()` — if it returns an error, force-destroy the handle and return the error to the caller.
2. Removed the `go.consumer.close.skip.on.fatal` config option entirely — this fix works by default for all users.
3. Updated tests to verify `Close()` returns immediately with an error when a fatal error is set.

Thanks for taking the time to reproduce and debug this with `debug=all` — the observation that there were no `Closing consumer` / `CGRPTERM` log lines was the key insight.

kafka/consumer_test.go
```diff
@@ -768,3 +768,65 @@ func testConsumerWaitAssignment(c *Consumer, t *testing.T) {
...
+
+	// Verify the fatal error is set
+	fatalErr := getFatalError(c)
+	if fatalErr == nil {
+		t.Fatalf("Expected fatal error to be set")
+	}
+
+	// Close should return immediately with an error (not hang)
+	start := time.Now()
+	err = c.Close()
```
Ankith-Confluent: If this bug comes back, this test will hang rather than fail, since `Close()` never returns and the `elapsed > 5s` check is never reached. Against the unfixed code it just gives `panic: test timed out`. Running it in a goroutine with a `select` and `time.After` keeps it a clean failure, same as `producer_test.go:282` does.

kafka/consumer_test.go
```diff
@@ -768,3 +768,65 @@ func testConsumerWaitAssignment(c *Consumer, t *testing.T) {
...
+	elapsed := time.Since(start)
+
+	// Close should complete nearly instantly (well under 1 second)
+	if elapsed > 5*time.Second {
+		t.Fatalf("Close() took %v, expected < 5s for fatal-errored consumer", elapsed)
+	}
+
+	// Close should return an error indicating the close protocol could not start
+	if err == nil {
```
Ankith-Confluent: Worth checking it is the fatal error specifically, not just any error. `TestFatalErrorClient` in `error_test.go` checks the code and `IsFatal()`. Asserting `c.IsClosed()` afterwards would also cover the flag the new branch sets.

kafka/consumer_test.go
```diff
@@ -768,3 +768,65 @@ func testConsumerWaitAssignment(c *Consumer, t *testing.T) {
...
+		t.Fatalf("Expected Close() to return an error for fatally-errored consumer")
+	}
+
+	t.Logf("Close() completed in %v with error: %v", elapsed, err)
+}
+
+// TestConsumerCloseNormalPath verifies that Close() completes successfully
+// through the normal path when no fatal error is set.
+func TestConsumerCloseNormalPath(t *testing.T) {
```
Ankith-Confluent, line 838: With no broker and no `Subscribe()` the consumer never joins a group, so this close has no assignment, no offsets to commit and no LeaveGroup. It confirms the new branch isn't taken but not that the real close path still works.

Could you add a broker backed one too? `TestConsumerCloseForStaticMember` is just above in this file and already does static membership behind the `testconfRead()` gate.

kafka/consumer_test.go
Ankith-Confluent: This could be the fix
```go
	// Close() must not block here. Run it in a goroutine so that a regression
	// fails this test instead of hanging the whole test binary.
	done := make(chan error, 1)
	start := time.Now()
	go func() { done <- c.Close() }()

	select {
	case err = <-done:
		if err == nil {
			t.Fatalf("Expected Close() to return an error for fatally-errored consumer")
		}
		t.Logf("Close() completed in %v with error: %v", time.Since(start), err)

	case <-time.After(5 * time.Second):
		t.Fatalf("Close() did not return within 5s, the close protocol was never started")
	}
```

Ankith-Confluent (changes requested):
Thanks for making the changes @nolansherman-spoton 
I have provided some comments.

nolansherman-spoton:
Addressed all three review comments in commit `f5f9c1b`:

1. **Goroutine + select pattern**: `TestConsumerCloseReturnsErrorOnFatal` now runs `Close()` in a goroutine with `select`/`time.After(5s)` — a regression will fail the test cleanly instead of hanging the binary.

2. **Error specificity assertions**: Now checks `kafkaErr.Code() == ErrFencedInstanceID`, `kafkaErr.IsFatal() == true`, and `c.IsClosed() == true`.

3. **Broker-backed integration test**: Added `TestConsumerCloseNormalPathWithBroker` — subscribes to a real topic, waits for partition assignment via rebalance callback, then verifies the full close protocol (offset commits, LeaveGroup) completes successfully. Gated behind `testconfRead()`.

kafka/consumer_test.go
nolansherman-spoton: Done — Close() now runs in a goroutine with select/time.After(5s). A regression will produce a clean test failure instead of hanging the binary.

kafka/consumer_test.go
nolansherman-spoton: Done — now asserts kafkaErr.Code() == ErrFencedInstanceID, kafkaErr.IsFatal() == true, and c.IsClosed() == true.

kafka/consumer_test.go
nolansherman-spoton: Done — added TestConsumerCloseNormalPathWithBroker which creates a topic, subscribes with a rebalance callback, waits for partition assignment, then verifies Close() completes the full protocol. Gated behind testconfRead().

kafka/consumer_test.go
nolansherman-spoton: Done — implemented this pattern.

Ankith-Confluent (changes requested):
Hey @nolansherman-spoton 
Thanks for addressing the comments.
Just one more comment related to the broker test.

kafka/consumer_test.go
```diff
@@ -768,3 +768,156 @@ func testConsumerWaitAssignment(c *Consumer, t *testing.T) {
...
+	err = c.Close()
+	if err != nil {
+		t.Fatalf("Close() returned error: %s", err)
+	}
+}
+
+// TestConsumerCloseNormalPathWithBroker verifies that the normal close path
+// works correctly with a real broker — the consumer joins the group,
+// subscribes, and closes cleanly with offset commits and LeaveGroup.
+func TestConsumerCloseNormalPathWithBroker(t *testing.T) {
```
Ankith-Confluent, line 859: This looks good now. All four Close tests pass against a local broker. Thanks for tightening the assertions and adding the broker backed test.

One small thing on `TestConsumerCloseNormalPathWithBroker`. The comment says it closes with offset commits and LeaveGroup, but the commit part is not actually exercised. I ran it with `debug=cgrp,consumer` and got:

```
CGRPTERM: Terminating group "test-close-normal-path-broker" in state up with 0 partition(s)
REMOVE:   Served 1 removed partition(s), with 0 offset(s) to commit
LEAVE:    Group "...": leave (in state up)
LEAVEGROUP: LeaveGroup response received in state up
```

Since the topic is created fresh and no messages are produced or consumed, there is no stored position, so librdkafka skips the commit. Producing a couple of messages and consuming at least one before the close would cover it, and then the log shows `OffsetCommit ... Success`. The LeaveGroup and terminate coverage is good as is.

Ankith-Confluent (changes requested):

kafka/consumer_test.go
nolansherman-spoton: Done — the test now produces 3 messages, consumes one, and commits its offset before calling Close(). This ensures the close path exercises the actual OffsetCommit request to the coordinator, not just LeaveGroup with no stored positions.

kafka/consumer_test.go
```diff
@@ -768,3 +768,199 @@ func testConsumerWaitAssignment(c *Consumer, t *testing.T) {
...
+	t.Logf("Produced 3 messages to %s", topic)
+
+	// Create consumer and subscribe.
+	c, err := NewConsumer(&ConfigMap{
+		"bootstrap.servers":    testconf.Brokers,
+		"group.id":             "test-close-normal-path-broker",
+		"session.timeout.ms":   6000,
+		"max.poll.interval.ms": 10000,
+		"auto.offset.reset":    "earliest",
+		"enable.auto.commit":   false,
```
Ankith-Confluent: `enable.auto.commit: false` stops the close from committing at all, since librdkafka gates that on `enable_auto_commit` in `rd_kafka_assignment_serve_removals()`, so setting it to true (or just dropping the line, true is the default) would cover it. Your `CommitMessage()` call runs before `Close()` and is a manual commit, not part of the close protocol.

Ankith-Confluent (changes requested):
One minor comment, rest of the PR looks good!

kafka/consumer_test.go
nolansherman-spoton: Done — removed enable.auto.commit=false and the manual CommitMessage() call. The test now relies on the default (auto-commit enabled), so Close() commits the stored offset as part of the revocation protocol.

nolansherman-spoton: done, thanks good call out.

Ankith-Confluent (approved):
Thanks for the PR @nolansherman-spoton 
LGTM!!

We will be merging this after a final review.

nolansherman-spoton:
Thank you @Ankith-Confluent ! We are really excited about this.

c1moore:
@Ankith-Confluent I'm not familiar with the release cycle for this SDK.  Do you know what the timeline for landing this PR in a production version will look like?

Ankith-Confluent:
Hey @c1moore 
The dates on the release is not finalised yet, but it should be by the end of this month.

nolansherman-spoton:
Hi @Ankith-Confluent! Is there any release date decided for this yet?

Ankith-Confluent:
Hi @nolansherman-spoton 
We’re currently working on the release and are incorporating a couple of additional fixes, so it’s taking a little longer than expected.

emasab (approved):

emasab (approved):

c1moore:
👋🏻 @Ankith-Confluent @emasab any update on the next release?

I see the PR was merged, but haven't seen a new release cut yet.  This leak is causing fairly significant resource issues for us, so fairly high priority on our end.

Appreciate pushing this through so far!

nolansherman-spoton:
Very happy with the results of this so far. Thanks for getting this into the latest release.
image

PR #1597, opened 2026-08-28
Avoid panic for empty consumer group filters

fallintoplace:
What
----
- Avoid a panic when ListConsumerGroups receives an empty consumer group state or type filter.
- Treat zero-length filters the same as unset filters.

Checklist
------------------
- [x] Contains customer facing changes? This prevents a runtime panic.
  - Added regression coverage for empty state and type filters.

Test & Review
------------
- `go test ./... -count=1`

### DIFF
--- a/kafka/adminapi_test.go
+++ b/kafka/adminapi_test.go
@@ -480,6 +480,26 @@ func testAdminAPIsListConsumerGroups(
 			what, listres, err)
 	}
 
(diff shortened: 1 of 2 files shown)


Ankith-Confluent (approved):
LGTM!
Thanks for the PR @fallintoplace

Ankith-Confluent:
We’ll be merging this and including it in the 2.16 release, which is planned for the end of this month.

ojasvajain (approved):

